Issue #25, opened 2018-05-25
Setting the maximum heap size as a percentage of available memory


SueChaplain:
Capturing the work for https://github.com/eclipse/openj9/pull/1699 and https://github.com/eclipse/openj9/issues/1428.

2 new -XX OpenJDK w/Hotspot options that are now supported by OpenJ9.

PR #36, opened 2018-06-06
Add two new -XX options to set heap as a %

SueChaplain:
-XXMaxRAMPercentage=N and -XXInitialRAMPercentage=N

New in OpenJDK 10 w/Hotspot and recognized by
OpenJ9 in all JDKs for compatibility.

Closes: #25

### DIFF
--- a/docs/xxinitialrampercentage.md
+++ b/docs/xxinitialrampercentage.md
@@ -1,0 +1,38 @@
+<!--
+* Copyright (c) 2017, 2018 IBM Corp. and others
+*
+* This program and the accompanying materials are made
+* available under the terms of the Eclipse Public License 2.0
+* which accompanies this distribution and is available at
+* https://www.eclipse.org/legal/epl-2.0/ or the Apache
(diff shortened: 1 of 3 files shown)


SueChaplain:
@dmitripivkine I believe that Ashutosh is on vacation this week, so please can you review these 2 new XX options for the user documentation. Thanks!

SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

dmitripivkine:
We should mention that type of input value (for both) is "double". So there is correct that value range is 0 to 100 but it can be for example 12.3456 (compatibility with Oracle).

SueChaplain:
Thanks @dmitripivkine - I updated the files to indicate type "double" was acceptable, with an example.
@pmhayward - I'm staging this again for final checks. Please can you look and merge if ok. Thanks!

SueChaplain:
Jenkins doc stage

PR #64, opened 2018-07-30
Remove RDMA components from trace topic

SueChaplain:
JSOR and JVerbs are available only in the
IBM SDK docs and should not be mentioned
here.


SueChaplain:
@pmhayward Minor change that is not worth staging. Removes 2 rows from a table that mention the JSOR and JVerbs features.

PR #70, opened 2018-08-27
Update to include the 6 release/year policy

DanHeidinga:
From the mailing list:
---
Now that the 0.9.0 release is complete, it’s time to think about the schedule for future releases.
With OpenJDK’s new 6 month release cycle, new Java versions are released every 6 months.  With
JDK11’s September release fast approaching and followed closely after by the October JDK8 quarterly
update, there’s a lot of work to do!

What does this OpenJDK release landscape look like?

1) New versions of Java are released every 6 months: March and Sept.
2) Regular updates to existing releases happen quarterly: Jan, April, July & Oct.

What does OpenJ9 need to do?

OpenJ9 needs to at a minimum align with the quarterly updates to ensure critical fixes (including
security patches) are available in OpenJDK with OpenJ9 builds as soon as possible after the changes
appear in OpenJDK.  This would commit the project to at least 4 releases a year.

OpenJ9 should also aim to match the delivery of feature releases and not wait for them appear in the
quarterly update cycle.  This allows roughly a month for projects that are following the feature
releases to migrate before the next quarterly update.  This would commit the project to 2 additional
releases a year.

This would commit the project to 6 releases a year:
* 4 of them are quarterly updates and would include all supported JDK levels
* 2 of them (March & Sept) would be special releases that introduce a single new JDK level

We would work with the Adopt community to ensure that it is clear to users which OpenJ9 releases
should be built for each JDK level (ie: quarterly vs special releases).

Following this proposal, the next OpenJ9 release would be 0.10.0 in Sept and would only target JDK11.
The next OpenJ9 release would be 0.11.0 in Oct and would include the quarterly updates for JDK 8, 10, & 11.

Signed-off-by: Dan Heidinga <daniel_heidinga@ca.ibm.com>

### DIFF
--- a/docs/openj9_support.md
+++ b/docs/openj9_support.md
@@ -47,7 +47,7 @@ supported by any open project.
 | OpenJ9 release  | Release date        | JDK8 (LTS)| JDK9 | JDK10 | JDK11 (LTS) | JDK12   | JDK13 |
 |-----------------|---------------------|-----------|------|-------|-------------|---------|-------|
 | v0.8.0          | March 2018          | Yes       | No   |       |             |         |       |
-| v0.9.0          | August 2018 (\*1)   | Yes       | No   | Yes   |             |         |       |
+| v0.9.0          | August 2018         | Yes       | No   | Yes   |             |         |       |
 | v0.10.0         | September 2018 (\*1)| No        | No   | No    | Yes(\*3)    |         |       |
 | v0.11.0         | October 2018 (\*1)  | Yes       | No   | No    | Yes         |         |       |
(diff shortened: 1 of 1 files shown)


DanHeidinga:
Thanks @SueChaplain.  I've fixed the placement of `only` and updated the table by removing the duplicate September release.  Let me know if there are further issues.

SueChaplain:
Thanks @DanHeidinga. LGTM

DanHeidinga:
Updated as requested.  Please don't merge this until the `[wip]` (work in progress) tag is removed from the title.

DanHeidinga:
Thanks @SueChaplain - I've removed the WIP tag and this is ready to merge.

PR #79, opened 2018-09-06
Add detail to initial Java dump topic

SueChaplain:
Includes sections to cover all the different
parts that make up a Java dump file, with explanations
for the data. Makes use of a very recent javacore
that includes the new cgroup information for
https://github.com/eclipse/openj9/issues/1932

### DIFF
--- a/docs/dump_javadump.md
+++ b/docs/dump_javadump.md
@@ -27,8 +27,638 @@
 Java dumps, sometimes referred to as *Java cores*, are produced when the VM ends unexpectedly because of an operating system signal, `OutOfMemoryError` exception, or a user-initiated keystroke combination. You can also generate a Java dump by calling the Dump API programmatically from your application or specifying the `-Xdump:java` option on the command line.
 
 Java dumps summarize the state of the VM when the event occurs, with most of the information relating to components of the VM.
+The file is made up of a number of sections that provide different types of information. The information that follows describes
+each section and provides examples to help you interpret the data.
 
-For more information about using the Javadump feature, see [Using Javadump](https://www.ibm.com/support/knowledgecenter/SSYKE2_8.0.0/com.ibm.java.vm.80.doc/docs/javadump.html).
+## TITLE
+
+The first section of the Java dump file provides information about the event that triggered the production of the dump.
+In the following example you can see that a `vmstop` event triggered the dump at a specified date and time.
+
+```
+0SECTION       TITLE subcomponent dump routine
+NULL           ===============================
+1TICHARSET     UTF-8
+1TISIGINFO     Dump Event "vmstop" (00000002) Detail "#0000000000000000" received
+1TIDATETIME    Date: 2018/08/30 at 21:55:47:607
+1TINANOTIME    System nanotime: 22012355276134
+1TIFILENAME    Javacore filename:    /home/doc-javacore/javacore.20180830.215547.30285.0001.txt
+1TIREQFLAGS    Request Flags: 0x81 (exclusive+preempt)
+1TIPREPSTATE   Prep State: 0x106 (vm_access+exclusive_vm_access+trace_disabled)
+```
+
+## GPINFO
+
+The GPINFO section provides general information about the system that the JVM is running on. The following example is taken
+from a Java dump that was generated on a Linux system.
+
+```
+NULL           ------------------------------------------------------------------------
+0SECTION       GPINFO subcomponent dump routine
+NULL           ================================
+2XHOSLEVEL     OS Level         : Linux 3.10.0-862.11.6.el7.x86_64
+2XHCPUS        Processors -
+3XHCPUARCH       Architecture   : amd64
+3XHNUMCPUS       How Many       : 4
+3XHNUMASUP       NUMA is either not supported or has been disabled by user
+NULL           
+1XHERROR2      Register dump section only produced for SIGSEGV, SIGILL or SIGFPE.
+NULL           
+```
+
+The content of this section can vary, depending on the cause of the dump. For example, if the dump was caused by a
+general protection fault (gpf), the library in which the crash occurred is also recorded, together with a value shown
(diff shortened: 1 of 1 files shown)


SueChaplain:
Work in progress. @pshipton please can you review what I have so far for accuracy. I wrote it up based on a recent javacore but there are some explanations missing and possibly some errors.

The core file did not have a SHARED CLASSES section as I guess it wasn't enabled. Can you provide me with something please?

Still work to do for THREADS but that's all I have time for this week.

pshipton:
Attaching a javacore containing shared cache info. The cgroup info change is not delivered yet so the javacore does not contain this.
[javacore.20180910.132612.3586.0001.txt](https://github.com/eclipse/openj9-docs/files/2367744/javacore.20180910.132612.3586.0001.txt)

docs/dump_javadump.md
```diff
@@ -27,8 +27,519 @@
...
+NULL           
+1LKREGMONDUMP  JVM System Monitor Dump (registered monitors):
+2LKREGMON          Thread global lock (0x00007FF4F0004FE8): <unowned>
+2LKREGMON          &(PPG_mem_mem32_subAllocHeapMem32.monitor) lock (0x00007FF4F0005098): <unowned>
+2LKREGMON          NLS hash table lock (0x00007FF4F0005148): <unowned>
+...
+NULL           
+```
+
+## THREADS
```
pshipton: There is no explanation of the THREADS section.

SueChaplain: Yep, WIP. Will get back to it soon. Thanks for the SHARED CLASSES output.

SueChaplain: SHARED CLASSES and THREADS sections now in place. I think that the topic now contains a relatively complete set of information. Do we know if the cgroup PR will go in to 0.10.0 please? If it will not, I'll comment out the parts that relate to it and I think this is good to merge.

What I'd like to do as a separate task is take 2 or 3 specific scenarios (hopefully from service) and create troubleshooting examples. E.g. I got a gpf, what can javadump tell me about the root cause? If we can supplement this topic with those examples, I think the info here is as good as it was in the J9 docs. If you have any thoughts on what the scenarios should be, please let me know.

pshipton: cgroup PR is still under review and blocked on the docs being ready. It doesn't need to go into 0.10.0, but it depends on when the review comments are suitably addressed and the branch for 0.10.0 occurs. In theory it should go in after 0.10.0 but no guarantees.

pshipton: gpf or assert, user signal, OOM, deadlock maybe.

docs/dump_javadump.md
```diff
@@ -27,8 +27,635 @@
...
+NULL
+```
+
+## SHARED CLASSES
+
+If the shared classes cache is enabled at run time, the information provided in a Java dump file describes settings that were used when creating the cache, together with summary information about the size and content of the cache.
+
+In the following example, the shared classes cache was created with a Class Debug Area (`-Xnolinenumbers=false`). Byte code instrumentation (BCI) is enabled, which is the default, and VMs sharing the cache are allowed to store classpaths, which is also the default.
+
+The `Cache Summary` shows a cache size (`2SCLTEXTCSZ`) of 16776608 bytes, with a soft maximum size (`2SCLTEXTSMB`) also of 16776608 bytes, which leaves 12691668 bytes of free space (`2SCLTEXTFRB`). The size of the Class Debug Area (`2SCLTEXTDAS`) is 1331200 bytes, which uses only 11% of the available space.
```
pshipton: 11% is the amount used inside of the class debug area, i.e. 156240 is used out of 1331200.

docs/dump_javadump.md
```diff
@@ -27,8 +27,635 @@
...
+A Java thread runs on a native thread. Several lines are recorded for each Java thread in the `Thread Details` subsection, which include the following key pieces of information:
+
+- `3XMTHREADINFO`: The thread name, address information for the VM thread structures and Java thread object, the thread state, and thread priority.
+- `3XMJAVALTHREAD`: The Java thread ID and daemon status from the thread object.
+- `3XMTHREADINFO1`: The native operating system thread ID, priority, scheduling policy, internal VM thread state, and VM thread flags.
+- `3XMTHREADINFO2`: The native stack address range.
+- `3XMTHREADINFO3`: Java callstack information (`4XESTACKTRACE`) or Native call stack information (`4XENATIVESTACK`).
+- `5XESTACKTRACE`: This line indicates whether locks were taken by a specific method.
+
+Java thread priorities are mapped to operating system priority values. The larger the thread priority, the more frequently the thread runs.
```
pshipton: > The larger the thread priority, the more frequently the thread runs.
Not necessarily true https://www.linuxjournal.com/node/10267

SueChaplain: OK, removed that statement.

docs/dump_javadump.md
SueChaplain: Fixed, thanks.

docs/dump_javadump.md
SueChaplain: Thanks for the suggestions on scenarios.

For cgroup, I think the docs are ready .... I've got it in the release notes and what's new, plus this javadump topic contains a mention and example output. One of the developers is writing a blog about it, which I might be able to extract more information from later on. If I can get this topic merged I can close the doc issue. Let me know ....

pshipton: The cgroup review still needs to close before it can be merged, and the review comments haven't yet been addressed. We were hoping to branch for 0.10.0 tomorrow, but seems likely we won't be ready.

SueChaplain:
Jenkins doc stage

docs/dump_javadump.md
SueChaplain: cgroup information removed for 0.10.0

PR #101, opened 2018-10-04
Include scenarios in the Javadump topic

SueChaplain:
To help users interpret a javadump we are
including (initially) 2 scenarios:

1. A dump produced as a result of a GPF
2. A dump produced as a result of a Java OOM

In each case, we walk through the parts of a
javadump that indicate these types of issues
with some explanations. Input provided by
service.

### DIFF
--- a/docs/dump_javadump.md
+++ b/docs/dump_javadump.md
@@ -24,13 +24,31 @@
 
 # Java dump
 
-Java dumps, sometimes referred to as *Java cores*, are produced when the VM ends unexpectedly because of an operating system signal, `OutOfMemoryError` exception, or a user-initiated keystroke combination. You can also generate a Java dump by calling the Dump API programmatically from your application or specifying the `-Xdump:java` option on the command line.
+Java dumps, sometimes referred to as *Java cores*, are produced when the VM ends unexpectedly because of an operating system signal, `OutOfMemoryError`, or a user-initiated keystroke combination. You can also generate a Java dump by calling the Dump API programmatically from your application or specifying the `-Xdump:java` option on the command line.
+
+If your Java application crashes or hangs, Java dumps can provide useful information to help you diagnose the root cause.
+
+- If your application crashes, Javadumps are generated automatically for the following types of failure:
+    - the VM receives an unexpected signal or an assertion failure
+    - the VM runs out of memory
+- If your application hangs, you can trigger the generation of a Java dump by sending a SIGQUIT signal (`kill -3`) to the VM.
+
+    <i class="fa fa-pencil-square-o" aria-hidden="true"></i> **Note:** On Windows, if you started the VM in a console window you can force the VM to produce a Java dump in response to a SIGBREAK signal
+    (Ctrl-Break keyboard combination). If you didn't start in a console window there is no equivalent to a Linux `kill` command on Windows for sending signals.
+    The only option here is to trigger a full system dump by finding the VM process in the **Processes** tab of the Windows Task Manager and clicking
+    **Create dump file**.
+
+To help you understand how a Java dump can help you with problem diagnosis, we've included a few scenarios at the end of the topic that might help you interpret the data:
+
+- [A crash caused by a general protection fault (gpf)](#general-protection-fault)
+- [A Java heap OutOfMemoryError (OOM)](#java-outofmemoryerror)
+
+## Java dump contents
 
 Java dumps summarize the state of the VM when the event occurs, with most of the information relating to components of the VM.
-The file is made up of a number of sections that provide different types of information. The information that follows describes
-each section and provides examples to help you interpret the data.
+The file is made up of a number of sections that provide different types of information.
 
-## TITLE
+### TITLE
 
 The first section of the Java dump file provides information about the event that triggered the production of the dump.
 In the following example you can see that a `vmstop` event triggered the dump at a specified date and time.
@@ -47,7 +65,7 @@ NULL           ===============================
 1TIPREPSTATE   Prep State: 0x106 (vm_access+exclusive_vm_access+trace_disabled)
 ```
 
-## GPINFO
+### GPINFO
 
 The GPINFO section provides general information about the system that the JVM is running on. The following example is taken
 from a Java dump that was generated on a Linux system.
@@ -90,7 +108,7 @@ The hexadecimal number recorded for `VM flags` ends in MSSSS, where M is the VM 
 
 A value of `0000000000000000` (0x00000) indicates that a crash occurred outside of the VM.
 
-## ENVINFO
+### ENVINFO
 
 This section contains useful information about the environment in which the crash took place, including the following data:
 
@@ -206,7 +224,7 @@ NULL           -----------------------------------------------------------------
 NULL           
 -->
 
-## NATIVEMEMINFO
+### NATIVEMEMINFO
 
 This section records information about native memory that is requested by using library functions such as `malloc()` and `mmap()`.
 Values are provided as a breakdown, per component, indicating the total number of bytes allocated and the number of native memory allocations.
@@ -278,7 +296,7 @@ NULL
 This section does not record memory that is allocated by application or JNI code and is typically a little less than the
 value recorded by operating system tools.
 
-## MEMINFO
+### MEMINFO
 
 This section relates to memory management, providing a breakdown of memory usage in the VM for the object heap,
 internal memory, memory used for classes, the JIT code cache, and JIT data cache in decimal and hexadecimal format.
@@ -350,7 +368,7 @@ NULL
 In the example, the GC History (`1STGCHTYPE`) section is blank. This section is populated if a garbage collection cycle occurred in
 a VM that is being diagnosed with the trace facility.
 
-## LOCKS
+### LOCKS
 
 This section of the Java dump provides information about locks, which protect shared resources from being accessed by more than one entity at a time. The information is essential in a deadlock situation, where two threads attempt to synchronize on an object and lock an instance of a class. Precise information is recorded about the threads that are causing the problem, which enables you to identify the root cause.
 
@@ -378,7 +396,7 @@ NULL
 NULL           
 ```
 
-## THREADS
+### THREADS
 
 The THREADS section of a Java dump file provides summary information about the VM thread pool and detailed information about Java threads, native threads, and stack traces. Understanding the content of this section can help you diagnose problems that are caused by blocked or waiting threads.
 
@@ -477,7 +495,7 @@ NULL
 NULL
 ```
 
-## HOOKS
+### HOOKS
 
 This section shows internal VM event callbacks, which are used for diagnosing performance problems in the VM. Multiple hook interfaces are listed, which include their individual hook events. The following example shows data for the `J9VMHookInterface`, including the call site location (source file:line number), start time, and duration of the last callback and the longest callback.
 
@@ -523,7 +541,7 @@ NULL           -----------------------------------------------------------------
 NULL
 ```
 
-## SHARED CLASSES
+### SHARED CLASSES
 
 If the shared classes cache is enabled at run time, the information provided in a Java dump file describes settings that were used when creating the cache, together with summary information about the size and content of the cache.
 
@@ -617,7 +635,7 @@ NULL
 NULL
 ```
 
-## CLASSES
+### CLASSES
 
 The classes section shows information about class loaders. The first part is a summary that records each available class loader (`2CLTEXTCLLOADER`) followed by the number of libraries and classes that it loaded. This information is followed by a more detailed list of libraries (`1CLTEXTCLLIB`) and classes (`1CLTEXTCLLO`) that are loaded.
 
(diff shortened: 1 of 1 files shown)


SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

SueChaplain:
@pshipton - I worked with service to include some scenarios for troubleshooting javadump files. At the moment there are just a couple (gpf and Java OOM). I wanted you to take a look to make sure you are happy with the content.

docs/dump_javadump.md
```diff
@@ -24,13 +24,31 @@
...
 
-Java dumps, sometimes referred to as *Java cores*, are produced when the VM ends unexpectedly because of an operating system signal, `OutOfMemoryError` exception, or a user-initiated keystroke combination. You can also generate a Java dump by calling the Dump API programmatically from your application or specifying the `-Xdump:java` option on the command line.
+Java dumps, sometimes referred to as *Java cores*, are produced when the VM ends unexpectedly because of an operating system signal, `OutOfMemoryError`, or a user-initiated keystroke combination. You can also generate a Java dump by calling the Dump API programmatically from your application or specifying the `-Xdump:java` option on the command line.
+
+If your Java application crashes or hangs, Java dumps can provide useful information to help you diagnose the root cause.
+
+- If your application crashes, Javadumps are generated automatically for the following types of failure:
+    - the VM receives an unexpected signal or an assertion failure
+    - the VM runs out of memory
+- If your application hangs, you can trigger the generation of a Java dump by sending a SIGSEGV signal (`kill -3`) to the VM.
```
DanHeidinga: I think `SIGSEGV` should be `SIGQUIT`.  We don't want people to send `SIGSEGV` as it makes it look like a crash rather than a user generated javacore

docs/dump_javadump.md
```diff
@@ -24,13 +24,31 @@
...
+
+    <i class="fa fa-pencil-square-o" aria-hidden="true"></i> **Note:** On Windows, if you started the VM in a console window you can force the VM to produce a Java dump in response to a SIGBREAK signal
+    (Ctrl-Break keyboard combination). If you didn't start in a console window there is no equivalent to a Linux `kill` command on Windows for sending signals.
+    The only option here is to trigger a full system dump by finding the VM process in the **Processes** tab of the Windows Task Manager and clicking
+    **Create dump file**.
+
+To help you understand how a Java dump can help you with problem diagnosis, we've included a few scenarios at the end of the topic that might help you interpret the data:
+
+- [A crash caused by a general protection fault (gpf)](#general-protection-fault)
+- [A Java heap OutOfMemoryError (OOM)](#java-outofmemoryerror)
```
DanHeidinga: It would be really good to add using javacores to diagnose a hang.  Most hangs aren't really hangs and using multiple javacores over a small time period can show that.  This can be added in a later update

docs/dump_javadump.md
```diff
@@ -660,6 +678,206 @@ NULL           =================================
...
+4XENATIVESTACK               (0xB7719F72 [libpthread.so.0+0x6f72])
+4XENATIVESTACK               clone+0x5e (0xB763543E [libc.so.6+0xee43e])
+```
+
+The extract tells you that the current thread was `java/lang/Thread`, and information is provided about the Java callstack and Native callstack
+(`3XMTHREADINFO3`) at the point at which the crash occurred. To simulate a crash caused by a bug in customer or third party JNI code, our testcase
+calls into a JNI method whose native implementation causes a crash. The Java callstack shows the call into the JNI native method (`JNIcrasher`), and the
+native callstack shows the point of failure. In this example, the native call stack does not include any function names to help you isolate the error
+in the native code. The system dump that would have been produced alongside the Java dump can provide this information. Open the
+system dump with the [Dump viewer](tool_jdmpview.md) and use the `info thread` command to print the Java and native stack for the current thread.
```
DanHeidinga: Does `info thread` need the j9vmthread address (the low-level `!stackslots 0xJ9VMThread_address` does)?  If it does. we should add more details to indicate the full command in this example.

docs/dump_javadump.md
```diff
@@ -660,6 +678,206 @@ NULL           =================================
...
+Although the Java code we used in this scenario was deliberately triggering an `OutOfMemoryError` in a spectacular way, similar allocation issues can and do occur when dealing with large data sets such as XML files.
+
+The next step in diagnosing the problems would be to open the system dump that gets generated automatically when an `OutOfMemoryError` occurs. Open
+the dump with the [Eclipse Memory Analyzer tool (MAT)](https://www.eclipse.org/mat/) and search for the `StringBuffer` object, which should provide further clues about what went wrong. A common
+example is seeing the same `String` duplicated over and over again, which might indicate that code is stuck in a loop.
+
+  <i class="fa fa-pencil-square-o" aria-hidden="true"></i> **Note:** If you want to use MAT to analyze your system dump, you must install the Diagnostic Tool Framework for Java (DTFJ) plugin in the Eclipse IDE. Select the following menu items:
+
+```
+Help > Install New Software > Work with "IBM Diagnostic Tool Framework for Java" >
```
DanHeidinga: Two concerns here:
1) OpenJ9 shouldn't redirect to IBM tools.  It would be better to direct to the Eclipse MAT downloads.
2) Have we confirmed that this works with OpenJ9?  I'd like confirmation that it works before this goes live

docs/dump_javadump.md
```diff
@@ -660,6 +678,206 @@ NULL           =================================
...
+example is seeing the same `String` duplicated over and over again, which might indicate that code is stuck in a loop.
+
+  <i class="fa fa-pencil-square-o" aria-hidden="true"></i> **Note:** If you want to use MAT to analyze your system dump, you must install the Diagnostic Tool Framework for Java (DTFJ) plugin in the Eclipse IDE. Select the following menu items:
+
+```
+Help > Install New Software > Work with "IBM Diagnostic Tool Framework for Java" >  
+```
+
+Unlike our scenario, if you receive an `OutOfMemoryError` and the MEMINFO section shows that there is very little space left
+on the Java heap, the current thread information is not important. The current thread is simply the unlucky thread that happened to be current when the space ran out. In this situation you might want to look at increasing your Java heap size. For help with this task, see [How to do heap sizing](https://www.ibm.com/support/knowledgecenter/SSYKE2_8.0.0/com.ibm.java.vm.80.doc/docs/mm_heapsizing.html).
```
DanHeidinga: Is there an OpenJ9 link rather than IBM one we can use here?

docs/dump_javadump.md
SueChaplain: Hangs is on the list, as is native OOM and deadlocks. 

I just have one core file for hangs, which doesn't sound like it covers your scenario .. but we can add multiples.

docs/dump_javadump.md
SueChaplain: I'll let @pshipton comment here. My understanding from earlier updates made in System dump was that the Eclipse MAT tool requires DTFJ to analyze the system dump. As this info came from Peter, my guess is that it has been tested.

docs/dump_javadump.md
paulcheeseman: Yes, that should definitely say `SIGQUIT`

docs/dump_javadump.md
paulcheeseman: > Does info thread need the j9vmthread address

I don't know.

However, it's probably worth mentioning that in this specific example when I tried running `info thread` on the system dump the stack was incomplete. When I tried getting the native stack with gdb the missing frames were present.

It's also not really clear to me why the native stack is very rarely shown properly in the javacore (i.e. most function names are usually missing).

docs/dump_javadump.md
paulcheeseman: The Hang example I created we have is for a real Hang. Specifically, in the example there is one thread holding a lock that is blocking all the other worker threads.

Providing examples for situations that look like hangs, but aren't actually hangs, is a much larger topic.

docs/dump_javadump.md
SueChaplain: Thanks.  Fixed.

docs/dump_javadump.md
DanHeidinga: Thanks for the confirmation.  One more thing we should add to the todo list

docs/dump_javadump.md
SueChaplain: > Open the system dump with the Dump viewer and use the `info thread` command to print the Java and native stack for the current thread.

We have this documented:

`info thread [*|all|<native_thread_ID>`

*If you supply no parameters, the command shows information about the current thread.*

So in this case, it sounds as if it is ok to tell people to use just `info thread`. 
??

paulcheeseman: Yes, `info thread` should give you the Java and native stack for the current thread, in dumps where this _is_ a "current thread". This is the case for dumps triggered by crashes and OOMs, but not for user generated dumps (e.g. dumps triggered to diagnose hang/performance issues).

SueChaplain: These instructions are in the gpf scenario, so should be relevant?

paulcheeseman: Yes, it should work fine for a dump triggered by a GPF. It's certainly the first thing you want to try after looking at the javacore and concluding that the crash happened in native code.

SueChaplain:
@DanHeidinga @pshipton - I hope all the review queries are answered to everyone's satisfaction. If so, please go ahead and merge. 

@paulcheeseman & I have been working on the next 2 scenarios: NativeOutOfMemoryError and Deadlocks .... coming soon to a pull request near you :-)

docs/dump_javadump.md
DanHeidinga: I think the important thing is to explain the technique of using multiple javacores to determine if an application is progressing or not - pointing out the relevant parts of the data to check.  

But I agree, it can be a separate task.  @SueChaplain Is there an issue opened with the list of future topics to cover?

docs/dump_javadump.md
DanHeidinga: Opened an issue at OpenJ9 so this isn't lost: https://github.com/eclipse/openj9/issues/3164

docs/dump_javadump.md
DanHeidinga: ok

docs/dump_javadump.md
```diff
@@ -660,6 +678,206 @@ NULL           =================================
...
+4XENATIVESTACK               (0xB6C502FA [libj9prt29.so+0x382fa])
+```
+
+To simulate a Java `OutOfMemoryError`, our testcase works by repeatedly appending characters to a `StringBuffer` object in an infinite loop. The Java callstack shows the `HeapBreaker.main` method appending characters (`java/lang/StringGuffer.append`) until the method `java/lang/StringBuffer.ensureCapacityImpl()` throws the `OutOfMemoryError`.
+
+StringBuffer objects are wrappers for character arrays (`char[]`) and when the capacity of the underlying array is reached, the contents are automatically copied into a new, larger array. The new array is created in the `StringBuffer.ensureCapacity()` method, which more or less doubles the size of the old array. In our scenario, the array takes up all the remaining space in the Java heap.
+
+Although the Java code we used in this scenario was deliberately triggering an `OutOfMemoryError` in a spectacular way, similar allocation issues can and do occur when dealing with large data sets such as XML files.
+
+The next step in diagnosing the problems would be to open the system dump that gets generated automatically when an `OutOfMemoryError` occurs. Open
```
pshipton: Would the gc history section help in this case? Showing the amount of heap memory being allocated which lead to the OOM.

docs/dump_javadump.md
pshipton: In general I'd expect the platform native debugger to do a better job on the native stack. Not sure if jdmpview does a reliable job as I've never tried this approach.

docs/dump_javadump.md
paulcheeseman: Yes, that information is useful for determining whether the problem was caused by a full heap or an unexpectedly large allocation request.

SueChaplain: @pshipton - I've added a short paragraph about the GC History section. Please can you check it to make sure I have not misinterpreted the output.

docs/dump_javadump.md
SueChaplain: I'm going to get this PR merged tomorrow as it will be getting stale. I don't think there is anything in here that is wrong, and we can discuss further scenarios and tweaks in another issue. @DanHeidinga I'll make sure I involve you and @pshipton in the reviews as early as possible.

SueChaplain:
Jenkins doc stage

SueChaplain:
@doveye - I staged the work. Please can you do one final check and merge if ok. Thanks!

AdamBrousseau:
Jenkins doc stage

doveye:
Some fairly minor comments. Although this isn't IBM doc, I think it would be good to keep some of the guidelines, for simplicity/clarity/consistency.

**GPF scenario**

- > To simulate a crash caused by a bug in customer or third party JNI code, our testcase...

Suggest something like "To simulate a crash caused by a bug in an application, this example...", to avoid mentioning customers or third parties, and because a testcase hasn't previously been mentioned/explained.

- "calls into" - remove "into", and "call into" - change "into" to "to"

- > The system dump that would have been produced alongside the Java dump can provide this information.

Suggest something like "You can get this information from a system dump, which is usually produced alongside the Java dump", to avoid the past conditional tense and put the "information" bit first, since that's the thing we're concerned with

**OOM scenario**

- > To simulate a Java OutOfMemoryError, our testcase works by repeatedly appending characters to a StringBuffer object in an infinite loop.

Suggest something like "To simulate a Java OutOfMemoryError, this example application repeatedly appends characters to a StringBuffer object in an infinite loop."

- "our sample output" change "our" to "the"

- > Although the Java code we used in this scenario was deliberately triggering an  OutOfMemoryError in a spectacular way

Suggest something like "Although the Java code in this scenario deliberately triggered an OutOfMemoryError in a obvious way" (or maybe a pronounced way?)

- > The next step in diagnosing the problems would be to...

Suggest changing "would be" to "is"

- > Unlike our scenario, if you receive an OutOfMemoryError...

Suggest "If, unlike the previous scenario, you receive an OutOfMemoryError..."

- Remove the word "unlucky" as it's not necessary

- > you might want to look at increasing your Java heap size

Suggest "you might want to increase your Java heap size"

SueChaplain:
Thanks for the review @doveye - I made the changes recommended. I've not staged the PR as the changes were just to words, which are probably quicker to check in GitHub viewer.

Issue #112, opened 2018-10-11
Enable options for idle tuning when running in container


ashu-mehra:
**Issue or pull request number:**
https://github.com/eclipse/openj9/issues/3226

**Overview:**
Enables `-XX:+IdleTuningGcOnIdle` and `-XX:+IdleTuningCompactOnIdle` when OpenJ9 is running in a container.
Doc pages https://github.com/eclipse/openj9-docs/blob/master/docs/xxidletuninggconidle.md and https://github.com/eclipse/openj9-docs/blob/master/docs/xxidletuningcompactonidle.md need to be updated to include this change in default behavior when running in container.

**Applies to the following JDK versions:**
All JDK versions

**Applies to the following platforms:**
Linux X|P|Z

**For new command line options:**
N.A.

SueChaplain:
@ashu-mehra - I have a query.

In https://www.eclipse.org/openj9/docs/xxidletuningminidlewaittime/, we state that:

>The value used for <secs> specifies the minimum length of time in seconds that the VM is idle before the state is set to idle. Setting the value to 0 disables this feature, which causes the following idle tuning options to have no effect:

>-XX:+IdleTuningCompactOnIdle
>-XX:+IdleTuningGcOnIdle
>-XX:IdleTuningMinFreeHeapOnIdle

We also state that the default value is `0`.

I'm assuming that this topic needs updating too? What do we need to say please?

ashu-mehra:
Yes, this is confusing. 
Actually the default value of `-XX:IdleTuningMinIdleWaitTime` is set as follows:
If either of `-XX:+IdleTuningGcOnIdle` or `-XX:+IdleTuningCompactOnIdle` is specified, then the default value is 180 secs, else, it is 0.
If the user explicitly sets the value to 0, then `-XX:+IdleTuningGcOnIdle` or `-XX:+IdleTuningCompactOnIdle` won't have any effect.

Does that help?

Issue #106, opened 2018-10-10
Enable -XX:+UseContainerSupport option by default


ashu-mehra:
**Issue or pull request number:**
https://github.com/eclipse/openj9/issues/2263

**Overview:**
*Describe the external change and the impact/benefit from a user perspective.*
This issue enables option `-XX:+UseContainerSupport` by default.
See https://github.com/eclipse/openj9-docs/blob/c604a7e9fcbdb3a1576d25ddaf4ae520092f2211/docs/xxusecontainersupport.md for the effect of enabling this option by default. Content in the same link also needs to be updated as the option has now become default.

**Applies to the following JDK versions:**
All JDK versions

**Applies to the following platforms:**
Linux X|P|Z

**For new command line options:**
N.A.

PR #114, opened 2018-10-15
[WIP] Container-aware options are on by default

SueChaplain:
To address docs issues #112 and #106

-XX:+UseContainerSupport is set by default if
the JVM is running in a container. This also
causes some idle tuning options to be enabled
by default.

Closes: #112
Closes: #106


pshipton:
@SueChaplain while the change for #106 is merged, the change for #112 is scheduled for the 0.12 milestone atm.

PR #116, opened 2018-10-16
fix markdown escaping in docs/cmdline_migration.md

sxa555:
Signed-off-by: Stewart Addison <sxa@uk.ibm.com>


pmhayward:
Stewart, am I right in thinking that this commit is to add backslash escape characters before the pipe symbols in the options `-XX:[+|-]DisableExplicitGC` and `-XX:[+|-]HeapDumpOnOutOfMemory`?
If so, are you sure that they are needed? In normal text, the pipes certainly would need escaping as they'd otherwise be interpreted as part of the table structure. But here, the options are highlighted in the Markdown source as code (using backticks) and all characters within that highlighting are simply rendered verbatim (including the backslashes)!

sxa555:
@pmhayward View the [markdown doc in github](https://github.com/eclipse/openj9-docs/blob/master/docs/cmdline_migration.md) - it's clearly broken. I'm not sure how you're looking at it if you can't see it. The [version in my repo shows correctly](https://github.com/sxa555/openj9-docs/blob/fix_migration_escaping/docs/cmdline_migration.md), so it's either wrong or the github markdown rendering engine is faulty - either way I suspect on github is how most people will view it so it kinda needs to be right
image

pmhayward:
I'm looking at it on the Eclipse docs website (www.eclipse.org/openj9/docs/cmdline_migration), which is what it's written for and where it's clearly _not_ broken :)
By all means use GitHub to give you a rough idea of what the docs look like, but the website is the place to go for the definitive version. Adding the backslash would obviously be wrong in that context (resulting in the wrong syntax being displayed for the options) and would also cause further problems in other places where this information is reused.

SueChaplain:
Raised a bugzilla to get the repo description updated by the Eclipse admins: https://bugs.eclipse.org/bugs/show_bug.cgi?id=540214

We can try and redirect new arrivals to the online docs and avoid them looking at rendered markdown in GitHub where different extensions we are using won't be supported. @sxa are you ok with us closing this PR with that in place? There must be many doc repos in GitHub that have the same issue.

sxa555:
If we're happy with it not displaying properly in github then I'm ok with closing this since it works on the eclipse openj9 page. It's a shame it doesn't work in both places though

SueChaplain:
@sxa I agree it would be nice to have it work in both, but some of the nice extensions we are using for the doc would have to be supported by GitHub. At least the project description is now updated to tell people how to view it (bug re-opened for them to fix the link!)

PR #119, opened 2018-10-17
Updates for concurrent scavenge on x86 support

### LINKED ISSUE #99: Support for concurrent scavenge on x86
https://github.com/eclipse/openj9/issues/3054

**Overview:**
The above issue is tracking a series of changes being made to OpenJ9 to support the pause-less garbage collector on the x86 platform. The pause-less garbage collector technology was originally implemented on the z architecture targeting the z14 hardware generation.

8 and later

64-bit Linux X / Windows

-Xgc:concurrentScavenge

This is the same option as on z. It takes no options and has the affect as it does on z. At present we expect to support all currently supported x86 hardware running 64bit Linux or Windows. We will provide further updates as implementation and testing concludes if further restrictions / qualifications on this are necessary.

This option is supported only on the 64-bit JVM with the Generational Concurrent (gencon) garbage collection policy. When this mode is enabled, the JVM attempts to reduce GC pause-times for response-time sensitive, large heap applications.

pmhayward:
Changes to -Xgc:concurrentScavenge
Resolves #99

### DIFF
--- a/docs/xgc.md
+++ b/docs/xgc.md
@@ -34,7 +34,7 @@ Options that change the behavior of the Garbage Collector (GC).
 
 | Parameter                                                       | Effect                                                                                                  |
 |-----------------------------------------------------------------|---------------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


pmhayward:
Jenkins doc stage

pmhayward:
Jenkins doc stage

PR #139, opened 2018-11-14
Correct 0.12.0 release line to say yes for 8 & 11

DanHeidinga:
It had incorrectly said "yes" only for JDK12 which won't ship until
March.  Update the "No" for JDK8 and JDK11 to say "Yes" based on the
statements by RedHat that they will step forward as the stewards of
JDK8 and by others to indicate they will help with backports.

I've left the `(*2)` note on both of those "Yes"s though.

Signed-off-by: Dan Heidinga <daniel_heidinga@ca.ibm.com>

### DIFF
--- a/docs/openj9_support.md
+++ b/docs/openj9_support.md
@@ -49,10 +49,10 @@ this table over time.
 | v0.9.0          | August 2018         | Yes       | Yes   |             |         |       |
 | v0.10.0         | September 2018      | No        | No    | Yes(\*3)    |         |       |
 | v0.11.0         | October 2018        | Yes       | No    | Yes         |         |       |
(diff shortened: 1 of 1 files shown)


pshipton:
Forgot to tag @DanHeidinga ^

DanHeidinga:
Updated as suggested

SueChaplain:
LGTM

pshipton:
Also note the 0.15.0 line is incorrect, as Java 13 is not introduced until Sept.

pshipton:
i.e. for 0.15.0, JDK12 should be yes, and nothing for JDK13
0.16.0 adds only JDK13 in Sept.

pshipton:
https://github.com/eclipse/openj9-docs/issues/144

PR #161, opened 2018-12-12
Add change for OPENJ9_JAVA_OPTIONS env. variable to "What's new"

pmhayward:
Topic version0.12.md updated


SueChaplain:
@pmhayward. Looks good but I think there is a slight change needed to the concurrentScavenge section, which has been steeped in confusion due to previous inaccuracies in the docs. I think it needs to say something along the lines of:


**Improved support for pause-less garbage collection on 64-bit Linux x86 platforms**

*In Eclipse OpenJ9 V0.11.0, support was added for -Xgc:concurrentScavenge on Linux x86-64 virtual machines that use compressed references. In this release, support is now available for Linux x86-64 large heap virtual machines (non-compressed references).*

No need to mention Z at all .... this is a statement about Linux support. The -Xgc topic is accurate.

PR #181, opened 2019-01-07
Remove Next / Previous buttons to fix bug

SueChaplain:
The next / previous buttons have a problem due
to the way we have implemented some of the topics.
Where we have related options, these still have their own
TOC entries but the content is implemented in a common
markdown file. For example: -Xminf and -Xmaxf are
both in topic xminf.md. A result of this is that
clicking "next" when you are in -Xmaxf takes you to
-Xmint instead of -Xmaxt.

Since most of our topics are either self contained
(E.g. javadump) or reference material (command options),
the next / previous buttons are somewhat superfluous.

This PR removes next/previous from the user documentation.

### DIFF
--- a/docs/stylesheets/oj9.css
+++ b/docs/stylesheets/oj9.css
@@ -155,9 +155,11 @@
 height 3rem;
 }
 */
+/*
 .md-footer-nav__inner.md-grid {
 margin-top: 5rem;
 }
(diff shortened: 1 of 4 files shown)


SueChaplain:
Think I need to start this one again! Closing

PR #184, opened 2019-01-09
Remove previous/next links from footer

### LINKED ISSUE #183: Issue with <- previous and next -> buttons for options that share .md files
Previous and Next links in topic footers can be incorrect if several entries in the navigation share a common .md file. There's no possible solution without creating dummy topics - cf. DITA! - so that the build can link to the appropriate previous/next topic. Simple solution for now is to remove links from footer.
(Framework of footer left in place to maintain structure of page and highlighting in LH navigation. But note, only *last* occurrence of a .md file in nav is highlighted, so this is not a perfect solution!)

pmhayward:
Resolves #183


SueChaplain:
LGTM

PR #189, opened 2019-01-14
Doc updates for reverting shared classes enabled by default

doveye:
Comment out section in What's New for 0.12.0, comment out section about enabling in modify shrc.md, and replace with old content about setting -Xshareclasses.

https://github.com/eclipse/openj9-docs/issues/187

[skip ci]

### DIFF
--- a/docs/shrc.md
+++ b/docs/shrc.md
@@ -37,10 +37,13 @@ Memory footprint is reduced by sharing common classes between applications that 
 
 ## Enabling class data sharing
 
+You enable class data sharing by setting the `-Xshareclasses` option on the command line when you start your application. By default, OpenJ9 always shares both the bootstrap and application classes that are loaded by the default system class loader.
+
+<!--
   Class data sharing is enabled by default for bootstrap classes only, unless your application is running in a container. You can use the `-Xshareclasses` option to change the default behavior, including the name and location of the default shared classes cache. Trace point `j9shr.2271` is activated if the default cache cannot be started, so you can enable this trace point to determine whether the default cache started successfully. You can treat the default cache like any other shared classes cache, for example you can print statistics for it, change the soft maximum limit size, or delete it. Note that if you have multiple VMs and you do not change the default shared classes behavior, the following applies:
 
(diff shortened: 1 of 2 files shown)


docs/shrc.md
```diff
@@ -37,10 +37,13 @@ Memory footprint is reduced by sharing common classes between applications that
...
 
 - If the VMs are from a single Java installation, they will share a single default cache.
-- If the VMs are from different Java installations, of the same Java release and installed by the same user, each VM checks whether the existing default shared cache in the cache directory is from the same Java installation as the VM. If not, the VM deletes that shared cache, then creates a new one. To avoid this situation, use `-Xshareclasses:cacheDir=<dir>` to specify a different cache directory for each Java installation.
+- If the VMs are from different Java installations, of the same Java release and installed by the same user, each VM checks whether the existing default shared cache in the cache directory is from the same Java installation as the VM. If not, the VM deletes that shared cache, then creates a new one. To avoid this situation, use `-Xshareclasses:cacheDir=<dir>` to specify a different cache directory for each Java installation. -->
```
pshipton: This paragraph should not be removed, it isn't related to shared classes by default.

pshipton: Well I guess it is a follow on from shared classes by default, but it remains true whether shared classes by default is enabled or not. Do we also have this information elsewhere?

doveye: I did wonder whether some of this info might still be relevant.
I can't see the info in this para anywhere else. The closest thing is in the original IBM content which has yet to be moved to OpenJ9, where we say that the VM will attempt to destroy a cache created by a different service release (https://www.ibm.com/support/knowledgecenter/SSYKE2_8.0.0/com.ibm.java.vm.80.doc/docs/shrc_pd_deploying_compat.html?view=kc). We don't mention same user or same release.
I could move it to the Best practices section in this topic, as another reason for setting a specific cache directory.

pshipton: Sounds good.

doveye: I've modified the cache dir bullet point in "Best practices" to contain this info:
shrc-cachedir

Issue #199, opened 2019-01-21
Fix incorrect information in the command-line migration topic


SueChaplain:
**Page containing the error:**

https://www.eclipse.org/openj9/docs/cmdline_migration/

-XX:ThreadStackSize is expressed in KB, it is not equivalent to -Xss
OpenJ9 does not implement -XX:ThreadStackSize

SueChaplain:
@pshipton 

If OpenJ9 does not implement -XX:ThreadStackSize, shall I just remove the table and the topic from the docs. I'm not sure what relevance the "expressed in KB" has.

pshipton:
> shall I just remove the row from the table and the topic from the docs

yes. This works for me.

pshipton:
> I'm not sure what relevance the "expressed in KB" has

If we have any other place that mentions -XX:ThreadStackSize is equivalent to -Xss, then it needs to be fixed because this is not true.

PR #207, opened 2019-02-04
Fix innaccuracy in the Switching to OpenJ9 topic

SueChaplain:
Incorrect reference to -XXThreadStackSize being
accepted by OpenJ9 and equivalent to -Xss.
Removed row from table and topic from user docs.

Closes: #199


SueChaplain:
@pshipton - change as requested for 0.12.0

PR #218, opened 2019-02-26
Updates to support topic for 0.13.0

SueChaplain:
Update support table and add sections for
OpenJDK 12 supported platforms and build system.

### DIFF
--- a/docs/openj9_support.md
+++ b/docs/openj9_support.md
@@ -43,26 +43,23 @@ this table over time.
 
 ## Eclipse OpenJ9 releases
 
-| OpenJ9 release  | Release date        | JDK8 (LTS)| JDK10 | JDK11 (LTS) | JDK12   | JDK13    |
-|-----------------|---------------------|-----------|-------|-------------|---------|----------|
-| v0.8.0          | March 2018          | Yes       |       |             |         |          |
-| v0.9.0          | August 2018         | Yes       | Yes   |             |         |          |
-| v0.10.0         | September 2018      | No        | No    | Yes(\*3)    |         |          |
-| v0.11.0         | October 2018        | Yes       | No    | Yes         |         |          |
-| v0.12.0         | January 2019        | Yes (\*2) | No    | Yes         |         |          |
-| v0.13.0         | March 2019 (\*1)    | No        | No    | No          | Yes(\*3)|          |
-| v0.14.0         | April 2019 (\*1)    | Yes (\*2) | No    | Yes (\*2)   | Yes     |          |
-| v0.15.0         | July 2019 (\*1)     | Yes (\*2) | No    | Yes (\*2)   | Yes     |          |
-| v0.16.0         | September 2019 (\*1)| No        | No    | No          | No      | Yes(\*3) |
+| OpenJ9 release  | Release date        | JDK8 (LTS)| JDK11 (LTS) | JDK12   | JDK13    |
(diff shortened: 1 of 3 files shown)


SueChaplain:
Jenkins doc stage

SueChaplain:
@pshipton - please can you suggest someone to review this please. In particular the support topic, staged here: https://eclipse.github.io/openj9-docs-staging/218/openj9_support/

I need confirmation on platform support and the build machines for 12. Thanks!

docs/openj9_support.md
```diff
@@ -180,3 +216,14 @@ The project build and test OpenJDK with OpenJ9 on a number of platforms. The ope
...
 | macOS x86 64-bit            | macOS 10.13.5            | xcode/clang 9.4                 |
 | AIX POWER BE 64-bit         | AIX 7.1 TL04             | xlc/C++ 13.1.3                  |
+
+### OpenJDK 12
+
+| Platform                    | Operating system         |  Compiler                       |
+|-----------------------------|--------------------------|---------------------------------|
+| Linux x86 64-bit            | Ubuntu 16.04             | gcc 7.3                         |
+| Linux on POWER LE 64-bit    | Ubuntu 16.04             | gcc 7.3                         |
+| Linux on IBM Z 64-bit       | Ubuntu 16.04             | gcc 7.3                         |
```
pshipton: For all versions, zlinux is built with gcc 7.4

SueChaplain:
Jenkins doc stage

docs/openj9_support.md
SueChaplain: Fixed

SueChaplain:
@pshipton - gcc levels on Linux for Z are fixed. There is an open query about our note:

>Not all of these distributions are tested, but Linux distributions that have a minimum glibc version 2.12 are expected to function without problems.

At AdoptOpenJDK there is a an extra part to this sentence, which I believe was due to differences in compiler levels on their build machines. I want to check that each statement is still accurate please?

Adopt note:

>Any other Linux distributions that have a minimum glibc version 2.12 on x86 or version glibc version 2.17 on other architectures are expected to function without problems.

pshipton:
The OpenJ9 note about the 2.12 minimum reflects IBM usage and testing of OpenJ9. Although the OpenJ9 project tests on Ubuntu 16 atm in the open (glibc 2.23), IBM builds and tests it on the minimum requirement of glibc 2.12 (RHEL 6) across x and zLinux platforms. pLinux (LE) was introduced and is built on RHEL 7 which is 2.17, however IBM also supports BE from the same OpenJ9 source, which is again built and tested on RHEL 6.

It isn't feasible to use RHEL 6 for open source because RHEL 6 is a paid product. Open source can use the CentOS "equivalent", but CentOS 6 is only available for xLinux and not zLinux. zLinux needs to use CentOS 7 (glibc 2.17). 

Adopt builds on CentOS 6 (2.12) for xLinux and CentOS 7 (2.17) for p and zLinux. This means the builds supplied by Adopt will only run on these minimums and above, but still the OpenJ9 source itself is compatible to be built and run on the 2.12 minimum, except for pLinux (LE) where there is no 2.12 OS available.

SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

SueChaplain:
@pshipton - please merge if you consider this is now ready.

PR #228, opened 2019-03-07
Add new AdoptOpenJDK topic

SueChaplain:
For users who obtain binaries from Adopt, they need to be aware of: 

- platform support (link people to Adopts platform support matrix, which is slightly different from OpenJ9)
- Add the pre-reqs (1. OpenSSL no longer bundled on OpenJDK 12, so if they
need it they must install it locally. 2. On OpenJDK8 you need a fontconfig package to fix
a gap between Oracle licensed fonts and those used at Adopt.)

### DIFF
--- a/docs/adoptopenjdk.md
+++ b/docs/adoptopenjdk.md
@@ -1,0 +1,28 @@
+
+
+# AdoptOpenJDK builds
+
+The AdoptOpenJDK community project provide pre-built binaries of OpenJDK with OpenJ9, which have been rigorously tested
+to meet expected quality standards. Binaries for the latest release of Eclipse OpenJ9 can be obtained from the
+following links:
+
+- [OpenJDK8 with OpenJ9](https://adoptopenjdk.net/releases.html?variant=openjdk8&jvmVariant=openj9)
+- [OpenJDK11 with OpenJ9](https://adoptopenjdk.net/releases.html?variant=openjdk11&jvmVariant=openj9)
+- [OpenJDK12 with OpenJ9](https://adoptopenjdk.net/releases.html?variant=openjdk12&jvmVariant=openj9)
+
+Nightly builds of OpenJDK with OpenJ9 are also available from the project.
+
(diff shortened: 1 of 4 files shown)


SueChaplain:
Jenkins doc stage

SueChaplain:
@pshipton - staged at https://eclipse.github.io/openj9-docs-staging/228/openj9_newuser/#installation-pre-requisites. Please merge if you are happy.

pshipton:
Can we make it more prominent and put it before "Command-line options"?

pshipton:
I was thinking this should be a new high level category in the doc, like `Introduction` and `New to OpenJ9?`, to help ensure it is seen. Was my previous comment misunderstood, or was it decided putting it in the `New to OpenJ9?` topic was sufficient?

SueChaplain:
@pshipton  - I did misunderstand your suggestion and was sort of OK with adding it to the New user topic.  Not a great fit, I agree. 

What we really need is space at the AdoptOpenJDK website to cover things like pre-reqs and point to their platform support statement. I've been trying to influence how the website could be changed to host more documentation, especially with things like IcedTea in mind. Whilst they've agreed in principle there is work to be done to agree with the steering committee and make it happen. 

In the meantime maybe we can have a topic **AdoptOpenJDK information** or **AdoptOpenJDK builds** where we can cover the pre-reqs, point to the platform support, point to the new to OJ9 topic etc .. and we can modify that later when the content we need at Adopt gets created. Would you be happy with that?

SueChaplain:
Jenkins doc stage

SueChaplain:
@pshipton - this is now staged here: https://eclipse.github.io/openj9-docs-staging/228/adoptopenjdk/

Still wondering whether it is in the right place. Looks odd to be before the Introduction ...thoughts?

pshipton:
Looks good. As long as it in the top three I'm happy.

SueChaplain:
Jenkins doc stage

Issue #225, opened 2019-02-28
Document jstack


pdbain-ibm:
**Issue or pull request number:**
https://github.com/eclipse/openj9/issues/4858

**Overview:**
*Describe the external change and the impact/benefit from a user perspective.*
This adds the `jstack` utility to OpenJ9 for compatibility with the reference implementation.

**Applies to the following JDK versions:**
8 and later / 9 and later

**Applies to the following platforms:**
All platforms
**For new command line options:**
Tentative command syntax:
`jstack <option>* <pid>*`
`<pid>*` is a list of process IDs; if none are supplied, the PID(s) are read from `stdin`. A user may then query all processes via `jps -q | jstack`. IDs of dead processes are silently ignored.
By default, `jstack` prints the Java stacks and thread information of the specified processes: this essentially the information in `java.lang.management.ThreadInfo`.

Options:

* `-p`: also print the process's system and agent properties
* The usual suspects: `-J`, `-h`, like `jps`.
- Restrictions:
	- This differs from the HotSpot version of `jstack` due to differences between the JVM implementations.  OpenJ9's `jstack` displays information about only local processes owned by the current user, due to security considerations in the interest of security.  A user may display remote processes using `ssh jstack ...`.  
	- Displaying data from core dumps is not supported:  `jdmpview` is preferred.  
	- Other options , such as `-F` (force a dump of an unresponsive) may be accomplished using `kill -QUIT <pid>`.  
	- Printing of native stack frames is not supported.

PR #246, opened 2019-04-03
Add jstack (8,11,12) and update jps (8,11)

SueChaplain:
- New tools topic for jstack
- New section for jstack in the release topic
- Updated to Tools "Switching to OpenJ9" to describe differences to
  HotSpot jstack

- Add jps to release topic for 8 & 11 (12 was in 0.13.0)
- Remove icons from jps content that specify Java 12 only.

Closes: #225

### DIFF
--- a/docs/tool_jstack.md
+++ b/docs/tool_jstack.md
@@ -1,0 +1,48 @@
+﻿<!--
+* Copyright (c) 2017, 2019 IBM Corp. and others
+*
+* This program and the accompanying materials are made
+* available under the terms of the Eclipse Public License 2.0
+* which accompanies this distribution and is available at
+* https://www.eclipse.org/legal/epl-2.0/ or the Apache
+* License, Version 2.0 which accompanies this distribution and
+* is available at https://www.apache.org/licenses/LICENSE-2.0.
+*
+* This Source Code may also be made available under the
+* following Secondary Licenses when the conditions for such
+* availability set forth in the Eclipse Public License, v. 2.0
+* are satisfied: GNU General Public License, version 2 with
+* the GNU Classpath Exception [1] and GNU General Public
+* License, version 2 with the OpenJDK Assembly Exception [2].
+*
+* [1] https://www.gnu.org/software/classpath/license.html
+* [2] http://openjdk.java.net/legal/assembly-exception.html
+*
+* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH
+* Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+-->
+
+# Java stack (`jstack`) tool
+
+Use the `jstack` tool to obtain Java stack traces and thread information for processes. The tool is similar to the HotSpot tool of the same name; the OpenJ9 version of `jstack` is an independent implementation, added for compatibility.
+
+The command syntax is as follows:
+
+    jstack <options>* <pid>*
+
+Where `<pid>*` is a list of process IDs. If none are supplied, the process IDs are read from `stdin`, which allows a user running a Bourne or equivalent shell to query all processes via `jps -q | jstack`. IDs of inactive processes are silently ignored. The output contains Java stacks and thread information of the specified processes (equivalent to the information provided in `java.lang.management.ThreadInfo`).
+
+The values for `<options>*` are as follows:
+
+- `-J`: supplies arguments to the Java VM that is running the `jstack` command. You can use multiple `-J` options, for example: `jstack -J-Xmx10m -J-Dcom.ibm.tools.attach.enable=yes`
+- `-p`: prints the system and agent properties of the process
+- `-l`: prints more verbose output, including information about locks
+- `-h`: prints help information
+
+
+<i class="fa fa-exclamation-triangle" aria-hidden="true"></i> **Restrictions:**
+
+- This tool is not supported and is subject to change or removal in future releases.
+- Although similar in usage and output to the HotSpot tool of the same name, this tool is a different implementation that is specific to OpenJ9. For more information about differences, see [Switching to OpenJ9](tool_migration.md).
+
+<!-- ==== END OF TOPIC ==== tool_jstack.md ==== -->
--- a/docs/tool_jps.md
+++ b/docs/tool_jps.md
@@ -22,7 +22,7 @@
 * Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
 -->
 
-# Java process status
+# Java process status (`jps`)
 
 ![Start of content that applies only to Java 12](cr/java12.png)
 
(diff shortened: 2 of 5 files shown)


SueChaplain:
Jenkins doc stage

pshipton:
Note there is a merge conflict.

SueChaplain:
Jenkins doc stage

docs/tool_jstack.md
```diff
@@ -0,0 +1,50 @@
...
+
+    jstack <options>* <pid>*
+
+Where `<pid>*` is a list of process IDs. If none are supplied, the process IDs are read from `stdin`, which allows a user to query all processes via `jps -q | jstack`. IDs of inactive processes are silently ignored. If this option is used on its own, the output contains Java stacks and thread information of the specified processes (equivalent to the information provided in `java.lang.managmenet.ThreadInfo`).
+
+The values for `<options>*` are as follows:
+
+- `-p`: Prints the system and agent properties of the process
+- `-J`: supplies arguments to the Java VM that is running the `jstack` command. You can use multiple `-J` options, for example: `jstack -J-Xmx10m -J-Dcom.ibm.tools.attach.enable=yes`
+- `-l`: prints the application package name
```
pshipton: According to the jstack help output, these are the only options supported
```
				+ "    -p: print the target's system and agent properties%n"
				+ "    -l: Long format. Print the thread's ownable synchronizers%n"
				+ "    -J: supply arguments to the Java VM running jps%n"
```

SueChaplain:
@pdbain-ibm @pshipton 

Please can this be reviewed your time today so that any changes I need to make can happen tomorrow, after which I am on vacation for a week.

I'm unsure about the list of options, because the issue says *The usual suspects: -J, -h, like jps.*

I've also added some references to Linux in there where I think options/syntax is not possible on Windows platforms. Might need some tweaks.

I've staged the doc but the latest updates are still building, so you'll need to wait for ghpages to refresh before checking: https://eclipse.github.io/openj9-docs-staging/246/tool_jstack/

I changed the title of the topic to "Java stack tool", because it doesn't look good in the table of contents, so when you see that change you'll know its updated.

pdbain-ibm:
@SueChaplain Thanks for working on this!

> I think options/syntax is not possible on Windows platforms

You could say "user __running a Bourne or equivalent shell__ to..."

Please drop these options:
```
-m: prints the application arguments
-v: prints the Java VM arguments, including those that are produced automatically
```
and add the same text for `-h` as `jps`.

`If this option is used on its own,`: I am not sure what you mean by this.  I don't believe it is necessary.

> I changed the title of the topic to "Java stack tool", because it doesn't look good in the table of contents, so when you see that change you'll know its updated.

Could we include the actual tool name table of contents, e.g. `Java stack tool (jstack)` or `jstack: Java stack tool`, or other variant of these (same for `jps`). This makes it clear we have these tools and helps search engines find them.

For reference:
https://docs.oracle.com/javase/7/docs/technotes/tools/share/jstack.html is titled `jstack - Stack Trace` and https://docs.oracle.com/javase/7/docs/technotes/tools/share/jps.html is titled `jps - Java Virtual Machine Process Status Tool`.  For reference also: https://docs.oracle.com/javase/7/docs/technotes/tools/

DanHeidinga:
@SueChaplain Oracle makes it very clear that the j* tools are 
> are unsupported and experimental, and should be used with that in mind. They may not be available in future JDK versions.

We should include similar restrictions in openj9 documentation for the these tools

docs/tool_jstack.md
```diff
@@ -0,0 +1,50 @@
...
+- `-q`: prints only the virtual machine identifiers
+- `-m`: prints the application arguments
+- `-v`: prints the Java VM arguments, including those that are produced automatically
+
+
+<i class="fa fa-exclamation-triangle" aria-hidden="true"></i> **Restrictions:**
```
pshipton: > We should include similar restrictions in openj9 documentation for the these tools

@DanHeidinga is this what you are looking for, or does it need to be clarified or stated in other places as well?

DanHeidinga: Yup, that's exactly what I was looking for.  Missed it on first read

SueChaplain:
@pdbain-ibm 

Some confusion over options still:

>add the same text for `-h` as jps

There is no `-h` option for `jps`. I assume you mean that you have added a `-h` to print help information, to make this compatible with `jstack`.  I added this.

>Please drop these options: (-m -v)

Done.

So, with these corrected we still have some discrepancies:

| HotSpot option | OpenJ9 option |
|-------------------|-----------------|
| -J | -J |
| -l | -l |
| -h | -h |
| -help | - |
| -F| not supported |
| -m| native info not supported|
|- | -p |
| -| -q |

You  gave us the `-p` option in your issue, but not `-q`, which got copied from `jps` because of the *all the usual suspects* comment. I assume we remove `-q` here (which I have done in my latest update)? Please confirm. Please also confirm that `-p` is correct.

Forgive me for not looking directly at the Oracle documentation when writing this up. We deliberately avoid doing so because of copyright infringement. It is therefore important that we have all the information directly from you.

>If this option is used on its own,: I am not sure what you mean by this. I don't believe it is necessary.

I meant if it is used as `jstack <pid*>` without adding any `<options>`. I removed the words anyway as the sentence is correct without them.

>Could we include the actual tool name table of contents

I played around with this a bit yesterday. I do agree with you. On their own they look odd. Although we try to avoid long TOC titles, we have longer ones, so I've added in *Full name (cmdname)* for `jstack` and `jps`. Hope that works for you.

SueChaplain:
Jenkins doc stage

docs/tool_jstack.md
```diff
@@ -0,0 +1,50 @@
...
+
+# The jstack tool
+
+Use the `jstack` tool to obtain Java stack traces and thread information for processes. The tool is similar to the HotSpot tool of the same name; the OpenJ9 version of `jstack` is an independent implementation, added for compatibility.
+
+The command syntax is as follows:
+
+    jstack <options>* <pid>*
+
+Where `<pid>*` is a list of process IDs. If none are supplied, the process IDs are read from `stdin`, which allows a user to query all processes via `jps -q | jstack`. IDs of inactive processes are silently ignored. If this option is used on its own, the output contains Java stacks and thread information of the specified processes (equivalent to the information provided in `java.lang.managmenet.ThreadInfo`).
```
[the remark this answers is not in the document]
SueChaplain: Fixed

SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

pdbain-ibm:
@SueChaplain 
Thanks for the updates. Reviewed and approved.

SueChaplain:
@pshipton - if you are also ok with the updates, please go ahead and merge.

SueChaplain:
@pshipton I'm out until after MS1, so if there are any other changes needed these will have to be done on top of my merge, or in a separate PR. Thanks!

Issue #269, opened 2019-05-15
OpenSSL MessageDigest is disabled (again) for the OpenJ9 0.14.2 release


pshipton:
**Issue or pull request number:**
https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/171

**Overview:**
OpenSSL MessageDigest is disabled in the OpenJ9 0.14.2 release. We'd done this and documented this before.
The system property jdk.nativeDigest cannot be used to enable MessageDigest.

**Release target:**
Eclipse OpenJ9 0.14.2

**Applies to the following JDK versions:**
jdk8 and jdk11 so far. Not sure if we will release a jdk12 update yet.

**Applies to the following platforms:**
All

Issue #260, opened 2019-04-29
OpenSSL 1.0.1 supported at runtime


groeges:
**Issue or pull request number:**
https://github.com/ibmruntimes/openj9-openjdk-jdk11/issues/159
this issue is for JDK11 but contains the PRs for all effected JDK versions.

**Overview:**
Support for the dynamic loading of the OpenSSL v1.0.1 library at runtime has been added to the OpenJDK Extensions for JDK8, JDK11, JDK12 and JDK next (13).  This support is only for runtime and the OpenSSL support is not able to be built using OpenSSL 1.0.1 (the build is still only supported with OpenSSL 1.0.2, 1.1.0 and 1.1.1)
   
**Release target:**
Eclipse OpenJ9 0.15.0 . (or whatever the next release is!!!)

**Applies to the following JDK versions:**
8 and later

**Applies to the following platforms:**
All platforms

**For new command line options:**
N/A

SueChaplain:
Confirmed that this is now targeted for 0.14.2.

PR #271, opened 2019-05-16
Add changes for OpenJ9 0.14.2 release

SueChaplain:
- Add OpenSSL 1.0.1 support
- Digest algorithm disabled by default

CLoses: #269
Closes: #260

### DIFF
--- a/docs/djdknativedigest.md
+++ b/docs/djdknativedigest.md
@@ -26,13 +26,15 @@
 
 This option enables or disables OpenSSL native cryptographic support for the Digest algorithm.
 
+<i class="fa fa-exclamation-triangle" aria-hidden="true"></i> **Restriction:** ![Start of content that applies to Java 8 (LTS)](cr/java8.png) ![Start of content that applies to Java 11 (LTS)](cr/java11.png) Due to issue [#5611](https://github.com/eclipse/openj9/issues/5611), the Digest algorithm is currently disabled. ![End of content that applies only to Java 8 and 11 (LTS)](cr/java_close_lts.png)
+
 ## Syntax
 
         -Djdk.nativeDigest=[true|false]
 
 | Setting              | value    | Default                                                                        |
 |----------------------|----------|:------------------------------------------------------------------------------:|
-| `-Djdk.nativeDigest` | true     | <i class="fa fa-check" aria-hidden="true"></i><span class="sr-only">yes</span> |
+| `-Djdk.nativeDigest` | true     | ![Start of content that applies to Java 12 (LTS)](cr/java12.png)<i class="fa fa-check" aria-hidden="true"></i><span class="sr-only">yes</span> |
 | `-Djdk.nativeDigest` | false    |                                                                                |
 
 ## Explanation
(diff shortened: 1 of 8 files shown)


SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

SueChaplain:
@pshipton @groeges - please review

docs/version0.14.2.md
```diff
@@ -0,0 +1,58 @@
...
+
+To learn more about support for OpenJ9 releases, including OpenJDK levels and platform support, see [Supported environments](openj9_support.md).
+
+### Support for OpenSSL 1.0.1
+
+OpenSSL version 1.0.1 support is now enabled; Earlier releases supported only OpenJDK 1.0.2 and 1.1.x. On Linux&reg; and AIX&reg; platforms, the OpenSSL libraries are expected to be available on the system path. For more information about cryptographic acceleration with OpenSSL, see [Cryptographic operations](introduction.md#cryptographic-operations).
+
+### OpenSSL Digest algorithm disabled
+
+Due to issue [#5611](https://github.com/eclipse/openj9/issues/5611), the Digest algorithm is disabled. As an alternative, the CBC, GCM, and RSA cryptographic algorithms can be used for cryptographic acceleration.
```
pshipton: CBC, GCM, and RSA are not alternatives to MessageDigest.

SueChaplain: Updated

docs/djdknativedigest.md
```diff
@@ -26,6 +26,8 @@
 
 This option enables or disables OpenSSL native cryptographic support for the Digest algorithm.
 
+<i class="fa fa-exclamation-triangle" aria-hidden="true"></i> **Restriction:** ![Start of content that applies to Java 8 (LTS)](cr/java8.png) ![Start of content that applies to Java 11 (LTS)](cr/java11.png) Due to issue [#5611](https://github.com/eclipse/openj9/issues/5611), the Digest algorithm is currently disabled. ![End of content that applies only to Java 8 and 11 (LTS)](cr/java_close_lts.png)
+
```
pshipton: Should the default in the chart be changed? It is somewhat misleading.

SueChaplain: I was assuming that the default applied to JDK 12? Is that not true?

pshipton: It is true the default applies to JDK 12, but it's not clear the table only applies to JDK 12. Also we expect very few people to be reading the docs for JDK 12 as the primary users are JDK 8 and 11.

SueChaplain:
Jenkins doc stage

docs/djdknativedigest.md
SueChaplain: I added an icon in the table. Does that help?
See staging: https://eclipse.github.io/openj9-docs-staging/271/djdknativedigest/

pshipton: Yes. Should `false` also have a checkmark and a tag for jdk 8 and 11?

SueChaplain: I assume for 8 & 11 the option has no effect. So would it be misleading to suggest that there is a default?

pshipton: I guess. There is a setting, which is false, and it cannot be modified.

pshipton: Let me know if you are done, and I will merge.

PR #274, opened 2019-05-22
Correct IcedTea-Web statement in release topic

SueChaplain:
IcedTea-Web is being packaged only with OpenJDK8 (64-bit).
Fix statement.


pshipton:
Not sure the decision at Adopt about whether to include IcedTea-Web in jdk11 installers has been finalized, so I haven't merged.

PR #307, opened 2019-06-27
Change docs to update -Xmso default on 64-bit z/OS and clarify behaviour

doveye:
See https://github.com/eclipse/openj9-docs/issues/295 and https://github.com/eclipse/openj9-docs/issues/247

[skip ci]


doveye:
Jenkins doc stage

doveye:
Jenkins doc stage

doveye:
Jenkins doc stage

doveye:
Jenkins doc stage

docs/xmso.md
```diff
@@ -24,12 +24,13 @@
...
-
-Sets the initial stack size for operating system threads.
+Sets the native stack size for operating system threads.
 
 You can use the `-verbose:sizes` option to find out the values that the VM is currently using.
 
-If you exceed the maximum value, a `java/lang/StackOverflowError` message is reported.
+When your application makes a JNI call, the VM calculates whether the memory required for the call will exceed the `-Xmso` value. If so, a `java/lang/StackOverflowError` error is thrown.
+
+<i class="fa fa-pencil-square-o" aria-hidden="true"></i> **Note:** This option relates to the native stack only. There is a separate, Java&trade; stack for operating system threads, for which you use different options (see the link in the [See also](#see-also) section).
```
pshipton: > There is a separate, Java&trade; stack for operating system threads

The "for operating system threads" part is incorrect, I expect it should be "There is a separate stack for Java threads".

docs/xmso.md
```diff
@@ -24,12 +24,13 @@
...
 # -Xmso
 
-
-Sets the initial stack size for operating system threads.
+Sets the native stack size for operating system threads.
 
 You can use the `-verbose:sizes` option to find out the values that the VM is currently using.
 
-If you exceed the maximum value, a `java/lang/StackOverflowError` message is reported.
+When your application makes a JNI call, the VM calculates whether the memory required for the call will exceed the `-Xmso` value. If so, a `java/lang/StackOverflowError` error is thrown.
```
pshipton: I think this needs more clarification. The memory requirements for the single call to JNI are calculated, but not what occurs inside the JNI call.

doveye: Dan reviewed in https://github.com/eclipse/openj9-docs/issues/247, but I'm happy to add any more clarification that's required (as long as someone explains it to me :) )

docs/xmso.md
doveye: Good spot! Fixed

docs/xmso.md
DanHeidinga: This isn't accurate.  We only check the stack size on call into the VM not on JNI call out.  We estimate whether the native stack is close to overflow on the callin and use that to determine if we should throw native stackoverflow.

We don't check when making a JNI call to avoid the performance cost.

docs/xmso.md
DanHeidinga: There is 1 thread - it operates either on a java stack or on a native stack.  Maybe phrasing this as:

```
Java methods and native methods run on two different stacks and the VM handles switching between them for JNI calls.  Each stack is sized using separate options.
```

docs/xss.md
```diff
@@ -24,14 +24,13 @@
...
-Determines the size and behavior of the stack size for Java&trade; threads. 
-
-To increase the maximum number of threads your system can support, reduce the maximum native stack size.
+Determines the size and behavior of the stack size for Java&trade; threads.
 
 If you exceed the maximum Java thread stack size, a `java/lang/OutOfMemoryError` message is reported.
 
 You can use the `-verbose:sizes` option to find out the values that the VM is currently using.
 
+<i class="fa fa-pencil-square-o" aria-hidden="true"></i> **Note:** These options relate to the Java stack only. There is a separate, native stack for operating system threads, for which you use a different option (see the link in the [See also](#see-also) section).
```
DanHeidinga: Same comment about 2 stacks here

docs/cmdline_migration.md
```diff
@@ -44,7 +44,7 @@ You can use the following command-line options in OpenJ9, just as you did in Hot
...
 | [`-Xnoclassgc`](xclassgc.md)                                     | Disables class garbage collection (GC).                                                                                                      |
 | [`-Xrs`](xrs.md)                                                 | Prevents the OpenJ9 run time environment from handling signals.                                                                              |
-| [`-Xss`](xss.md)                                                 | Sets the thread stack size. (Equivalent to `-XX:ThreadStackSize`)                                                                            |
+| [`-Xss`](xss.md)                                                 | Sets the Java&trade; thread stack size. (Equivalent to `-XX:ThreadStackSize`). <i class="fa fa-pencil-square-o" aria-hidden="true"></i> **Note:** Unlike HotSpot, this option applies only to Java threads. OpenJ9 has a separate native stack for operating system threads (see [`-Xmso`](xmso.md))                                                                      |
```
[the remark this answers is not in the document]
doveye: Done, thanks

docs/version0.14.md
```diff
@@ -87,6 +88,9 @@ By default, a network query is used to determine the host name and IP address fo
 
 On all platforms, the format of classes that are stored in the shared classes cache is changed, which causes the JVM to create a new shared classes cache, rather than re-creating or reusing an existing cache. To save space, all existing shared caches can be removed unless they are in use by an earlier release. For more information about destroying a shared classes cache, see [`-Xshareclasses`](xshareclasses.md).
 
+### Change to the default native stack size on 64-bit z/OS
+The default stack size for operating system threads on 64-bit z/OS is changed from 384 KB to 1 MB. For more information about this setting, see [`-Xmso`](xmso.md).
```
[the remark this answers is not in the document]
doveye: Done, thanks

docs/xss.md
doveye: A lot of the doc currently talks about java threads and native threads as separate entities. For example, in https://www.eclipse.org/openj9/docs/dump_javadump/#threads we say "Java threads, native threads, and stack traces" and even "A Java thread runs on a native thread." (We used to say "Each native thread is paired with the corresponding Java thread, if one exists.", but not anymore it seems!) Not sure if these are just different ways of looking at the same thing.

docs/xmso.md
doveye: I could change it to: 

> When a native method makes a call into the VM, the VM estimates whether the native stack is close to overflowing. If the native method call is likely to overflow the native stack, the VM throws a java/lang/StackOverflowError.

(Assuming we really mean "estimates" here, as opposed to "calculates", ie it's a bit of a guess). However, there's no reference to -Xmso. What part does that value actually play?

pshipton: If the native method call is likely to overflow the native stack, `as set by -Xmso`, the VM throws a java/lang/StackOverflowError.

doveye:
Jenkins doc stage

SueChaplain:
All updates complete, merging.

docs/xss.md
doveye: Made the suggested update, thanks

Issue #301, opened 2019-06-25
Document jmap


pshipton:
**Issue or pull request number:**
https://github.com/eclipse/openj9/pull/5538

**Overview:**
This adds the jmap utility to OpenJ9 for compatibility with the reference implementation.

**Release target:**
Eclipse OpenJ9 0.15.0

**Applies to the following JDK versions:**
All

**Applies to the following platforms:**
All

**For new command line options:**

Tentative command syntax:
`jmap <option>* <pid>*`
`<pid>*` is a list of process IDs; if none are supplied, the PID(s) are read from `stdin`. A user may then query all processes via `jps -q | jmap -histo`. IDs of dead processes are silently ignored.
By default, `jmap` prints help, a `-histo` argument must be supplied.

Options:

*    -histo Print statistics about classes on the heap, including number of objects and aggregate size
*    -histo:live  Print only live objects
* The usual suspects: `-J`, `-h`, like `jps`.

- Restrictions:
	- This differs from the HotSpot version of `jmap` due to differences between the JVM implementations.  OpenJ9's `jmap` displays information about only local processes owned by the current user, due to security considerations in the interest of security.  A user may display remote processes using `ssh jmap...`.  
	- Displaying data from core dumps is not supported:  `jdmpview` is preferred.  
	- Other options , such as `-F` (force a dump of an unresponsive) may be accomplished using `kill -QUIT <pid>`.

PR #313, opened 2019-07-03
New jmap tool

SueChaplain:
Topic to describe the use of the jmap tool, which is added
for compatibility with the reference implmentation.

Closes: #301

### DIFF
--- a/docs/tool_jmap.md
+++ b/docs/tool_jmap.md
@@ -1,0 +1,72 @@
+﻿<!--
+* Copyright (c) 2017, 2019 IBM Corp. and others
+*
+* This program and the accompanying materials are made
+* available under the terms of the Eclipse Public License 2.0
+* which accompanies this distribution and is available at
+* https://www.eclipse.org/legal/epl-2.0/ or the Apache
+* License, Version 2.0 which accompanies this distribution and
+* is available at https://www.apache.org/licenses/LICENSE-2.0.
+*
+* This Source Code may also be made available under the
+* following Secondary Licenses when the conditions for such
+* availability set forth in the Eclipse Public License, v. 2.0
+* are satisfied: GNU General Public License, version 2 with
+* the GNU Classpath Exception [1] and GNU General Public
+* License, version 2 with the OpenJDK Assembly Exception [2].
+*
+* [1] https://www.gnu.org/software/classpath/license.html
+* [2] http://openjdk.java.net/legal/assembly-exception.html
+*
+* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH
+* Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+-->
+
+# Java memory map (`jmap`) tool
+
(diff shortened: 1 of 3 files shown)


SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

SueChaplain:
@pdbain-ibm @pshipton 

I downloaded a nightly docker to try out jmap and answer questions like "how do you pass multiple IDs to the program". I noticed that the help for the tool is slightly different from the input in the issue, and aligns with the jps topic (references VMID instead of PS) so I made changes accordingly. I was also able to get some sample output, although I leave it to you as to whether this is suitable or needs updating.

Content is staged: https://eclipse.github.io/openj9-docs-staging/313/version0.15/#new-java-memory-map-tool

docs/tool_jmap.md
```diff
@@ -0,0 +1,72 @@
...
+  6            342           8208    java.lang.StringBuilder
+  7            151           7248    jdk.internal.org.objectweb.asm.Item
+  8            396           6336    [Ljava.lang.Object;
+```
+
+<i class="fa fa-exclamation-triangle" aria-hidden="true"></i> **Restrictions:** This tool is not supported and is subject to change or removal in future releases. Although similar in usage and output to the HotSpot tool of the same name, this tool is a different implementation that is specific to OpenJ9.
+
+The following tool limitations apply:
+
+- Displays information only for local processes that are owned by the current user, due to security considerations. You can display information for remote processes by  using `ssh jmap <options> <pid>`.
```
DanHeidinga:
```suggestion
- Displays information only for local processes that are owned by the current user, due to security considerations. You can display information for remote processes by  using `ssh user@host jmap <options> <pid>`.
```

docs/tool_jmap.md
```diff
@@ -0,0 +1,72 @@
...
+* [1] https://www.gnu.org/software/classpath/license.html
+* [2] http://openjdk.java.net/legal/assembly-exception.html
+*
+* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH
+* Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+-->
+
+# Java memory map (`jmap`) tool
+
+Use the `jmap` tool to get memory information for a particular Java&trade; process, or list of processes. The tool shows information for every Java process that is owned by the current user ID on the current host. The command syntax is as follows:
```
pshipton: Not sure about `The tool shows information for every Java process that is owned by the current user ID on the current host.`. Should this say the tool *can* show information...? This seems like it was copied from jps, but it doesn't apply to jmap.

SueChaplain: Hmm ... yes. This was started by someone else last week and I took over yesterday - assumed what was there so far was received as input, although I did strip out some things that didn't seem to apply. Will change to

>The tool shows statistics about classes on the heap, including the number of objects and their aggregate size.

docs/tool_jmap.md
SueChaplain: Added

Issue #308, opened 2019-06-28
Document how to create a smaller jlinked image when translations not required


pshipton:
When jlinking an image, if translations for non-English languages are not required, the translation files can be removed. 

https://github.com/eclipse/openj9/issues/4488#issuecomment-496279476

We should find a place to document this tip.

SueChaplain:
I've been trying to think about where to put this in the current doc set so that it can be easily found.

One idea: There is an issue on the website to move the build instructions into the user doc. If we did this, would it make sense to include `jlink` as another mechanism for creating a runtime image?

 Any other thoughts.
BTW - although we need to get this done, I am not prioritising it for 0.15.0 unless we have an easy solution.

pshipton:
How about adding a jlink section under Tools? We shouldn't document jlink itself, but can add the tips about using jlink with OpenJ9 there.

SueChaplain:
Tools has just moved into the Diagnostics section, so not sure that's appropriate. However, we could have a section in Getting Started -> Runtime Options -> Using Jlink?

SueChaplain:
Took a chance on Getting started being appropriate and created https://github.com/eclipse/openj9-docs/pull/318

PR #318, opened 2019-07-04
Include tip for using jlink

SueChaplain:
If translations from English are not required,
tell users how to exclude the translated property
files, which reduces the size of the image.

Closes: #308

### DIFF
--- a/docs/introduction.md
+++ b/docs/introduction.md
@@ -112,6 +112,11 @@ Runtime options are specified on the command line and include system properties,
 
 If you do not specify any options on the command line at run time, the OpenJ9 VM starts with default settings that define how it operates. For more information about these settings, see [Default settings for the OpenJ9 VM](openj9_defaults.md).
 
+## ![Start of content that applies to Java 11 (LTS) and later](cr/java11plus.png) Using Jlink
+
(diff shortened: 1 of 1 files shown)


SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

dinogun:
Minor update, the command to exclude files should be `--exclude-files=**java_**.properties` instead of `--exclude-files=**java_*.properties` (Notice the `**`).

I have a Dockerfile that works with the above change [here](https://github.com/dinogun/watson-springboot/blob/master/docker/Dockerfile.openj9.v11.jlink#L32). This is inline with what the [jlink](https://docs.oracle.com/javase/9/tools/jlink.htm#JSWOR-GUID-CECAC52B-CFEE-46CB-8166-F17A8E9280E9) documentation says.
@pshipton

SueChaplain:
Thanks @dinogun 

>Minor update, the command to exclude files should be --exclude-files=**java_**.properties instead of --exclude-files=**java_*.properties (Notice the **).

Fixed

SueChaplain:
@pshipton  - are you ok with these changes?

pshipton:
Yes, I was waiting to see if anyone else had comments.

PR #329, opened 2019-07-18
Add links to release binaries

SueChaplain:
- add missing links
- explain 0.15.1 naming at Adopt

### DIFF
--- a/docs/version0.15.md
+++ b/docs/version0.15.md
@@ -47,12 +47,14 @@
 
 ### Binaries and supported environments
 
- OpenJ9 release 0.15.0 supports OpenJDK 8, 11, and 12.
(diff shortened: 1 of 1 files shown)


SueChaplain:
@pshipton - are you happy with the explanation. If so, please merge

docs/version0.15.md
```diff
@@ -49,10 +49,12 @@
 
  OpenJ9 release 0.15.0 supports OpenJDK 8, 11, and 12.
```
pshipton: I think this should be updated as well. Either from 0.15.0 to 0.15.1, or include both numbers.

SueChaplain: Updated. Please merge and I will create release tags and publish

PR #350, opened 2019-08-29
Document -XX:[+|-]CompactStrings

### LINKED ISSUE #348: Document -XX:[+|-]CompactStrings
N/A

**Overview:**
Document the availability of the String compression feature in OpenJ9 which completes JEP 254 [1].

**Release target:**
Eclipse OpenJ9 0.8.0

8 and later

All platforms

- Option name and syntax: -XX[+|-]CompactStrings
- Purpose: Provides heap space savings by using an ISO8859-1 internal character representation for String objects while preserving full API compatibility and improving performance under certain conditions. This feature is in general beneficial when most, or all Strings the user application uses can be encoded using the ISO8859-1 character encoding.
- Default value: -XX:-CompactStrings

[1] https://openjdk.java.net/jeps/254

pmhayward:
Resolves #348

### DIFF
--- a/docs/cmdline_migration.md
+++ b/docs/cmdline_migration.md
@@ -44,11 +44,12 @@ You can use the following command-line options in OpenJ9, just as you did in Hot
 | [`-Xmx`](xms.md)                                                 | Specifies the maximum size of the object memory allocation pool. (Equivalent to `-XX:MaxHeapSize`)                                           |
 | [`-Xnoclassgc`](xclassgc.md)                                     | Disables class garbage collection (GC).                                                                                                      |
 | [`-Xrs`](xrs.md)                                                 | Prevents the OpenJ9 run time environment from handling signals.                                                                              |
-| [`-Xss`](xss.md)                                                 | Sets the Java&trade; thread stack size. (Equivalent to `-XX:ThreadStackSize`). <i class="fa fa-pencil-square-o" aria-hidden="true"></i> **Note:** Unlike HotSpot, this option applies only to the Java stack. OpenJ9 has a separate native stack for operating system threads (see [`-Xmso`](xmso.md))                                                                      |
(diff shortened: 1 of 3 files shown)


pmhayward:
Jenkins doc stage

pmhayward:
@fjeremic (fyi @DanHeidinga) Please confirm that this command is correctly documented.
A draft version is staged here:
https://eclipse.github.io/openj9-docs-staging/350/xxcompactstrings/
and also mentioned here, under "Compatible options":
https://eclipse.github.io/openj9-docs-staging/350/cmdline_migration/

pmhayward:
@fjeremic   - Filip, Happy to add your extra comments, but I'm a bit confused about "with OpenJ9 the user does not pay any performance penalty for disabling the feature since the JIT is tuned to optimize specifically for Compact Strings."
1. If there's no performance hit if disabling feature, why bother enabling it?! (Or does that imply that you're not including savings in heap space as part of "performance"?)
2. Not sure how the "since" works! - How does Jit being optimised for Compact Strings explain why there's no performance hit when Compact Strings aren't being used?
Thanks

fjeremic:
It's a matter of perspective, i.e. what your baseline is.

> 1. If there's no performance hit if disabling feature, why bother enabling it?! (Or does that imply that you're not including savings in heap space as part of "performance"?)

Enabling compact Strings will benefit via heap savings if most or all Strings in the application can be represented via the ISO8859-1 encoding. If this is not the case then we will have a mix of compressed Strings and decompressed Strings which is not ideal for performance.

> 2. Not sure how the "since" works! - How does Jit being optimised for Compact Strings explain why there's no performance hit when Compact Strings aren't being used?

Compact Strings are currently disabled by default, however the code to enable them is present. What is meant by "performance penalty for disabling" is that turning the feature off (as is the current default) has no performance implications on the application, even though for all String APIs we have to make checks all the time for "are compact Strings enabled?". The JIT optimizes away all such checks, so the application does not pay a performance penalty if it doesn't want compact Strings.

pmhayward:
Jenkins doc stage

pmhayward:
@fjeremic ... thanks for the explanation; I'm much clearer now! 
I've updated the topic here: https://eclipse.github.io/openj9-docs-staging/350/xxcompactstrings/
If you're OK with that, I'll merge the PR.

PR #354, opened 2019-09-05
No longer run on master for RELEASE builds

AdamBrousseau:
- Under the new Elcipse infra we cannot run on master
- Push directly to Eclipse from inside the container

Signed-off-by: Adam Brousseau <adam.brousseau88@gmail.com>

PR #365, opened 2019-09-19
Push to Eclipse outside of container

AdamBrousseau:
- There is an issue trying to push from within
  the container. The linux user id doesn't match
  because of how containers work. There were a few
  workaround solutions on SO that didn't fully
  resolve the issue for us. As a result we're now
  exiting the container before doing the push.

Related #363 #354

Signed-off-by: Adam Brousseau <adam.brousseau88@gmail.com>


AdamBrousseau:
Test build
https://ci.eclipse.org/openj9/view/Website-Doc/job/Build-Doc-Push_to_Eclipse/157/

PR #381, opened 2019-10-02
Add z/OS as a supported platform for Xdump allocation events

SueChaplain:
Current platform restriction for Linux, macOS and Windows
only. However, user queried whether this should also state
z/OS. Confirmed by service.

### DIFF
--- a/docs/xdump.md
+++ b/docs/xdump.md
@@ -564,7 +564,6 @@ You can filter the **slow** event to change the time threshold from the default 
 ```
 
 #### allocation event
-**(Linux, macOS, and Windows only)**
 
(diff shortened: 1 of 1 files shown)


SueChaplain:
@pshipton - we had someone test and confirm that this works on z/OS. If you are also happy, please merge. Thanks

docs/xdump.md
```diff
@@ -564,7 +564,8 @@ You can filter the **slow** event to change the time threshold from the default
 ```
 
 #### allocation event
-**(Linux, macOS, and Windows only)**
+
+**(Linux, macOS, Windows, and z/OS only)**
```
pshipton: It should be all platforms, it's not platform specific.

pshipton: Well, all platforms that use OpenJ9

SueChaplain: OK. Removed restriction completely. Should be ready to merge

PR #407, opened 2019-10-23
Update jextract on macOS

### LINKED ISSUE #268: Document jextract, in the tools section
The jextract tool still has a purpose: it collects a system dump, all the shared libraries (DLLs) that were loaded by the java process in the dump and the TraceFormat .dat files, and puts them into a zip file. This zip is often requested by support teams, so they can debug the system dump. The shared libraries (DLLs) help a debugger to format the stack traces captured in the system dump. It would be helpful if the jextract tool could be documented so that support teams can point users at the document, when asking for a system dump.

pmhayward:
Indicate that jextract is now available fopr all current
versions of OpenJDK running on macOS.

Update jextract and jdmpview topics, and add to a new version0.18 topic.

Closes #268

### DIFF
--- a/docs/version0.18.md
+++ b/docs/version0.18.md
@@ -1,0 +1,47 @@
+<!--
+* Copyright (c) 2017, 2019 IBM Corp. and others
+*
+* This program and the accompanying materials are made
+* available under the terms of the Eclipse Public License 2.0
+* which accompanies this distribution and is available at
+* https://www.eclipse.org/legal/epl-2.0/ or the Apache
+* License, Version 2.0 which accompanies this distribution and
+* is available at https://www.apache.org/licenses/LICENSE-2.0.
+*
+* This Source Code may also be made available under the
+* following Secondary Licenses when the conditions for such
+* availability set forth in the Eclipse Public License, v. 2.0
+* are satisfied: GNU General Public License, version 2 with
+* the GNU Classpath Exception [1] and GNU General Public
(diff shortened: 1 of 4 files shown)


pmhayward:
Jenkins doc stage

pmhayward:
Jenkins doc stage

pmhayward:
Jenkins doc stage

docs/version0.18.md
```diff
@@ -0,0 +1,47 @@
...
+- [jextract on macOS](#jextract-on-macos)
+- [Add more changes here...](#add-more-changes-here)
+
+
+## Features and changes
+
+### jextract on macOS
+
+The [`jextract` tool](tool_jextract.md) is now available on macOS&reg; platforms (as well as AIX&reg; and Linux&trade;) for all current versions of OpenJDK: 8, 11, and 13.
+
```
doveye: I think we need to clarify the bit that is actually new, which is jextract on macos for Java 8.

pmhayward: Fair point... I think the body of the text could stay as is, but perhaps change the header to:
### jextract now available on macOS for OpenJDK version 8

pmhayward:
Jenkins doc stage

docs/version0.18.md
doveye: Yes, that would be good - covers the new change and also the resulting overall status.

pmhayward: Changes made and restaged

Issue #409, opened 2019-10-23
Document -Xshareclasses:noTimestampChecks


pshipton:
-Xshareclasses:noTimestampChecks is not a documented change, however this blog https://wordpress.com/block-editor/post/blog.openj9.org/1901 talks about, therefore it should now be documented.

pshipton:
@hangshao0 can you please take this on.
Note https://github.com/eclipse/openj9-docs/issues/408

PR #410, opened 2019-10-23
Add documentation for -Xshareclasses:noTimestampChecks

hangshao0:
Fixes: #409

Signed-off-by: Hang Shao <hangshao@ca.ibm.com>

### DIFF
--- a/docs/xshareclasses.md
+++ b/docs/xshareclasses.md
@@ -410,6 +410,12 @@ case, the VM continues without using shared classes.
 
 :   Disables the storage of classes that are loaded by the bootstrap class loader in the shared classes cache. Often used with the `SharedClassURLFilter` API to control exactly which classes are cached. For more information about shared class filtering, see [Using the SharedClassHelper API](https://www.ibm.com/support/knowledgecenter/SSYKE2_8.0.0/com.ibm.java.vm.80.doc/docs/shrc_pd_helper_class.html).
 
+### `noTimestampChecks`
+
+        -Xshareclasses:noTimestampChecks
+
+:   Turns off timestamp checking when finding classes in the shared cache. Use this option only when you know there are no updates to the classes from the class paths or module paths in your application. Otherwise, obsolete classes might be loaded from the shared cache. If this happens, remove the `noTimestampChecks` option.
+
 ### `nocheckURLTimestamps`
 
         -Xshareclasses:nocheckURLTimestamps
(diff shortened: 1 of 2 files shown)


pmhayward:
Jenkins doc stage

docs/xshareclasses.md
```diff
@@ -353,6 +353,12 @@ case, the VM continues without using shared classes.
 
 :   Disables the storage of classes that are loaded by the bootstrap class loader in the shared classes cache. Often used with the `SharedClassURLFilter` API to control exactly which classes are cached. For more information about shared class filtering, see [Using the SharedClassHelper API](https://www.ibm.com/support/knowledgecenter/SSYKE2_8.0.0/com.ibm.java.vm.80.doc/docs/shrc_pd_helper_class.html).
 
+### `noTimestampChecks`
+
+        -Xshareclasses:noTimestampChecks
+
+:   Completely turn off timestamp checking when finding classes from the shared cache. Use this option only when you know there are no updates to the classes from the class paths or module paths in your application. Otherwise, obsolete classes may be loaded from the shared cache. If this scenario occurs, remove the `noTimestampChecks` option.
```
pmhayward: There are a few writing tweaks worth making here:

Saying "completely" might imply that somewhere, there's a "partly"! 
In line with other options, use "Turns" rather that "Turn"
I think you'd either "find" classes "in" the shared cache, or you'd "retrieve" them "from" the shared cache. Your choice!
So, something like: "Turns off timestamp checking when classes are retrieved from the shared cache..."

"May" tends to imply permission. In most cases, better to use either "can" (ability) or "might" (possibility). In this case, I think it's "might".

"If this scenario occurs..." might be better simplified to: "If this happens..."

---
This also needs an entry in the 0.18.0 "What's New" topic (version0.18.md). Add an entry just above "Add more changes here..." and follow the format that's already there. The entry needs to be short and should link here for more information as in " [-Xshareclasses:noTimestampChecks](xshareclasses.md#notimestampchecks) "
(It's OK to to refer to a feature being "new" in that topic, but avoid saying that elsewhere - think maintenance!)
Ensure that a link to this entry is added at the top of the topic - see way other entries are done.

hangshao0:
I've updated the change. Can you review again ? @pmhayward

doveye:
Jenkins doc stage

doveye:
@hangshao0  lgtm, apart from the URL in version0.18.md:

> For more information, see the -Xshareclasses:noTimestampChecks option.

This link is not working because URL should be all lowercase. You can only really check this sort of link after a doc build, because it relies on IDs that are generated by the system (always lowercase). See https://github.com/eclipse/openj9-docs/blob/master/CONTRIBUTING.md#user-content-previewing-pull-requests for how to do a draft doc build.

hangshao0:
> This link is not working because URL should be all lowercase ...

Tested locally, the link is fixed now.

PR #411, opened 2019-10-24
Updates for -Xmso 1 MB minimum on z/OS 64-bit

### LINKED ISSUE #396: -Xmso has 1MB minimum size on z/OS 64-bit
https://github.com/eclipse/openj9/pull/7304

**Overview:**
Even when the user requests a smaller size, the minimum -Xmso is set to 1MB since this is the size allocated by the /OS.

**Release target:**
Eclipse OpenJ9 0.18.0

All

z/OS (64-bit only)

pshipton:
Fixes #396

### DIFF
--- a/docs/openj9_defaults.md
+++ b/docs/openj9_defaults.md
@@ -81,7 +81,7 @@ The last 2 columns show whether the default setting can be changed by a command-
 |Maximum stack size for Java Threads **(32-bit)**. Use `-Xss<size>`  |320 KB    |320 KB    |N/A              |320 KB           |320 KB    |<i class="fa fa-check" aria-hidden="true"></i><span class="sr-only">yes</span>| <i class="fa fa-times" aria-hidden="true"></i><span class="sr-only">no</span> |
 |Maximum stack size for Java Threads **(64-bit)**. Use `-Xss<size>`  |1024 KB   |1024 KB   |1024 KB          |1024 KB          |1024 KB   |<i class="fa fa-check" aria-hidden="true"></i><span class="sr-only">yes</span>| <i class="fa fa-times" aria-hidden="true"></i><span class="sr-only">no</span> |
 |Stack size for OS Threads **(32-bit)**. Use `-Xmso<size>`           |256 KB    |256 KB    |N/A              |32 KB            |256 KB    |<i class="fa fa-check" aria-hidden="true"></i><span class="sr-only">yes</span>| <i class="fa fa-times" aria-hidden="true"></i><span class="sr-only">no</span> |
-|Stack size for OS Threads **(64-bit)**. Use `-Xmso<size>`           |256 KB    |256 KB    |256 KB           |256 KB           |1 MB    |<i class="fa fa-check" aria-hidden="true"></i><span class="sr-only">yes</span>| <i class="fa fa-times" aria-hidden="true"></i><span class="sr-only">no</span> |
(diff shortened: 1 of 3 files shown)


doveye:
Jenkins doc stage

docs/version0.18.md
```diff
@@ -37,6 +37,10 @@ The following new features and notable changes since v 0.17.0 are included in th
 
 The [`jextract` tool](tool_jextract.md) is now available on macOS&reg; platforms (as well as AIX&reg; and Linux&trade;) for _all_ current versions of OpenJDK: 8, 11, and 13.
 
+### `-Xmso` 1 MB minimum value on z/OS 64-bit
+
+On z/OS 64-bit, [`-Xmso`](xmso.md) has a 1 MB minimum value, to match the minimum stack space provided by the OS. Setting a value smaller than 1 MB is ignored.
+
```
doveye: - Need a link to this section from the top of the topic - you can use the jextract one as an example (currently there's an "Add more changes here" link that needs removing, so you could just change that). MkDocs generates html IDs for headings by making the heading text lowercase and adding dashes between words. So the ID for your section is -xmso-1-mb-minimum-value-on-zos-64-bit.
- Expand "OS" in "provided by the OS" to "operating system"
- For the last sentence, I suggest "If you set a value that is smaller than 1 MB, the value is ignored."

docs/openj9_defaults.md
```diff
@@ -81,15 +81,16 @@ The last 2 columns show whether the default setting can be changed by a command-
...
 
-
 <i class="fa fa-pencil-square-o" aria-hidden="true"></i> **Notes:**
 
-The default value of `-Xmx` depends on the version of Java.
+1. On z/OS 64-bit, [`-Xmso`](xmso.md) has a 1 MB minimum value, to match the minimum stack space provided by the OS. Setting a value smaller than 1 MB is ignored.
+
+2. The default value of [`-Xmx`](xms.md) depends on the version of Java.
 
 - ![Start of content that applies only to Java 8 (LTS)](cr/java8.png) The value is half the available memory with a minimum of 16MB and a maximum of 512 MB. ![End of content that applies only to Java 8 (LTS)](cr/java_close_lts.png)
```
doveye: I don't think this topic needs changing at all as the default value remains the same - this topic is just a reference topic to tell users what the defaults are for the whole VM. Sue suggested changing xmso.md, which I agree with - that's where users should go to find out about the behaviour of the option. In xmso.md, there is this sentence:

> Default values vary by platform. See Default settings for the OpenJ9 VM.

I suggest adding the following after that sentence:

> `<i class="fa fa-pencil-square-o" aria-hidden="true"></i> **Note:** On 64-bit z/OS, the default size is the minimum that can be allocated by the operating system. So if you set a value that is smaller, that value is ignored by the VM.`

(I've avoided mentioning the actual value here, so that it doesn't become out of date if the openj9_defaults.md topic changes in future).

pshipton: Done. I kept the link for -Xmx and aligned the columns in the text, just to make it easier when making future changes.

docs/version0.18.md
pshipton: Done

pshipton: I left in the "Add more changes here" because I think it helps avoid merge conflicts.

doveye:
Jenkins doc stage

docs/version0.18.md
doveye: lgtm, thanks

docs/openj9_defaults.md
doveye: lgtm, thanks

PR #436, opened 2019-11-25
Correct documentation regarding XcompilationThreads option

### LINKED ISSUE #435: JVM -XcompilationThreads option documentation incorrect?
**Page containing the error:**

https://www.eclipse.org/openj9/docs/xcompilationthreads

**Describe the error and the update needed:**

```
-XcompilationThreads<n>

where <n> is the number of threads, in the range 1-4 inclusive. A number greater than 4 prevents the Java™ VM from starting successfully. 
```

Seems incorrect, because `-XcompilationThreads7` works.

If `-XcompilationThreads<n>` is not specified, I see seven compilation threads (`JIT Compilation Thread-0`... `JIT Compilation Thread-6`). So `-XcompilationThreads7` seems to be the default (at least on my machine).

Or am I miss understanding something?

I am using:
```
openjdk version "11.0.5" 2019-10-15
OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.5+10)
Eclipse OpenJ9 VM AdoptOpenJDK (build openj9-0.17.0, JRE 11 Windows 10 amd64-64-Bit 20191031_114 (JIT enabled, AOT enabled)
OpenJ9   - 77c1cf708
OMR      - 20db4fbc
JCL      - 775c5eb03a based on jdk-11.0.5+10)
```

Thank you very much for providing OpenJ9!

mpirvu:
Existing documentation specifies that the number of compilation threads
needs to be in the 1-4 interval. Starting with Java8 we have increased the
maximum number of compilation threads to 7.
Also, the behavior has changed in that the JVM will no longer fail if the
user specifies an invalid number of compilation threads; instead it will use
the default number according to its internal heuristics.

Closes #435

Signed-off-by: Marius Pirvu <mpirvu@ca.ibm.com>

### DIFF
--- a/docs/xcompilationthreads.md
+++ b/docs/xcompilationthreads.md
@@ -31,7 +31,7 @@ Use this option to specify the number of compilation threads that are used by th
 
         -XcompilationThreads<n>
 
-: where `<n>` is the number of threads, in the range 1-4 inclusive. A number greater than 4 prevents the Java&trade; VM from starting successfully. 
+: where `<n>` is the number of threads, in the range 1-7 inclusive. Any number outside this range is ignored. 
 
 : Setting the compilation threads to zero does not prevent the JIT from working. Instead, if you do not want the JIT to work, use the [`-Xint`](xint.md) option.
 


docs/xcompilationthreads.md
```diff
@@ -31,7 +31,7 @@ Use this option to specify the number of compilation threads that are used by th
 
         -XcompilationThreads<n>
 
-: where `<n>` is the number of threads, in the range 1-4 inclusive. A number greater than 4 prevents the Java&trade; VM from starting successfully. 
+: where `<n>` is the number of threads, in the range 1-7 inclusive. Any number outside this range will be ignored.
```
SueChaplain: Please can you change the "will be" to "is" for consistency with the rest of the docs. Future tense should be avoided, where possible.

mpirvu: Done. Ready for another review.

PR #428, opened 2019-11-11
Docs Website Sidebar Logo Mobile View Visible

Bazif-Khan:
fixes:https://github.com/eclipse/openj9-docs/issues/426

PR #452, opened 2019-12-06
Attempt to fix large header on release docs

SueChaplain:
Last fix has minor diffs compared with master, which
do not cause a problem in local builds and
on their own do not look at all significant, but
re-applying files from master to overwrite those in
the 0.17.0 branch to ensure files are identical.

### DIFF
--- a/docs/stylesheets/oj9.css
+++ b/docs/stylesheets/oj9.css
@@ -81,7 +81,7 @@
 @media only screen and (max-width: 76.1875em) {
 
   html [data-md-color-primary=cyan] .md-nav--primary .md-nav__title--site {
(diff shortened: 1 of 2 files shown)


SueChaplain:
This PR shows that there are some diffs between the HEAD css and openj9-docs-0.17.0 CSS that could be the cause of this problem. Some changes were made in #428 for the mobile view to HEAD, which we were going to pick up in the next set of release docs. Part of this change affects the height/width of the logo. Local testing shows the mobile fixes in place. Header size looks fine, but then it did before.

SueChaplain:
Jenkins website stage

PR #462, opened 2020-01-02
Add further info for heap dumps

SueChaplain:
Detailed information for PHD and classic dumps
produced by the OpenJ9 VM dump component.


SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

SueChaplain:
Staged here: https://eclipse.github.io/openj9-docs-staging/462/dump_heapdump/

docs/dump_heapdump.md
```diff
@@ -24,25 +24,354 @@
...
+
+#### General structure
+
+The following structure comprises the header section of a PHD file:
+
+- A UTF string indicating that the file is a `portable heap dump`
+- An `int` containing the PHD version number
+- An `int` containing flags:
+    - `1` indicates that the `word` length is 64-bit.
+    - `2` indicates that all the objects in the dump are hashed. This flag is set for heap dumps that use 16-bit hash codes. These heap dumps use 32-bit hash codes that are only created when used. For example, these hash codes are created when the APIs `Object.hashCode()` or `Object.toString()` are called in a Java application. If this flag is not set, the presence of a hash code is indicated by the hash code flag on the individual PHD records.
```
pshipton: This is a bit confusing. I believe it's saying that flag 2 is obsolete, but this isn't explicitly mentioned.

docs/dump_heapdump.md
```diff
@@ -24,25 +24,354 @@
...
+	0x00000000E0000B00
+```
+
+The object instance (length 32 bytes) of type char array, as referenced from the `java/lang/String`, is shown in the following example:
+
+```
+0x00000000E0000B00 [32] OBJ [C
+```
+
+The following example shows an object instance (24 bytes in length) of type array of `java/lang/String`:
```
pshipton: The length 24 doesn't make sense, since there are 32 bytes of data shown.

SueChaplain: I thought the length seemed to correlate with the `[xx]` value, although 24 is very odd:

```
0x00000000FFF07498 [24] OBJ [Ljava/lang/String;
	0x00000000E0005D78 0x00000000E0005D50 0x00000000E0005D28 0x00000000E0005D00
```

What does the square bracket indicate?

docs/dump_heapdump.md
SueChaplain: Not just me then. I was going to ask you what it meant! Should I just replace the words with "obsolete"?

docs/dump_heapdump.md
pshipton: Either change the 24's to 32's, or remove one of the four array entries.
The `[` in `[Ljava/lang/String;` means it's an array of String objects.

pshipton: Hmm, there is obviously something I'm not understanding about the format, as the following is taken from a classic heap dump I just created. `(24 bytes in length)` doesn't match having 4 array entries of 8 bytes. I think my confusion is that although they are printed as 8 byte values, since this is compressed refs they are really 4 byte values. 
```
0x00000000FFDF1F60 [24] OBJ [Ljava/lang/String;
	0x00000000DFFF5790 0x00000000DFFF5768 0x00000000DFFF5740 0x00000000DFFF5718 
```

docs/dump_heapdump.md
pshipton: I don't like the `These` in `These heap dumps use ...`. I'd replace it with `Current` or something similar.

pshipton: Or `OpenJ9`

SueChaplain: Updated as suggested.

PR #471, opened 2020-01-14
Add JITServer tech preview

pmhayward:

### DIFF
--- a/docs/jitserver.md
+++ b/docs/jitserver.md
@@ -1,0 +1,91 @@
+<!--
+* Copyright (c) 2017, 2020 IBM Corp. and others
+*
+* This program and the accompanying materials are made
+* available under the terms of the Eclipse Public License 2.0
+* which accompanies this distribution and is available at
+* https://www.eclipse.org/legal/epl-2.0/ or the Apache
+* License, Version 2.0 which accompanies this distribution and
+* is available at https://www.apache.org/licenses/LICENSE-2.0.
+*
+* This Source Code may also be made available under the
+* following Secondary Licenses when the conditions for such
+* availability set forth in the Eclipse Public License, v. 2.0
+* are satisfied: GNU General Public License, version 2 with
+* the GNU Classpath Exception [1] and GNU General Public
+* License, version 2 with the OpenJDK Assembly Exception [2].
+*
+* [1] https://www.gnu.org/software/classpath/license.html
+* [2] http://openjdk.java.net/legal/assembly-exception.html
+*
+* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH
+* Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+-->
+
+# JITServer technology (technical preview)
+
+![Start of content that applies to Java 8 and later](cr/java8.png)
+![Start of content that applies to Java 11 and later](cr/java11.png)  
+**Linux&reg; on x86-64 only**
+
+JITServer technology decouples the JIT compiler from the VM and lets the JIT compiler run remotely in its own process. This mechanism prevents your Java&trade; application suffering possible negative effects due to CPU and memory consumption caused by JIT compilation.
+
+This technology can improve quality of service, robustness, and even performance of Java applications. We recommend trying this technology if the following criteria are met:
+
+- Your Java application is required to compile many methods using JIT in a relatively short time.
+- The application is running in an enviroment with limited CPU or memory, which can worsen interference from the JIT compiler.
+- The network latency between JITServer and client VM is relatively low.
+
+For more details about JITServer technology, its pros and cons, and when best to use it, see the blog [Free your JVM from the JIT with JITServer Technology](https://blog.openj9.org/2020/01/09/free-your-jvm-from-the-jit-with-jitserver-technology/).
+
+## Using JITServer technology
+
+JITServer technology is not enabled by default: you must explicitly invoke it. Running OpenJ9 without either of the following options launches it as a regular VM with embedded JIT compilation.
+
+### Launch OpenJ9 in client mode
+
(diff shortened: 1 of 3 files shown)


pmhayward:
Jenkins doc stage

pmhayward:
@mpirvu, @pshipton  ... the JITServer topic is staged at https://eclipse.github.io/openj9-docs-staging/471/jitserver/ for review. (Also in ["What's new"](https://eclipse.github.io/openj9-docs-staging/471/version0.18/#technical-preview-of-jitserver-technology))  
Couple of points for clarification, please:  
- Please check that I've understood correctly how JITServer is started in either client or server mode - as in first table and Notes 1&2
- I assume the timeout option `-XX:JITServerTimeout` is used to set default in either client or server mode and that it has different defaults depending on which mode its in

mpirvu:
The table on how the client and server are started is correct, though I am not sure it is clear.  There is a sentence before that: `To use JITServer technology, use one of the following options:`,
but then the first entry in the table refers to non-JITServer (just a vanilla OpenJ9 JVM). Maybe that first entry should be taken out of the table and have a comment saying that in the absence of any specific options the OpenJ9 will operate like a normal JVM which performs JIT compilations "in-process"

Yes, the default value or JITServerTimeout is different depending on the persona of the JVM, either client or server.

docs/jitserver.md
```diff
@@ -0,0 +1,88 @@
...
+* Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+-->
+
+# JITServer technology (technical preview)
+
+![Start of content that applies to Java 8 and later](cr/java8.png)
+![Start of content that applies to Java 11 and later](cr/java11.png)  
+**Linux&reg; on x86-64 only**
+
+JITServer technology decouples the JIT compiler from the VM and lets the JIT compiler run remotely in its own process. This prevents your Java&trade; application suffering possible negative effects due to CPU and memory consumption caused by JIT compilation.
```
SueChaplain: When I read this para I got thrown by the "This prevents...". Should we add a qualifier in here? I think we mean that the JITServer technology helps with resources. Maybe say "This capability.." or "This mechanism.."?

docs/jitserver.md
```diff
@@ -0,0 +1,88 @@
...
+
+- Your Java application is required to compile many methods using JIT in a relatively short time.
+- The application is running in an enviroment with limited CPU or memory, which can worsen interference from the JIT compiler.
+- The network latency between JITServer and client VM is relatively low.
+
+For more details about JITServer technology, its pros and cons, and when best to use it, see the blog [Free your JVM from the JIT with JITServer Technology](https://blog.openj9.org/2020/01/09/free-your-jvm-from-the-jit-with-jitserver-technology/).
+
+## Using JITServer technology
+
+OpenJ9 builds available from [AdoptOpenJDK](https://adoptopenjdk.net/archive.html?variant=openjdk11&jvmVariant=openj9) have JITServer technology included, though you must explicitly invoke it.
```
SueChaplain: Might be better to avoid talking about the AdoptOpenJDK binaries here and just say that it is not enabled by default. The topic is qualified as Linux on x86-64 and we know that all the related binaries at Adopt are building it in.

SueChaplain:
I'm wondering if the JITServer topic should be under JIT Compiler in the TOC rather than above it, with the standard JIT Compiler being the default?

docs/jitserver.md
```diff
@@ -0,0 +1,88 @@
...
+To use JITServer technology, use one of the following options:
+
+|  Setting            | Effect                                                             |
+|---------------------|--------------------------------------------------------------------|
+| none                | Launches OpenJ9 as a regular VM with embedded JIT compilation      |
+| `-XX:+UseJITServer` | Launches OpenJ9 in client mode (**Note 1**)                        |
+| `jitserver`         | Launches OpenJ9 in JITServer mode (**Note 2**)                     |
+
+**Note 1:** With this command-line option, the VM sends compilation requests to an available JITServer. The client operates as a regular VM if a server is not available.  
+**Note 2:** Use the `jitserver` command to start a VM that listens for incoming compilation requests.
```
pshipton: This shouldn't say "to start a VM". It starts the JIT server.

mpirvu: Implementation wise we are actually creating and starting a JVM, but it can be confusing. Maybe just say 'it starts a JITServer process that listens for incoming compilation requests'

pmhayward:
Jenkins doc stage

pmhayward:
Jenkins doc stage

pmhayward:
@mpirvu, @pshipton, @SueChaplain ... all above comments now resolved (thanks!) ... restaged at https://eclipse.github.io/openj9-docs-staging/471/jitserver/

SueChaplain:
LGTM +1  I'll give Marius and Peter an opportunity to comment before merging.

docs/jitserver.md
```diff
@@ -0,0 +1,91 @@
...
+
+# JITServer technology (technical preview)
+
+![Start of content that applies to Java 8 and later](cr/java8.png)
+![Start of content that applies to Java 11 and later](cr/java11.png)  
+**Linux&reg; on x86-64 only**
+
+JITServer technology decouples the JIT compiler from the VM and lets the JIT compiler run remotely in its own process. This mechanism prevents your Java&trade; application suffering possible negative effects due to CPU and memory consumption caused by JIT compilation.
+
+This technology can potentially improve quality of service, robustness, and even performance of Java applications. We recommend trying this technology if the following criteria are met:
```
pshipton: I'd remove "potentially". "This technology can improve quality of service, ...". In this context "can" means it's possible. The following information explains when to use it and the blog contains more information.

pmhayward:
Jenkins doc stage

pmhayward:
Jenkins doc stage

PR #488, opened 2020-01-27
Document options not supported by OPENJ9_JAVA_OPTIONS

### LINKED ISSUE #487: Document options that don't work in the environment variables
Some options don't work when specified in the environment variables, OPENJ9_JAVA_OPTIONS, JAVA_TOOL_OPTIONS, see https://www.eclipse.org/openj9/docs/env_var/, and this is not properly documented. There is a list in https://www.eclipse.org/openj9/docs/xoptionsfile/ but it's missing `-XshowSettings`.

pshipton:
Refer to -Xoptionsfile to document which options are not supported by
the OPENJ9_JAVA_OPTIONS environment variable. Add `-XshowSettings` to
the list of unsupported options.

Fixes #487


SueChaplain:
Jenkins doc stage

SueChaplain:
@pshipton  - this is almost ready to merge. I had one outstanding question above.

pshipton:
@SueChaplain I don't see the question.

SueChaplain:
@pshipton  - OK not quite a question:

>If this option is supported by OpenJ9 when used on its own it should be recorded in the command reference. It is currently missing.

Do we need to update the -X command reference section with it?

pshipton:
> Do we need to update the -X command reference section with it?

I assume you are referring to `-XshowSettings`. This is an OpenJDK option, up to you if you want to document it.

SueChaplain:
>I assume you are referring to -XshowSettings. This is an OpenJDK option, up to you if you want to document it.

Had it been a HotSpot option, yes, but in this case, no :-) Thanks for confirming.

PR #495, opened 2020-01-29
Fix redirects

SueChaplain:
Replace redirected URLs with the final URL to avoid
these being flagged repeatedly in test results.


SueChaplain:
@pmhayward - most of these are switches from http -> https. If others are not obvious, feel free to challenge me :-)

PR #498, opened 2020-01-30
Document -Djava.lang.stringBuffer.growAggressively

### LINKED ISSUE #497: Document -Djava.lang.stringBuffer.growAggressively
https://github.com/eclipse/openj9/pull/8467

**Overview:**
See the PR.

**Release target:**
0.19

Java 8

all

pshipton:
Fixes #497

### DIFF
--- a/docs/djavalangstringbuffergrowaggressively.md
+++ b/docs/djavalangstringbuffergrowaggressively.md
@@ -1,0 +1,43 @@
+<!--
+* Copyright (c) 2020, 2020 IBM Corp. and others
+*
+* This program and the accompanying materials are made
+* available under the terms of the Eclipse Public License 2.0
+* which accompanies this distribution and is available at
+* https://www.eclipse.org/legal/epl-2.0/ or the Apache
+* License, Version 2.0 which accompanies this distribution and
+* is available at https://www.apache.org/licenses/LICENSE-2.0.
+*
+* This Source Code may also be made available under the
+* following Secondary Licenses when the conditions for such
+* availability set forth in the Eclipse Public License, v. 2.0
+* are satisfied: GNU General Public License, version 2 with
+* the GNU Classpath Exception [1] and GNU General Public
+* License, version 2 with the OpenJDK Assembly Exception [2].
+*
+* [1] https://www.gnu.org/software/classpath/license.html
+* [2] http://openjdk.java.net/legal/assembly-exception.html
+*
+* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH
+* Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+-->
+
+# -Djava.lang.stringBuffer.growAggressively
+
+<i class="fa fa-exclamation-triangle" aria-hidden="true"></i> **Restriction:** This system property is supported only on Java&trade; 8.
+
+![Start of content that applies only to Java 8 (LTS)](cr/java8.png) Setting this property to `false` reverts to the behavior (OpenJ9 0.18 and earlier) of growing a 1 G `char[]` or larger `StringBuffer` or `StringBuilder` only as much as necessary to accommodate the `String` being added. The default behavior is to immediately grow to the maximum possible size, similarly to Java 11 and later. The default behavior is compatible with the Oracle Hotspot VM.
+
+
+
+## Syntax
+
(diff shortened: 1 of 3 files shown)


SueChaplain:
Jenkins doc stage

SueChaplain:
@pshipton - this property is so long that it is a challenge for the doc build, see https://eclipse.github.io/openj9-docs-staging/498/djavalangstringbufferandbuildergrowaggressively/ .... any chance we can shorten it? Maybe -Djavalangstringbufferandbuildergrowmax?

Also, although I  see some lack of consistency in other topics (which we will fix), we should force API / classes etc into monospace so that it is clear they are code terms and not words. So, `StringBuffer`, `StringBuilder` etc. Please can you update?

pshipton:
Renamed the option to java.lang.stringBuffer.growAggressively in https://github.com/eclipse/openj9/pull/8484

SueChaplain:
Jenkins doc stage

pshipton:
Just fyi the associated code is delivered now.

docs/version0.19.md
```diff
@@ -44,6 +45,10 @@ The latest builds of OpenJDK with OpenJ9 for Java 8 and 11 at the AdoptOpenJDK c
 
 To learn more about support for OpenJ9 releases, including OpenJDK levels and platform support, see [Supported environments](openj9_support.md).
 
+### stringbuffer-and-stringbuilder-above-1-g-grow-to-the-maximum-size
```
pmhayward: - I think you copied the link string here, rather than the title!
- Please add `Start of content that applies to Java 8` at beginning of both titles (doesn't affect link string). See last item in Version 16 release notes for an example.

docs/djavalangstringbuffergrowaggressively.md
```diff
@@ -0,0 +1,43 @@
...
+*
+* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH
+* Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+-->
+
+# -Djava.lang.stringBuffer.growAggressively
+
+<i class="fa fa-exclamation-triangle" aria-hidden="true"></i> **Restriction:** This system property is supported only on Java&trade; 8.
+
+![Start of content that applies only to Java 8 (LTS)](cr/java8.png) Setting this property to `false` reverts to the older behavior (OpenJ9 0.18 and earlier) of growing a 1 G `char[]` or larger `StringBuffer` or `StringBuilder` only as much as necessary to accommodate the `String` being added. The new behavior is to immediately grow to the maximum possible size, similarly to Java 11 and later. The new behavior is compatible with the Oracle Hotspot VM.
```
pmhayward: We should always avoid referring to 'old' and 'new' in the docs (except in "What's New" topics!) as the text would need changing at the next release - a maintenance nightmare! So we just document behavior 'as is'. Suggest something like this:

By default, a 1 G `char[]` or larger `StringBuffer` or `StringBuilder` immediately grows to the maximum possible size, similarly to Java 11 and later. Set this property to `false` to limit growth to only as much as necessary to accommodate the `String` being added. The default behavior is compatible with the Oracle Hotspot VM.

pshipton: Updated. I kept it basically the same but removed old and new.

docs/version0.19.md
pshipton: Fixed.

pmhayward:
Jenkins doc stage

pmhayward:
Thanks for making changes. Just one query... 
In relation to a `char[]`, it refers  to "1 G".
Is that 1 GB of _data_ (i.e. either 10^9 or 2^30 bytes?)...
Or is it 1 G (10^9) of _characters_ - which I assume would be 2 GB of _data_, if they're 2 byte characters?
I know I'm not writing for IBM any more(!), but I know that IBM guidance is not to use a unit prefix on its own. 
...or am I just being far too pedantic?! :)

pshipton:
It's 1G of char data, which is 2GB, but the important part is the 1G elements. If you have a suggestion for how to write it better pls let me know.

pmhayward:
@pshipton ... Apparently the guidelines were updated about three months ago to allow for situations like 'a 1G char[]'. So that's all good then! Though JFYI, only proviso is that it's '1G' (no space) but '1 GB' (with a space)... no, I didn't make the rule :)
Will merge this PR now and fix the spaces along with some other tidying up later.

PR #502, opened 2020-02-04
Move -Djava.lang.string.substring.nocopy out of compatible options

### LINKED ISSUE #501: -Djava.lang.string.substring.nocopy=true not a Hotspot compatible option
https://www.eclipse.org/openj9/docs/cmdline_migration/

Just realized that `-Djava.lang.string.substring.nocopy=true` doesn't belong in the table of compatible options. It's an option that should be set for Hotspot compatibility. We'll need to add another table for this.

pshipton:
Fixes #501

### DIFF
--- a/docs/openj9_newuser.md
+++ b/docs/openj9_newuser.md
@@ -31,7 +31,7 @@ The Eclipse OpenJ9 virtual machine (VM) implements the [Java Virtual Machine Spe
 
 Although OpenJ9 implements its own command-line interface, many HotSpot options are recognized and accepted by the VM for compatibility. Any `-XX:` options that are not recognized by the VM are ignored by default, which prevents an application failing to start. You can turn off this behavior with the [-XX:-IgnoreUnrecognizedXXColonOptions](xxignoreunrecognizedxxcolonoptions.md) option.
 
-For a list of compatible options, see [Switching to OpenJ9](cmdline_migration.md) in the Command-line options section.
+For a list of compatible options, equivalent options, and options that need to be set for compatibility, see [Switching to OpenJ9](cmdline_migration.md).
 
 
 ## Garbage collection policies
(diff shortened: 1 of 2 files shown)


pmhayward:
Jenkins doc stage

docs/openj9_newuser.md
```diff
@@ -31,6 +31,12 @@ The Eclipse OpenJ9 virtual machine (VM) implements the [Java Virtual Machine Spe
 
 Although OpenJ9 implements its own command-line interface, many HotSpot options are recognized and accepted by the VM for compatibility. Any `-XX:` options that are not recognized by the VM are ignored by default, which prevents an application failing to start. You can turn off this behavior with the [-XX:-IgnoreUnrecognizedXXColonOptions](xxignoreunrecognizedxxcolonoptions.md) option.
 
+Set the following options for compatibility with Hotspot.
```
pmhayward: After some discussion, the ID team decided that this this new table should preferably live in the "Switching to OpenJ9" (`cmdline_migration.md`) topic, at the end in a new section, something like this:
**Creating compatible behavior**
You can set the following options to make OpenJ9 behave in the same way as Hotspot.

Looking at that topic, I wondered if the `-XX:+UseNUMA` option currently in the "Equivalent options" table might arguably belong in this new table - isn't it an option you have to set to get the same default behavior as Hotspot?

pshipton: I've moved the table and added -Xnuma:none to the list.

pmhayward:
Jenkins doc stage

pmhayward:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

pmhayward:
Jenkins doc stage

docs/openj9_newuser.md
pmhayward: Thanks - all looks good

PR #476, opened 2020-01-16
Update Balanced GC documentation

bragaigor:
Include more detailed explanation on Balanced GC policy
and add information about arraylets and double mapping

Related to issue: https://github.com/eclipse/openj9/issues/8349
Related to issue: https://github.com/eclipse/openj9-docs/issues/475

Signed-off-by: Igor Braga <higorb1@gmail.com>

Issue #475, opened 2020-01-15
Allow double mapping by default on Linux Platforms


bragaigor:
**Issue or pull request number:**
*https://github.com/eclipse/openj9-docs/pull/476*

**Overview:**
*Large arrays in balanced garbage collection gets allocated discontiguously (in different regions). Double mapping allows the GC to have a contiguous view of the array without having to rely on iterators. One example of double mapping's benefits is JNI critical, which requires a contiguous view of an array. Without double mapping JNI critical first needs to copy element by element the entire array to a temporary one; second, perform its operation and third, copy all the elements back to the original array. This is extremely expensive. With double mapping, JNI critical can perform its operation on the contiguous representation of the array where its changes gets reflected on the original discontiguous array, without the need to copy any element. For such operations double mapping proved to be 20 times faster on average. For more information please refer to the blog post: https://eclipse-omr.org/double-map-arraylets/*

**Current limitations:**
*Double mapping does not support systems that has huge pages as default (normally 2MB). In such systems if double mapping is requested, it will get disabled. Please see: https://github.com/eclipse/openj9/pull/8254*

**Release target:**
*Eclipse OpenJ9 ~~0.18.0 0.19.0 0.20.0 0.21.0~~*

**Applies to the following JDK versions:**
*8 and later*

**Applies to the following platforms:**
*Linux X|P*

**For new command line options:**
- Option name and syntax: 
*-XXgc:enableArrayletDoubleMapping*
*-XXgc:disableArrayletDoubleMapping*

- Default value: *-XXgc:enableArrayletDoubleMapping*

pshipton:
It's too late for the 0.18.0 release. If the change is delivered by Feb 9 it can be in the 0.19.0 release, although this release is for Java 14 only.

amicic:
We can probably wait for next Java 8/11. Is that 0.20, and when?

pshipton:
0.20 ships mid April. We haven't set a branch date yet, but tentatively March 8.

DanHeidinga:
> We can probably wait for next Java 8/11. Is that 0.20, and when?

Why would we hold the code for that?  I'd rather get it out to users so they can benefit from it and we can get more feedback on it

amicic:
I'd rather soak it for several weeks (when is 0.20) in nightly and perf testing. There are techniques and OS APIs that we have not relied on in past - there is potential risk. 
[...]

PR #514, opened 2020-02-20
Tidy information about Balanced GC

pmhayward:
Edit information added to gc.md about Balanced GC, arraylets, and double mapping. 
Fix links. 
Replace first diagram in this section
Add entry in 0.19 "What's new"

References #475, #476


pmhayward:
Jenkins doc stage

amicic:
Enabling double mapping by default has been deferred. It will not be a part of 0.19. We don't need to document it now.

Some general documentation about Balanced GC that does not talk about double mapping could go in, but it's probably easier to just defer this update altogether.

SueChaplain:
That's bad news @amicic and @bragaigor - a lot of this stuff already went in. We thought it was all enabled, so merged Igor's updates. This PR is to do some further editing on the input.

amicic:
We were not sure up until the last moment of code split if it will be enabled or not. Don't know if this was early enough, but I tried to notify Peter that deferral is likely to happen here: https://github.com/eclipse/openj9/issues/8349#issuecomment-583448629

SueChaplain:
NP  - we can merge and hide it until 0.20.0. Leaving with Peter to address

Issue #558, opened 2020-04-29
Linux on ARM 64-bit uses gcc 7.5


pshipton:
**Page containing the error:**

https://www.eclipse.org/openj9/docs/openj9_support/

**Describe the error and the update needed:**

It's documented as gcc 7.4 but is actually 7.5

PR #557, opened 2020-04-24
Remove support for Windows 8.1, Server 2012, but not R2

pshipton:
Closes #556

Signed-off-by: Peter Shipton <Peter_Shipton@ca.ibm.com>

PR #560, opened 2020-04-29
Update the 0.20.0 branch support statement

pshipton:
Cherry pick #557 - Remove support for Windows 8.1, Server 2012, but not R2
and #558 - Linux on ARM 64-bit uses gcc 7.5
for the 0.20.0 branch.


SueChaplain:
LGTM

PR #568, opened 2020-05-21
Note update to NoClassDefFoundError exception message

sharon-wang:
Update What's New for 0.21 to include `NoClassDefFoundError` exception message update

Fixes: https://github.com/eclipse/openj9-docs/issues/563
Related: https://github.com/eclipse/openj9/pull/9588

### DIFF
--- a/docs/version0.21.md
+++ b/docs/version0.21.md
@@ -29,6 +29,7 @@ The following new features and notable changes since v 0.20.0 are included in th
 
 - [New binaries and changes to supported environments](#binaries-and-supported-environments)
 - [New feature...](#feature-title)
+- [Update to `NoClassDefFoundError` exception message](#update-to-noclassdeffounderror-exception-message)
 
 
 ## Features and changes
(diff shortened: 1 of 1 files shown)


sharon-wang:
@pshipton Should more detail be provided in this note?

pshipton:
We can include an example, and describe the meaning of each name in the message.

sharon-wang:
@pshipton I added a simple example and noted that one class name is the "name encountered by the VM" and the other class name is the "wrong name provided to the VM". I can try to reword the description if it's a bit confusing.

docs/version0.21.md
```diff
@@ -48,6 +49,30 @@ To learn more about support for OpenJ9 releases, including OpenJDK levels and pl
...
+		System.out.println("Hello!");	
+	}
+}
+```
+Compiling and running this program in the `mypackage` directory will result in the following error:
+```
+root@mycomputer:~/path/to/mypackage javac Main.java
+root@mycomputer:~/path/to/mypackage java Main
+Error: Could not find or load main class Main
+Caused by: java.lang.NoClassDefFoundError: mypackage/Main (wrong name: Main)
```
pshipton, line 72: By an example, I just meant this line and the following paragraph. I don't think the code example to recreate it is necessary, but it doesn't need to be removed unless the docs team thinks it's too much for a What's New.

docs/version0.21.md
```diff
@@ -48,6 +49,30 @@ To learn more about support for OpenJ9 releases, including OpenJDK levels and pl
...
+```
+Where `mypackage/Main` is the class name encountered by the VM, but "wrong name" `Main` was the provided class name. Prior to this update to the exception message, the encountered class name and the provided class name were swapped in the `NoClassDefFoundError` exception message.
```
pshipton, line 74: could this say something like `encountered by the VM in the .class file`.

sharon-wang:
@pshipton I moved the long example to my openj9 [PR](https://github.com/eclipse/openj9/pull/9588) description and updated the explanation to include `encountered by the VM in the .class file`.

sharon-wang:
@pmhayward Hi Peter, this is ready for review.

pmhayward:
Jenkins doc stage

pmhayward:
lgtm... thanks!

PR #578, opened 2020-06-16
Document -XX:[+/-]PrintFlagsFinal

### LINKED ISSUE #571: Document -XX:[+/-]PrintFlagsFinal
https://github.com/eclipse/openj9/pull/9567
https://github.com/eclipse/openj9/pull/9725

**Overview:**
OpenJ9 adds minimal support for -XX:+PrintFlagsFinal

**Release target:**
Eclipse OpenJ9 0.21.0

all

all

-XX:[+/-]PrintFlagsFinal

pmhayward:
Create new xxprintflagsfinal.md topic and add entry to version0.21.md
Resolves #571

### DIFF
--- a/docs/version0.21.md
+++ b/docs/version0.21.md
@@ -28,9 +28,10 @@
 The following new features and notable changes since v 0.20.0 are included in this release:
 
 - [New binaries and changes to supported environments](#binaries-and-supported-environments)
-- [New `-XX:[+|-]HandleSIGABRT` option added](#new--xx-handlesigabrt-option-added)
+- [New `-XX:[+|-]HandleSIGABRT` option added](#new-xx-handlesigabrt-option-added)
 - [Update to `NoClassDefFoundError` exception message](#update-to-noclassdeffounderror-exception-message)
 - [OSX shared libraries version updated](#osx-shared-libraries-version-updated)
+- [New `-XX:[+|-]PrintFlagsFinal` option added](#new-xx-printflagsfinal-option-added)
 
 ## Features and changes
 
@@ -61,6 +62,10 @@ java.lang.NoClassDefFoundError: mypackage/Main (wrong name: Main)
 ### OSX shared libraries version updated
 
 The version information for shared libraries on OSX has been updated from 0.0.0 to 1.0.0. If an application has linked against a shared library from a previous OpenJ9 release, it needs to be re-linked against the new release. Failure to re-link causes an error `Incompatible library version`, `requires version 0.0.0`.
+
+### New `-XX:[+|-]PrintFlagsFinal` option added
+
+This release provides an initial implementation of the `-XX:[+|-]PrintFlagsFinal` option. It is currently incomplete and outputs only a subset of parameters. Over time, we expect more options to be added to the output. See [`-XX:[+|-]PrintFlagsFinal`](xxprintflagsfinal.md) for more details about this option.
 
 ## Full release information
 
--- a/docs/xxprintflagsfinal.md
+++ b/docs/xxprintflagsfinal.md
@@ -1,0 +1,50 @@
+<!--
+* Copyright (c) 2017, 2020 IBM Corp. and others
+*
+* This program and the accompanying materials are made
+* available under the terms of the Eclipse Public License 2.0
+* which accompanies this distribution and is available at
+* https://www.eclipse.org/legal/epl-2.0/ or the Apache
+* License, Version 2.0 which accompanies this distribution and
+* is available at https://www.apache.org/licenses/LICENSE-2.0.
+*
+* This Source Code may also be made available under the
+* following Secondary Licenses when the conditions for such
+* availability set forth in the Eclipse Public License, v. 2.0
+* are satisfied: GNU General Public License, version 2 with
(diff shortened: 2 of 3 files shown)


pmhayward:
Jenkins doc stage

docs/version0.21.md
```diff
@@ -62,6 +63,9 @@ java.lang.NoClassDefFoundError: mypackage/Main (wrong name: Main)
 
 The version information for shared libraries on OSX has been updated from 0.0.0 to 1.0.0. If an application has linked against a shared library from a previous OpenJ9 release, it needs to be re-linked against the new release. Failure to re-link causes an error `Incompatible library version`, `requires version 0.0.0`.
 
+### New `-XX:[+|-]PrintFlagsFinal` option added
+This Hotspot option outputs the values of all configuration parameters. See [`-XX:[+|-]PrintFlagsFinal`](xxprintflagsfinal.md) for more details about this option.
```
mstoodle, line 67: I think we should be clear that this option is incomplete at the moment, and falls far short of "outputs the values of all configuration parameters".

How about adding: "With this release, an initial implementation of this option is incomplete and outputs only a subset of parameters. Over time, we expect more options to be added to the output."

docs/xxprintflagsfinal.md
```diff
@@ -0,0 +1,40 @@
...
+* [1] https://www.gnu.org/software/classpath/license.html
+* [2] http://openjdk.java.net/legal/assembly-exception.html
+*
+* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH
+* Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+-->
+
+# -XX:\[+|-\]PrintFlagsFinal
+
+When enabled, this Hotspot option outputs the values of all configuration parameters.
```
mstoodle, line 27: We should probably also be clear here that only a small subset of options are currently being output by this option. I'm not sure we want to highlight just how minimal the subset is, though (only two options are currently displayed). Otherwise we'll have to keep updating this document if we add new options. Perhaps just mentioning that this option shows only a subset of options found to be necessary, and that new options will be added as we learn of new requirements, or something like that?

pmhayward:
Jenkins doc stage

pmhayward:
Thanks for comments @mstoodle on my 'bare bones' topic - I was just about to ask for some input!
I've changed it in line with your suggestions, but have avoided saying anything in the topic itself about future plans or anything like "this is new in this release" - it all becomes a maintenance nightmare! The "What's New" topic is best for that, I think.
Can you confirm what the correct default value is. (50:50 guess without rummaging in the code!)
Anything else that needs saying?

docs/xxprintflagsfinal.md
```diff
@@ -0,0 +1,40 @@
...
+* [1] https://www.gnu.org/software/classpath/license.html
+* [2] http://openjdk.java.net/legal/assembly-exception.html
+*
+* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH
+* Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+-->
+
+# -XX:\[+|-\]PrintFlagsFinal
+
+When enabled, this Hotspot option outputs the values of an essential subset of configuration parameters.
```
mstoodle, line 27: "essential" doesn't carry the right meaning to me....I would expect a much larger list of options to be included in an "essential" list, but I know what you mean. I can't come up with a better word or phrase at the moment :( .

I'm also not sure we should call it a "Hotspot" option here, unless that's consistent with any other option we implement for Hotspot compatibility....Maybe instead the sentence could end with "in a format compatible with that produced by Hotspot" ?

mstoodle:
I wouldn't bother mentioning the Elasticsearch motivation; we have the issue for that, and the output for the option is not particular to Elasticsearch at all except in the choice of the two options currently included.

mstoodle:
> Can you confirm what the correct default value is. (50:50 guess without rummaging in the code!)

The option is disabled by default.

mstoodle:
I suppose we could give an example of what the output looks like...
```
[Global flags]
   size_t MaxHeapSize                              = 4294967296                                {product} {ergonomic}
 uint64_t MaxDirectMemorySize                      = 3758096384                                {product} {ergonomic}
```

docs/version0.21.md
mstoodle: ..."options we have been told or found to be expected by various software projects and packages?"

sounds awkward, but maybe you can find a cleaner way to say that?

docs/xxprintflagsfinal.md
mstoodle: ..."options we have been told or found to be expected by various software projects and packages?"

sounds awkward, but maybe you can find a cleaner way to say that?

pmhayward:
Jenkins doc stage

pmhayward:
@mstoodle ... staging build respun with latest updates. I didn't like "essential" either, but "couldn't come up with a better word or phrase" :) Have tried again - and other changes made as you suggest.

PR #579, opened 2020-06-18
Remove .DS_Store files

keithc-ca:
I expect `.DS_Store` files were added accidentally.


doveye:
Yes, those shouldn't have gone in. Thanks :)

PR #586, opened 2020-06-24
Better advertise jdk14+ requirement for the AIX XL C++ Runtime

### LINKED ISSUE #585: Better advertise jdk14+ requirement for the AIX XL C++ Runtime
**Subject:**
https://www.eclipse.org/openj9/docs/introduction/

**Suggestion:**

Re https://github.com/AdoptOpenJDK/openjdk-website/issues/741, add the information about the AIX jdk14+ requirement for the XL C++ Runtime to https://www.eclipse.org/openj9/docs/introduction/. It's only found in https://www.eclipse.org/openj9/docs/openj9_support/

pshipton:
Fixes #585


pmhayward:
Jenkins doc stage

pmhayward:
Not sure that an intro to OpenJ9 topic is necessarily the right place for this comment about AIX build requirements - we think it's a bit too much detail at that point! 
Also, you've put it in the "Configuring your system" section, which I thought is about *running* an OpenJDK with OpenJ9 - whereas that comment is from a topic that's to with *building* it? (Or am I confused?) 
But it would certainly be worth adding the comment to the 0.19.0 "What's New" topic, where JDK14 is introduced.

pshipton:
@pmhayward the XL C++ Runtime is required to run the JVM, which is why I think it's appropriate for the "Configuring your system" topic.

PR #599, opened 2020-07-12
FIX Typo: disableImlining => disableInlining

xiacijie:
Signed-off-by: Cijie Xia <cijie@ualberta.ca>


keithc-ca:
Please see https://github.com/eclipse/openj9/blob/master/CONTRIBUTING.md, in particular the part about signing the Eclipse Contributor Agreement (ECA).

xiacijie:
@keithc-ca ECA signed

SueChaplain:
@keithc-ca - the changes to the /site dir were not required. In fact, the site dir should not be in there at all - looks like it was accidentally added as part of the API doc update and is being removed.

(The site dir is created by the MkDocs static site build process and is not required in the source)

keithc-ca:
@SueChaplain Was that content derived from something else (that needs to be corrected)?

SueChaplain:
@keithc-ca - the doc source is in the `docs/` directory in markdown format. At build time, MkDocs builds the static HTML and creates the `site/` directory, which then gets published. The GitHub repo should contain only the source, but someone did a local build and accidentally included the `site/` dir in their commit.

keithc-ca:
I see #600 appears to remove the unwanted content on the `site` directory. Thanks for the explanation.

PR #602, opened 2020-07-16
Remove sentence about default behaviour of shared classes

doveye:
https://github.com/eclipse/openj9-docs/issues/569

Removed sentence about default behaviour of shared classes as there's already a link to the "Class data sharing" topic, which describes that. Instead, expanded on the info the user can get from that topic. This way, the user can still easily get the shared classes info but the -Xaot topic won't need to be updated (possibly getting missed again and providing incorrect info) when/if the default shared classes behaviour changes in the future.

### DIFF
--- a/docs/xaot.md
+++ b/docs/xaot.md
@@ -42,7 +42,7 @@ Startup performance can be improved by using the shared AOT code to provide nati
 
 ## Default behavior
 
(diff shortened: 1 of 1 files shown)


doveye:
jenkins doc stage

PR #618, opened 2020-08-07
MemoryMXBean method rename

theresa-m:
Fixes: https://github.com/eclipse/openj9-docs/issues/607

- Java 8 API updates
- Java 11 API updates
- 0.22 release notes

Signed-off-by: Theresa Mammarella <Theresa.T.Mammarella@ibm.com>


pmhayward:
Jenkins doc stage

pmhayward:
@theresa-m ... just a quick query, but I assume these same changes have been made in the Javadoc in the code; the APIs here are regenerated from time to time from that source, so will over-write any manual changes made here in between. Sorry if stating the obvious :)

theresa-m:
Yes I have update the openj9 javadocs with the same information.

doveye:
The commits appear to all be html files, not code. The API doc is generated from the code - there's only been a few manual attempts so far, but hopefully there will be a more automated process before too long. The manual attempts are how the [current API doc](https://www.eclipse.org/openj9/docs/api-overview/) was created. So, if the code has been updated, the changes will be picked up in the next regeneration, ie changes to individual html files are not required.

doveye:
It looks to me like the code is updated, so I *think* the version0.22.md change is the only one that's needed in this PR, though we probably need to keep the PR as-is at the moment in case we don't get the API process up and running in time for 0.22.

theresa-m:
Is this ready to be merged?

doveye:
We will hopefully have the new API doc update process partially underway in time for 0.22, so the changes in this PR might be overwritten, though that should be ok because the result should be the same. And if the process isn't ready, then this PR will have updated the text anyway, so I'm happy to merge.

PR #627, opened 2020-08-27
The Default max heap size limit

### LINKED ISSUE #626: Expain default maximum heap size limit
Re https://github.com/eclipse/openj9/issues/10200

**Subject:**
Default Maximum heap size

**Suggestion:**
Default Maximum heap size is limited to 25GB, explain the reason for the limit and how to overwrite the limit.

LinHu2016:
- explain the default max heap size limit to 25GB.
        - options to overwrite the limit.

Closes #626
Closes https://github.com/eclipse/openj9/issues/10200

### DIFF
--- a/docs/openj9_defaults.md
+++ b/docs/openj9_defaults.md
@@ -93,6 +93,10 @@ The default value of [`-Xmx`](xms.md) :
 
 - The value is 25% of the available memory with a maximum of 25 GB. However, where there is 2 GB or less of physical memory, the value set is 50% of available memory with a minimum value of 16 MB and a maximum value of 512 MB.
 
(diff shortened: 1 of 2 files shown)


pshipton:
Does this close one of the OpenJ9 issues, such as https://github.com/eclipse/openj9/issues/10200? If so we should use an auto-close keyword in the description. https://docs.github.com/en/enterprise/2.13/user/articles/closing-issues-using-keywords#:~:text=You%20can%20include%20keywords%20in,pull%20request%20and%20the%20issue.
i.e. `closes https://github.com/eclipse/openj9/issues/10200`

pshipton:
Note the docs team requests that every doc PR have an associated issue.

pshipton:
Ah, I found #626 but it isn't referenced here. Updated the description.

pmhayward:
Jenkins doc stage

PR #629, opened 2020-09-03
Update the jdk 11 iframe to use a name which search.js expects

davehobbs:
The search.js script expects an iframe with a name of "classFrame" otherwise the search silently (to the user) fails.


davehobbs:
Jenkins doc stage

PR #631, opened 2020-09-09
Update documentation for PortableSharedCache

harryyu1994:
Introduced additional changes to the Portable Shared Cache feature (-XX:[+|-]PortableSharedCache) which eliminates the AOT portability issues of Compressed Reference OpenJ9 builds with heap sizes from 0GB to 28GB. Previously compressed reference JVMs with different heap sizes cannot share AOT compiled code with each other. When Portable Shared Cache is enabled (either via -XX:[+|-]PortableSharedCache or by default in containers), Compressed Reference OpenJ9 builds with heap size in the range of 0GB to 28GB will be able to share AOT complied code with each other. However there will be a slight steady-state throughput penalty of 1-2% when using this option now compared to before.

Signed-off-by: Harry Yu <harryyu1994@gmail.com>

### DIFF
--- a/docs/version0.23.md
+++ b/docs/version0.23.md
@@ -1,0 +1,48 @@
+<!--
+* Copyright (c) 2017, 2020 IBM Corp. and others
+*
+* This program and the accompanying materials are made
+* available under the terms of the Eclipse Public License 2.0
+* which accompanies this distribution and is available at
+* https://www.eclipse.org/legal/epl-2.0/ or the Apache
+* License, Version 2.0 which accompanies this distribution and
+* is available at https://www.apache.org/licenses/LICENSE-2.0.
+*
+* This Source Code may also be made available under the
+* following Secondary Licenses when the conditions for such
+* availability set forth in the Eclipse Public License, v. 2.0
+* are satisfied: GNU General Public License, version 2 with
+* the GNU Classpath Exception [1] and GNU General Public
+* License, version 2 with the OpenJDK Assembly Exception [2].
+*
+* [1] https://www.gnu.org/software/classpath/license.html
+* [2] http://openjdk.java.net/legal/assembly-exception.html
+*
+* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH
+* Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+-->
+
+
+# What's new in version 0.23.0
+
+The following new features and notable changes since v 0.22.0 are included in this release:
+
+
+## Features and changes
+
+### Binaries and supported environments
+
+OpenJ9 release 0.20.0 supports OpenJDK 8, 11, and 14. Binaries are available from the AdoptOpenJDK community at the following links:
+
+- [OpenJDK version 8](https://adoptopenjdk.net/archive.html?variant=openjdk8&jvmVariant=openj9)
+- [OpenJDK version 11](https://adoptopenjdk.net/archive.html?variant=openjdk11&jvmVariant=openj9)
+- [OpenJDK version 14](https://adoptopenjdk.net/archive.html?variant=openjdk14&jvmVariant=openj9)
+
+To learn more about support for OpenJ9 releases, including OpenJDK levels and platform support, see [Supported environments](openj9_support.md).
+
+
+### `-XX:[+|-]PortableSharedCache` option behavior update
+The `-XX:[+|-]PortableSharedCache` option is updated to improve the portability of AOT-compiled code further. This update allows AOT-compiled code to be portable across OpenJ9 VMs that use compressed references and have a heap size of 1 MB to 28 GB when this option is enabled. This option might introduce a small (1-2%) steady-state throughput penalty when compressed references are used and the heap size is between 1 MB and 3 GB. See [`-XX:[+|-]PortableSharedCache`](xxportablesharedcache.md) for more details about this option.
+
+
+<!-- ==== END OF TOPIC ==== version0.23.md ==== -->
--- a/docs/xxportablesharedcache.md
+++ b/docs/xxportablesharedcache.md
@@ -24,11 +24,17 @@
 
 # -XX:\[+|-\]PortableSharedCache
 
-**(x86 only)**
+**(x86 platforms only)**
 
-Use this command line option to choose whether AOT code should be generated using an older processor (Intel&reg; Sandybridge) feature set on the x86 platform, which therefore allows the AOT code to be portable; it can be run on any Intel Sandybridge or newer processor without you having to choose a machine with a suitably old processor for generating the AOT code.
+Use this command line option to choose whether AOT-compiled code should be portable.
 
-This feature is particularly relevant for packaging a shared classes cache into a container image (for example, applications deployed on the cloud in the form of Docker containers) because the processor on which the container image is built is likely to be different from the processor on which the container is deployed. 
+This option, when enabled, increases the portability of AOT-compiled code, in the following ways:
+- The code is generated based on a particular set of processor features (present in the Intel® Sandy Bridge microarchitecture) that ensures the AOT-compiled code to be portable across any Intel Sandybridge or newer processor.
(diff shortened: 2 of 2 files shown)


harryyu1994:
@mpirvu Hi Marius, could I get a review from you?

docs/version0.22.md
```diff
@@ -49,7 +49,7 @@ To learn more about support for OpenJ9 releases, including OpenJDK levels and pl
 
 ### New `-XX:[+|-]PortableSharedCache` option added
 
-This option enables AOT compiled code to be generated based on a chosen set of processor features that maximizes its portability across machines. It is currently supported only on x86. The feature is turned on by default in Docker containers and can be disabled with `-XX:-PortableSharedCache`. For more information, see [`-XX:[+|-]PortableSharedCache`](xxportablesharedcache.md).
+This option enables AOT compiled code to be portable across CPUs of different microarchitectures and portable across OpenJ9 Compressed Reference builds with a heap size in the range of 0 GB to 28 GB. It is currently supported only on x86. This portable AOT feature has a steady-state throughput penalty of 1% to 2% for OpenJ9 Compressed Reference builds with a heap size in the range of 0 GB to 3GB. It is turned on by default in Docker containers and can be disabled with `-XX:-PortableSharedCache`. For more information, see [`-XX:[+|-]PortableSharedCache`](xxportablesharedcache.md).
```
mpirvu, line 52: This comment pertains to the 0.22 release (`docs/version0.22.md`) and it should not change since the 0.22 release was not changed.
We may want to add some words in a `docs/version023.md` file

docs/xxportablesharedcache.md
```diff
@@ -24,9 +24,9 @@
...
+This option enables AOT compiled code to be generated based on a chosen set of processor features (features present in the Intel&reg; Sandy Bridge microarchitecture) that maximizes its portability across processors of different microarchitectures (Intel&reg; Sandy Bridge or newer). This option also enables AOT compiled code to be generated in a way that is portable across OpenJ9 Compressed Reference builds with a heap size in the range of 0 GB to 28 GB. This portable AOT feature has a steady-state throughput penalty of 1% to 2% for OpenJ9 Compressed Reference builds with a heap size in the range of 0 GB to 3GB. It is turned on by default in Docker containers and can be disabled with `-XX:-PortableSharedCache`. It is currently supported only on x86. The feature is turned on by default in Docker containers but can be disabled with `-XX:-PortableSharedCache`. To enable the portable shared cache feature outside Docker containers, `-XX:+PortableSharedCache` needs to be specified for the initial Java Virtual Machine instance (when the creation of the Shared Class Cache happens) as well as for every subsequent instance that makes use of the same Shared Class Cache.
```
mpirvu, line 27: "heap size in the range of 0 GB to 28 GB", technically speaking the smallest heap one can have is 4MB.

mpirvu: "and can be disabled" --> "but can be disabled"

mpirvu: Actually I see that that line is duplicated

mpirvu: This option enables AOT compiled code to be generated based on a chosen set of processor features (features present in the Intel&reg; Sandy Bridge microarchitecture) that maximizes its portability across processors of different microarchitectures (Intel&reg; Sandy Bridge or newer). This option also enables AOT compiled code to be generated in a way that is portable across OpenJ9 Compressed Reference builds with a heap size in the range of 4 MB to 28 GB. This portable AOT feature may introduce a small (1-2%) steady-state throughput penalty when OpenJ9 Compressed Reference builds are used with a heap size in the range of 4 MB to 3GB. It is currently supported only on x86. The feature is turned on by default in Docker containers but can be disabled with `-XX:-PortableSharedCache`. To enable the portable shared cache feature outside Docker containers, `-XX:+PortableSharedCache` needs to be specified for the initial Java Virtual Machine instance (when the creation of the Shared Class Cache happens) as well as for every subsequent instance that makes use of the same Shared Class Cache.
This option is especially useful for applications deployed on the cloud in the form of Docker containers, as it allows the Shared Class Cache to be portable across processors with different microarchitectures and across Compressed Reference OpenJ9 builds with different heap sizes.

doveye: I suggest the following rewrite of Marius's update with tweaks for consistency and clarity, including moving the platform restriction up front and making it similar to other topics such as https://www.eclipse.org/openj9/docs/xxhandlesigxfsz/, de-capitalising various terms, and using a more active voice. Copyright symbols are required only on the first instance. Also...  
- I changed "chosen set" to "particular" set otherwise you wonder who's doing the choosing :) 
- I assume we don't really mean builds, ie a build of OpenJDK that contains OpenJ9 , but rather instances of the VM that are using compressed refs 

-----

**(x86 platforms only)**

This option, when enabled, increases the portability of AOT-compiled code, in the following ways:
- The code is generated based on a particular set of processor features (present in the Intel® Sandy Bridge microarchitecture) that maximize its portability across processors of different microarchitectures (Intel Sandy Bridge or newer). 
- The code is generated so that it is portable across OpenJ9 VMs that use compressed references and have a heap size of 4 MB to 28 GB (previously, AOT-compiled code could not be shared between VMs that use compressed references and that have different heap sizes). This feature might introduce a small (1-2%) steady-state throughput penalty when compressed references are used and the heap size is between 4 MB and 3GB. 

This option is especially useful for applications that are deployed on the cloud in the form of Docker containers because it allows the shared classes cache to be portable across processors with different microarchitectures and across compressed reference VMs with different heap sizes. 

This option is therefore enabled by default in Docker containers but you can disable it by specifying `-XX:-PortableSharedCache` on the command line. To enable the portable shared cache feature outside Docker containers, specify `-XX:+PortableSharedCache` for the initial virtual machine instance (which is when the shared classes cache is created) and for every subsequent instance that uses the same shared classes cache.

mpirvu: Sounds much better. Thanks @doveye !

harryyu1994:
@mpirvu @doveye ready for review. I'm not sure if I should be the one adding `docs/version0.23.md` though.

docs/version0.23.md
```diff
@@ -0,0 +1,48 @@
...
+
+- [OpenJDK version 8](https://adoptopenjdk.net/archive.html?variant=openjdk8&jvmVariant=openj9)
+- [OpenJDK version 11](https://adoptopenjdk.net/archive.html?variant=openjdk11&jvmVariant=openj9)
+- [OpenJDK version 14](https://adoptopenjdk.net/archive.html?variant=openjdk14&jvmVariant=openj9)
+
+To learn more about support for OpenJ9 releases, including OpenJDK levels and platform support, see [Supported environments](openj9_support.md).
+
+
+### `-XX:[+|-]PortableSharedCache` option behavior update
+The `-XX:[+|-]PortableSharedCache` option is updated to improve the portability of AOT-compiled code further. The new update allows AOT-compiled code to be portable across OpenJ9 VMs that use compressed references and have a heap size of 4 MB to 28 GB when this option is enabled. This option might introduce a small (1-2%) steady-state throughput penalty when compressed references are used and the heap size is between 4 MB and 3 GB. See [`-XX:[+|-]PortableSharedCache`](xxportablesharedcache.md) for more details about this option.
```
mpirvu, line 45: Maybe replace "The new update" with "This update"

mpirvu (commented):
I have one small tweak and one correction: replace 4MB with 1MB throughout this document. I just checked and the minimum heap size is 1MB (at least for Java 8)

mpirvu (approved):
LGTM

doveye:
Jenkins doc stage

harryyu1994:
Going to need another round of review after resolving conflicts. @mpirvu @doveye

mpirvu (approved):
LGTM

pmhayward:
@harryyu1994 I see that #637 is no longer for 0.23. Does that mean this PR is not for 0.23 either? Should it be therefore be backed out for the time being?

mpirvu:
This PR is still good because the features that it describes have been included in the 0.23 branch.
#637 was describing new features that missed the 0.23 code split by a few days.
@harryyu1994 correct me if I am wrong

PR #634, opened 2020-09-14
IdleTuningCompactOnIdle disabled

amicic:
Update documentation for IdleTuningCompactOnIdle, stating new behavior
since starting with 0.23

Signed-off-by: Aleksandar Micic <amicic@ca.ibm.com>

### DIFF
--- a/docs/xxidletuningcompactonidle.md
+++ b/docs/xxidletuningcompactonidle.md
@@ -39,14 +39,7 @@ This option controls garbage collection processing with compaction when the stat
 
         -XX:[+|-]IdleTuningCompactOnIdle
 
-| Setting                        | Effect  | Default  | Default when running in a docker container                                 |
-|--------------------------------|---------|:--------:|:--------------------------------------------------------------------------:|
-| `-XX:+IdleTuningCompactOnIdle` | Enable  |  | <i class="fa fa-check" aria-hidden="true"></i><span class="sr-only">yes</span>     |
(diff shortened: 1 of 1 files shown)


SueChaplain:
@amicic 
In the file you've changed, I think I would preserve more of the older information for now as some users will be running on older versions of OpenJ9. We also have blogs and PPTs that cover the idle behavior, so removing what it did might cause some confusion. 

Maybe you could add a line under **Linux only**:

**Warning:** From OpenJ9 version 0.23.0 this option has no effect.

Add a few words before the opening sentence "In versions of OpenJ9 before 0.23.0, this option controls garbage collection processing with compaction when the state of the OpenJ9 VM is set to idle."

Then extend the deprecation announcement,  so it is clear why there is a change in direction. Something like ... 

**Restrictions:**
1. This option was deprecated in release 0.15.0 due to operational changes. A compaction is now triggered by internal heuristics that look into the amount of fragmented pages. Typically there is no need to force a compaction. This option will be removed in the future.

Then leave the rest of this topic as-is.

Notes:  

- In addition to changing `xxidletuningcompactonidle.md`, I think a change is also required here: https://eclipse.github.io/openj9-docs/introduction/#cloud-optimizations, where we say "The object heap is also compacted to make best use of the available memory for further application processing."??
- We might also change content in the -xx related topics to indicate that this -xx option is a no-op.
- We also need a whats new entry in the release topic.

Someone from the team here can fix up the other topics if you can confirm the changes required. Sometimes a small change has further reaching consequences ;-)

pmhayward:
These changes and extra updates suggested by @SueChaplain  incorporated into #647

PR #642, opened 2020-09-24
Further updates for balanced gc

SueChaplain:
- add more info about heap layout, incl. diagram

Signed-off-by: Sue Chaplain <sue_chaplain@uk.ibm.com>

PR #646, opened 2020-09-29
Further updates for metronome/optavgpause/optthruput GC

pmhayward:

### DIFF
--- a/docs/gc.md
+++ b/docs/gc.md
@@ -130,17 +130,50 @@ There are a number of advantages to using arraylets.
 <i class="fa fa-pencil-square-o" aria-hidden="true"></i> **Note:** Despite the general advantage of using arraylets, they can slow down processing when the Java Native Interface (JNI) is being used. The JNI provides flexibility by enabling Java programs to call native code; for example C or C++, and if direct addressability to the inside of an object is needed, a JNI critical section can be used. However, that requires the object to be in a contiguous region of memory, or at least _appear_ to be so. The JNI therefore creates a temporary contiguous array that is the same size as the original array and copies everything, element by element, to the temporary array. After the JNI critical section is finished, everything is copied from the temporary array back to the arraylet, element by element.
 
 
-### `metronome` policy
-
-[`-Xgcpolicy:metronome`](xgcpolicy.md#metronome-aix-linux-x86-only) is designed for applications that require precise response times. Garbage collection occurs in small interruptible steps to avoid stop-the-world pauses. This policy is available only on x86 Linux and AIX platforms.
-
 ### `optavgpause` policy
 
-[`-Xgcpolicy:optavgpause`](xgcpolicy.md#optavgpause) uses concurrent mark and sweep phases, which means that pause times are reduced when compared to optthruput, but at the expense of some performance throughput.
+[`-Xgcpolicy:optavgpause`](xgcpolicy.md#optavgpause) uses the global collector to manage a flat heap comprised of a single region and to compact the heap if the heap becomes fragmented. The global collector is started preemptively so that the cycle finishes before the heap is completely exhausted. By anticipating global collections and initiating some marking before the stop-the-world global collection phase, the `optavgpause` policy reduces the likelihood of long interruptions of service due to GC activity and pause times are reduced when compared to `optthruput` though at the expense of some performance throughput.
+
+#### When to use
+
+You are likely to find that in many situations, the default [`gencon` policy](#gencon-policy-default) offers better performance than `optavgpause`. However, the `optavgpause` policy lacks a generational write barrier (though it does have a write barrier to support concurrent mark, and might therefore be of benefit with benefit workloads that have a costly generational barrier; for example, in situations that frequently change large and old reference arrays.
+
+An additional potential problem with generation GCs is that they split the heap so that a very large object in a worst case might not fit in either of the 2 areas, while there is enough total free memory, or in better case, allocation may succeed but require additional GC cycles and copying of objects (relative to non-generational GC) to accommodate large enough contiguous free space.
+
+Overall, `optavgpause`, along with `optthruput`, is best suited to short-lived applications and to long-running services involving concurrent sessions that have short lifespans. Short-lived applications with adequate heap sizes usually complete without compaction, and the flat heap fragments more slowly when session-bound objects are allocated and drop out of the live set in short overlapping clusters.
+
 
 ### `optthruput` policy
 
-[`-Xgcpolicy:optthruput`](xgcpolicy.md#optthruput) is optimized for throughput by disabling the concurrent mark phase, which means that applications will stop for long pauses while garbage collection takes place. You might consider using this policy when high application throughput, rather than short garbage collection pauses, is the main performance goal.
+[`-Xgcpolicy:optthruput`](xgcpolicy.md#optthruput) uses the global collector to manage a flat heap comprised of a single region and to compact the heap if the heap becomes fragmented. The global collector runs to mark and sweep the heap when the heap is exhausted, which means that applications stop for long pauses while garbage collection takes place.
+
+#### When to use
+
+You might consider using this policy when your application can tolerate longer GC pauses to obtain better overall throughput. However, you are likely to find that in many situations, the default [`gencon` policy](#gencon-policy-default) offers better performance than `optthruput`. Nevertheless, the `optthruput` policy is free of any object access barriers (unlike `gencon` that has a write barrier to support concurrent mark, and a generational write barrier) and in some workloads, the cost of those barriers might be high enough to make this policy preferable.
+
+An additional potential problem with generation GCs is that they split the heap so that a very large object in a worst case might not fit in either of the 2 areas, while there is enough total free memory, or in better case, allocation may succeed but require additional GC cycles and copying of objects (relative to non-generational GC) to accommodate large enough contiguous free space.
+
+Overall, `optthruput`, along with `optavgpause`, is best suited to short-lived applications and to long-running services involving concurrent sessions that have short lifespans. Short-lived applications with adequate heap sizes usually complete without compaction, and the flat heap fragments more slowly when session-bound objects are allocated and drop out of the live set in short overlapping clusters.
+
+
+### `metronome` policy
+
+**(Linux on x86-64 and AIX platforms only)**
+
+[`-Xgcpolicy:metronome`](xgcpolicy.md#metronome-aix-linux-x86-only) is designed for applications that require a precise upper bound on collection pause times as well as specified application utilization: the proportion of time that the application is permitted to use, with the remainder being devoted to GC. The realtime GC runs in short interruptible bursts to avoid long stop-the-world pauses.
+
+The realtime heap is allocated as a contiguous range of RAM partitioned into small regions of equal size, usually about 64 KB. There is no dynamic heap resizing when you use the `metronome` policy. The heap is always fully expanded even if [`-Xms`](xms.md) is not set to [`-Xmx`](xms.md).
+
+As with the `balanced` GC policy, arrays are represented as arraylets with a spine pointing to a series of regions containing the array elements (see [Arraylets: dealing with large arrays](#arraylets-dealing-with-large-arrays) for more information). Note that JNI access to array data might involve reconstituting arraylets as contiguous arrays, which can significantly slow down processing.
+
+Each region of the heap is either empty, or contains only objects in one of 16 size classes, or an arraylet. This organization improves the use of available heap space, reducing the need for heap compaction and defragmentation, and providing more precise control over the *incremental sweep* operation.
+
+Although high application utilization is desirable for optimal throughput, the GC must be able to keep up with the application's memory  allocation rate. 
+
+A higher utilization typically requires a larger heap because the GC isn't allowed to run as much as a lower utilization would permit. The relationship between utilization and heap size is highly application dependent, and striking an appropriate balance requires iterative experimentation with the application and VM parameters. You might need to adjust heap size or pause time or target utilization to achieve an acceptable runtime configuration.
+
+See [`-Xgcpolicy:metronome`](xgcpolicy.md#metronome-aix-linux-x86-only) for details of the configuration parameters you can set.
+
 
 ### `nogc` policy
 
--- a/docs/xgcpolicy.md
+++ b/docs/xgcpolicy.md
@@ -73,9 +73,9 @@ The initial heap size is *Xmx/1024*, rounded down to the nearest power of 2, whe
 
 The following options can also be specified on the command line with `-Xgcpolicy:balanced`:
 
-- `-Xalwaysclassgc`
-- `-Xclassgc`
-- `-Xcompactexplicitgc`
+<!-- - `-Xalwaysclassgc` -->
+<!-- - `-Xclassgc`
+<!-- - `-Xcompactexplicitgc` -->
 - `-Xdisableexcessivegc`
 - `-Xdisableexplicitgc`
 - `-Xenableexcessivegc`
@@ -94,8 +94,8 @@ The following options can also be specified on the command line with `-Xgcpolicy
 - `-Xmnx<size>`
 - `-Xms<size>`
 - `-Xmx<size>`
-- `-Xnoclassgc`
-- `-Xnocompactexplicitgc`
+<!-- - `-Xnoclassgc` -->
+<!-- - `-Xnocompactexplicitgc` -->
 - `-Xnuma:none`
 - `-Xsoftmx<size>`
 - `-Xsoftrefthreshold<number>`
@@ -103,11 +103,17 @@ The following options can also be specified on the command line with `-Xgcpolicy
 
 The behavior of the following options is different when specified with `-Xgcpolicy:balanced`:
 
-`-Xcompactgc`
-: Compaction occurs when a System.gc() call is received (default). Compaction always occurs on all other collection types.
+[`-Xcompactgc`](xcompactgc.md) (default)
+: Forces compaction in each Global GC cycle.
 
-`-Xnocompactgc`
-: Compaction does not occur when a System.gc() call is received. Compaction always occurs on all other collection types.
+[`-Xnocompactgc`](xcompactgc.md)
+: Disables internal compaction heuristics in Global GC cycles.
+
+[`-Xcompactexplicitgc`](xcompactexplicitgc.md) (default)
+: Forces compaction in Explicit Global GC cycles, such as those invoked by `System.gc()`. Compaction in implicit Global GC remains optional, triggered by internal heuristics.
+
+[`-Xnocompactexplicitgc`](xcompactexplicitgc.md)
+: Disables compaction in Explicit Global GC cycles. Compaction in implicit Global GC remains optional, triggered by internal heuristics.
 
 The following options are ignored when specified with `-Xgcpolicy:balanced`:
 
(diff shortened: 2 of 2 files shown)


pmhayward:
Jenkins doc stage

docs/gc.md
```diff
@@ -132,15 +132,35 @@ There are a number of advantages to using arraylets.
...
 
-### `optthruput` policy
+As with the `balanced` GC policy, arrays are represented as arraylets with a spine pointing to a series of regions containing the array elements (see [Arraylets: dealing with large arrays](#arraylets-dealing-with-large-arrays) for more information). Note that JNI access to array data might involve reconstituting arraylets as contiguous arrays, which can significantly slow down processing.
 
-[`-Xgcpolicy:optthruput`](xgcpolicy.md#optthruput) is optimized for throughput by disabling the concurrent mark phase, which means that applications will stop for long pauses while garbage collection takes place. You might consider using this policy when high application throughput, rather than short garbage collection pauses, is the main performance goal.
+Each region of the heap is either empty or contains only objects of the same size or an arraylet. This organization simplifies new object allocation and consolidation of free heap space, allowing the realtime collector to maintain GC throughput while supporting a consistent service level.
```
amicic, line 143: more precisely, it's same _size class_

There is a lower and upper bound for the size (upper bound of one size class is same as the lower bound of the next large size class. Next size class (their bounds) is approx 20% larger than previous.

The point is not in allocation simplification nor consolidation of free space, but in
1) reducing heap fragmentation, so that need for heap compaction/de-fragmentation is minimized
2) more precise control of incremental sweep phase

docs/gc.md
```diff
@@ -132,15 +132,35 @@ There are a number of advantages to using arraylets.
...
+
+Although high application utilization is desirable for optimal throughput, the GC must be able to keep up with the application's memory  allocation rate. 
+
+A higher utilization typically requires a larger heap because the GC isn't allowed to run as much as a lower utilization would permit. The relationship between utilization and heap size is highly application dependent, and striking an appropriate balance requires iterative experimentation with the application and VM parameters. You might need to adjust heap size or pause time or target utilization to achieve an acceptable runtime configuration.
+
+See [`-Xgcpolicy:metronome`](xgcpolicy.md#metronome-aix-linux-x86-only) for details of the configuration parameters you can set.
+
+### `optavgpause` and `optthruput` policies
+
+Both of these policies use the global collector to manage a flat heap comprised of a single region. The application runs until the heap is exhausted and the global collector runs to mark and sweep the heap. Additionally, if the heap becomes fragmented, the global collector compacts the heap.
```
amicic, line 153: global Mark-and-Sweep collector 

'runs until the heap is exhausted' is only true for throughput. For optavgpause, a cycle is preemptively started (kicked off) so that the cycle finishes before heap is completely exhausted.

docs/gc.md
```diff
@@ -132,15 +132,35 @@ There are a number of advantages to using arraylets.
...
+
+These policies are best suited to short-lived applications and to long-running services involving concurrent sessions that have short lifespans. Short-lived applications with adequate heap sizes usually complete without compaction, and the flat heap fragments more slowly when session-bound objects are allocated and drop out of the live set in short overlapping clusters.
```
amicic, line 155: In reality these two policies are rather inferior to gencon and it's hard to find a scenario where they would really applicable. I'll elaborate a bit for hopefully better general understanding, although I'm not sure how much of it we really want to document.

The best advantage of optthruput is that it's free of any object access barriers (unlike optavgpause tjat has write barrier to support concurrent mark, and gencon which in addition to that one also has a generational write barrier). In some workloads, cost of those barriers may be high enough to outweigh the other benefits of concurrent and generational aspect (like low pauses and high yield of free memory).

To some extent we could say that for optavgpause, but applying only to the lack of generational barrier. So, for example worklaods that frequently change large and old reference arrays may have costly generational barrier, since those array have to be scanned as part of root set for Scavenge.

An additional potential problem with generation GCs is that it splits heap so that a very large object in worst case might not fit in either of the 2 areas, while there is enough total free memory, or in better case, allocation may succeed but require additional GC cycles and copying of objects (relative to non-generational GC) to accommodate large enough contiguous free space.

docs/xgcpolicy.md
```diff
@@ -104,10 +104,10 @@ The following options can also be specified on the command line with `-Xgcpolicy
 The behavior of the following options is different when specified with `-Xgcpolicy:balanced`:
 
 `-Xcompactgc`
-: Compaction occurs when a System.gc() call is received (default). Compaction always occurs on all other collection types.
+: Compaction occurs when a `System.gc()` call is received (default). Compaction always occurs on all other collection types.
```
amicic, line 107: I don't think this has anything to do with System GC.

It's as simple as that global (mark-sweep) GC will have mandatory compact operation (which otherwise is just optional triggered by various internal heuristics.

It's irrelevant global GC is triggered explictly (System GC) or implicitely, if it's concurrnet or STW global etc.

docs/xgcpolicy.md
```diff
@@ -104,10 +104,10 @@ The following options can also be specified on the command line with `-Xgcpolicy
...
 
 `-Xnocompactgc`
-: Compaction does not occur when a System.gc() call is received. Compaction always occurs on all other collection types.
+: Compaction does not occur when a `System.gc()` call is received. Compaction always occurs on all other collection types.
```
amicic, line 110: Similarly as before, this has nothing to do System GC. It's just disable internal heuristics that trigger Global GC.

We do however two special options that control that for just System GC:

-Xcompactexplicitgc		compact on all system GC's
-Xnocompactexplicitgc	do not compact on system GC's

amicic: A comment for a section down below that is not edited: I would not bother mentioning -Xpartialcompactgc/  -Xnopartialcompactgc there and actually anywhere in documentation. They are not doing anything even in standard 3 GCs policies

docs/xgcpolicy.md
```diff
@@ -136,15 +136,16 @@ The following options are ignored when specified with `-Xgcpolicy:balanced`:
...
 
 #### Defaults and options
 
 `-Xgc:synchronousGCOnOOM | -Xgc:nosynchronousGCOnOOM`
 : One occasion when garbage collection occurs is when the heap runs out of memory. If there is no more free space in the heap, using `-Xgc:synchronousGCOnOOM` stops your application while garbage collection removes unused objects. If free space runs out again, consider decreasing the target utilization to allow garbage collection more time to complete. Setting `-Xgc:nosynchronousGCOnOOM` implies that when heap memory is full your application stops and issues an out-of-memory message. The default is `-Xgc:synchronousGCOnOOM`.
 
 `-Xnoclassgc`
-: Disables class garbage collection. This option switches off garbage collection of storage associated with Java classes that are no longer being used by the OpenJ9 VM. The default behavior is -Xnoclassgc.
+: Disables class garbage collection. This option switches off garbage collection of storage associated with Java classes that are no longer being used by the OpenJ9 VM. The default behavior is `-Xnoclassgc`.
```
amicic, line 148: default is -Xclassgc (not -Xnoclassgc), which heuristically decides which GC cycle will attempt to unload classes.

We also have -Xalwaysclassgc that will attempt to unload classes on each applicable GC (what is Global GC).

These 3 options are not applicable to Balanced  GC polices. Class unloading there is done incrementally in Partial GCs. There are a couple of -XXgc: options that can disable/enable it, but I don't know if we want to expose them (we generally don't document -XXgc options)

docs/gc.md
```diff
@@ -132,15 +132,35 @@ There are a number of advantages to using arraylets.
...
+Each region of the heap is either empty or contains only objects of the same size or an arraylet. This organization simplifies new object allocation and consolidation of free heap space, allowing the realtime collector to maintain GC throughput while supporting a consistent service level.
+
+Although high application utilization is desirable for optimal throughput, the GC must be able to keep up with the application's memory  allocation rate. 
+
+A higher utilization typically requires a larger heap because the GC isn't allowed to run as much as a lower utilization would permit. The relationship between utilization and heap size is highly application dependent, and striking an appropriate balance requires iterative experimentation with the application and VM parameters. You might need to adjust heap size or pause time or target utilization to achieve an acceptable runtime configuration.
+
+See [`-Xgcpolicy:metronome`](xgcpolicy.md#metronome-aix-linux-x86-only) for details of the configuration parameters you can set.
```
amicic, line 149: I don't know if it's mentioned anywhere, but there is no dynamic heap resizing in Metronome. Heap is always fully expanded even if -Xms is not set to -Xmx.

docs/xgcpolicy.md
```diff
@@ -173,13 +172,21 @@ The following options are ignored when specified with `-Xgcpolicy:balanced`:
 
         -Xgcpolicy:optavgpause
 
-: The "optimize for pause time" policy uses concurrent mark and concurrent sweep phases. Pause times are shorter than with `optthruput`, but application throughput is reduced because some garbage collection work is taking place while the application is running. Consider using this policy if you have a large heap size (available on 64-bit platforms), because this policy limits the effect of increasing heap size on the length of the garbage collection pause. However, if your application uses many short-lived objects, the `gencon` policy might produce better performance.
+: The "optimize for pause time" policy uses concurrent mark and concurrent sweep phases. Pause times are shorter than with `optthruput`, but application throughput is reduced because some garbage collection work is taking place while the application is running.
```
amicic, line 175: GC must run one way (STW) or another (concurrent) stealing CPU cycles, so that is not really a reason it's reducing throughput.
It's  because 1) mutator threads are penalized with doing GC work (for example execute object assess barriers) and 2) concurrent GC generate floating garbage (effectively dead objects that GC must keep alive), what in turn reduces effective free memory, which increases frequency of GCs.

docs/gc.md
pmhayward: Updates made.

docs/gc.md
pmhayward: Updates made.

docs/gc.md
pmhayward: I've changed the docs and incorporated some of the information above. Does it now make sense? Is it accurate? Have I said enough / too much?!

docs/xgcpolicy.md
pmhayward: Sorry, not sure what this should say instead

docs/xgcpolicy.md
pmhayward: Default changed.
-Xalwaysclassgc added.
3 options removed from Balanced GC section.

docs/gc.md
pmhayward: Have added a sentence to that effect.

docs/xgcpolicy.md
pmhayward: Wonder if those 'real' reasons are a bit too much detail for the user doc? Would it be better to just not go into detailed reasons? (i.e. just delete "...because some garbage collection work is taking place while the application is running.")
If you think those two reasons _should_ be included, is there some simpler wording that might work?

docs/xgcpolicy.md
pmhayward: As above re -Xnocompactgc. Some help please!
..........
Have removed mention of two deprecated options. (Only referenced in this topic.)

pmhayward:
Jenkins doc stage

pmhayward:
@amicic ... thanks for comments above. I've dealt with most of them as commented above, but have left you a few queries. Please review the changes I've made, and help me with other questions.
Thanks (I'm back in (home!) office on 1 Dec.)

docs/gc.md
```diff
@@ -132,15 +132,43 @@ There are a number of advantages to using arraylets.
 
 ### `metronome` policy
 
-[`-Xgcpolicy:metronome`](xgcpolicy.md#metronome-aix-linux-x86-only) is designed for applications that require precise response times. Garbage collection occurs in small interruptible steps to avoid stop-the-world pauses. This policy is available only on x86 Linux and AIX platforms.
+**(Linux on x86-64 and AIX platforms only)**
 
-### `optavgpause` policy
+[`-Xgcpolicy:metronome`](xgcpolicy.md#metronome-aix-linux-x86-only) is designed for applications that require a precise upper bound on collection pause times as well as specified application utilization: the proportion of time that the application is permitted to use, with the remainder being devoted to GC. The realtime GC runs in short interruptible bursts to avoid stop-the-world pauses.
```
amicic, line 137: Those interruptible bursts are also a series of STW pauses, but very short. So either we should say 'to avoid _long_
STW pauses', or we should complete rephrase things like: 'realtime breaks done a single GC STW cycle into a burst/series of very short STW increments'

docs/xgcpolicy.md
amicic: A simple examples:
Compaction is forced in each Global GC cycle 
or in active:
(The option) Forces compaction in each Global GC cycle.

and for nocompact (in passive):
Internal compaction heuristics in Global GC cycle are disabled.

docs/xgcpolicy.md
amicic: Good, but we need to add -Xcompactexplicitgc  and -Xnocompactexplicitgc
It should be the same explanation as for -Xcompactgc and -Xnocompactgc, with emphasis  on _Explicit_ Global GC.

For example, 

for -Xcompactexplicitgc 

Forces compaction in Explicit Global GC cycles (such as those invoked by System.gc(). Compaction in implicit Global GC remains optional, triggered by internal heuristics. 

-Xnocompactexplicitgc 
Disables compaction in Explicit Global GC cycles. Compaction in implicit Global GC remains optional, triggered by internal heuristics.

docs/xgcpolicy.md
amicic: how about: 
because some garbage collection work is taking place in the context of mutator (application) threads, and GC frequency is increased.

docs/xgcpolicy.md
pmhayward: Updates made.

docs/xgcpolicy.md
pmhayward: Updates made, and  -X[no]compactexplicitgc added similarly.

docs/xgcpolicy.md
pmhayward: Updated as suggested.

docs/gc.md
pmhayward: Just added 'long', as suggested.

pmhayward:
Jenkins doc stage

pmhayward:
Jenkins doc stage

pmhayward:
@amicic ... thanks for updated comments. I think I've dealt with them all as you suggest, but could you take a look before we merge these changes. Thanks.
@SueChaplain ... could you have a look too, please?
Staged at... https://eclipse.github.io/openj9-docs-staging/646

docs/gc.md
```diff
@@ -130,17 +130,47 @@ There are a number of advantages to using arraylets.
...
+
+You are likely to find that in many situations, the default [`gencon` policy](#gencon-policy-default) offers better performance than these two policies.
+
+The primary advantage of `optthruput` is that it is free of any object access barriers (unlike `gencon` that has a write barrier to support concurrent mark, and a generational write barrier). In some workloads, the cost of those barriers might be high enough to outweigh the benefits of other policies.
+
+The `optavgpause` policy similarly lacks a generational write barrier, though it does have a write barrier to support concurrent mark. But worklaods that frequently change large and old reference arrays, for example, might have a costly generational barrier and would therefore benefit from this policy.
+
+An additional potential problem with generation GCs is that it splits heap so that a very large object in worst case might not fit in either of the 2 areas, while there is enough total free memory, or in better case, allocation may succeed but require additional GC cycles and copying of objects (relative to non-generational GC) to accommodate large enough contiguous free space.
```
SueChaplain, line 143: ?? Neither of these policies are generational, so this is, I guess a reason for choosing flat heaps over generational. However, I think we need to break this down a bit to make it more consumable.

By using a *flat* heap, these policies avoid potential issues with very large objects. With a generational policy, the heap is divided into smaller spaces in order to manage generations of objects. For `gencon` this is the *nursery* area and for `balanced` this is the *eden* region. Although there might be sufficient free space on the overall Java heap for a very large object, it might not fit into these spaces. If the allocator does succeeds in allocating a very large object, further GC cycles might be required to create enough contiguous free space.

So for me, this becomes a "When to use" (see my TOC suggestions for laying out the policies) .... i.e. If your application requires very large objects that persist on the Java heap ..... 

@amicic - is this what you intended here?

docs/gc.md
```diff
@@ -130,17 +130,47 @@ There are a number of advantages to using arraylets.
 <i class="fa fa-pencil-square-o" aria-hidden="true"></i> **Note:** Despite the general advantage of using arraylets, they can slow down processing when the Java Native Interface (JNI) is being used. The JNI provides flexibility by enabling Java programs to call native code; for example C or C++, and if direct addressability to the inside of an object is needed, a JNI critical section can be used. However, that requires the object to be in a contiguous region of memory, or at least _appear_ to be so. The JNI therefore creates a temporary contiguous array that is the same size as the original array and copies everything, element by element, to the temporary array. After the JNI critical section is finished, everything is copied from the temporary array back to the arraylet, element by element.
 
 
+### `optavgpause` and `optthruput` policies
```
SueChaplain, line 133: I still prefer that we split these out, each with a "When to use" section.

docs/gc.md
```diff
@@ -130,17 +130,47 @@ There are a number of advantages to using arraylets.
...
+
+The realtime heap is allocated as a contiguous range of RAM partitioned into small regions of equal size, usually about 64 KB. There is no dynamic heap resizing when you use the `metronome` policy. The heap is always fully expanded even if [`-Xms`](xms.md) is not set to [`-Xmx`](xms.md).
+
+As with the `balanced` GC policy, arrays are represented as arraylets with a spine pointing to a series of regions containing the array elements (see [Arraylets: dealing with large arrays](#arraylets-dealing-with-large-arrays) for more information). Note that JNI access to array data might involve reconstituting arraylets as contiguous arrays, which can significantly slow down processing.
+
+Each region of the heap is either empty, or contains only objects in one of 16 size classes, or an arraylet. This organization reduces heap fragmentation, so that need for heap compaction and defragmentation is minimized, and provides more precise control of the incremental sweep phase.
```
SueChaplain, line 166: Had trouble reading this .... maybe we need to add a "the" ... " so that the need" ... and of/over to "more precise control over the incremental *sweep* operation". Hmm ... still struggling with ", and provides". 

Maybe:

"This organization improves the use of available heap space, reducing the need for heap compaction and defragmentation, and providing more precise control over the incremental *sweep* operation."

Note: We are calling `<gc-op>`s "operations" with the name in *italics*. Probably easier to sweep the whole topic once we have it merged.

docs/gc.md
pmhayward: Have split `optavgpause` and `optthruput` back into two separate sections, with `metronome` following.

docs/gc.md
pmhayward: Reworded as suggested.

pmhayward:
Jenkins doc stage

pmhayward:
Restaged at... https://eclipse.github.io/openj9-docs-staging/646
(fwiw, would like to get this PR merged (along with other GC-related PRs) and then review the overall position!)

docs/gc.md
```diff
@@ -130,17 +130,50 @@ There are a number of advantages to using arraylets.
...
 
 
-### `metronome` policy
+### `optavgpause` policy
 
-[`-Xgcpolicy:metronome`](xgcpolicy.md#metronome-aix-linux-x86-only) is designed for applications that require precise response times. Garbage collection occurs in small interruptible steps to avoid stop-the-world pauses. This policy is available only on x86 Linux and AIX platforms.
+[`-Xgcpolicy:optavgpause`](xgcpolicy.md#optavgpause) uses the global collector to manage a flat heap comprised of a single region and to compact the heap if the heap becomes fragmented. The global collector is started preemptively so that the cycle finishes before the heap is completely exhausted. By anticipating global collections and initiating some marking before the stop-the-world global collection phase, the `optavgpause` policy reduces the likelihood of long interruptions of service due to GC activity and pause times are reduced when compared to `optthruput` though at the expense of some performance throughput.
```
amicic, line 135: in a recent discussion with @SueChaplain, I think we adopted a term 'area' when talking about total heap, or specific parts of heap (like nursery, tenure, soa, loa). 

Region is closely related to Metronome and Balanced

docs/xgcpolicy.md
```diff
@@ -127,24 +133,53 @@ The following options are ignored when specified with `-Xgcpolicy:balanced`:
...
-- `-Xpartialcompactgc` (deprecated)
+<!-- - `-Xnopartialcompactgc` (deprecated) -->
+<!-- - `-Xpartialcompactgc` (deprecated) -->
+
+
+### `optavgpause`
+
+        -Xgcpolicy:optavgpause
+
+: The "optimize for pause time" policy uses concurrent mark and concurrent sweep phases. Pause times are shorter than with `optthruput`, but application throughput is reduced because some garbage collection work is taking place in the context of mutator (application) threads, and GC frequency is increased.
```
amicic, line 144: just English thing... should we repeat here 'because' (first reason is so long, that a reader might lose context when start reading about the second reason)?

and _because_ GC frequency is increased.

docs/gc.md
amicic: Correct, these are not generational, so the intentions was to enlist potential weakness of generational GCs, more specifically Gencon (object access barrier and heap segregation) as a justification to consider using Flat heap based ones.

I don't know if it's wise to mention here Balanced/Eden etc.  In Balanced, the problem of large object allocation that does not fit into a region is resolved with arraylets. BTW, Eden is actually a set of regions, not a single region.

Overall the text good to me (probably because it's based on my suggestions :) )

minor mistakes:
- problem with generational GCs (not generation)
- worklaods  -> workloads

SueChaplain:
@amicic - I've made a note of the additional comments, which I will handle in a further PR. I'm going to merge this work now so that we can progress overall. When I've got #642 updated and merged I'll do a further sweep on the terminology changes we agreed.

PR #681, opened 2020-11-19
Further updates to support changes to message logging

SueChaplain:
- Further mods to help clarify relationship between -Xlog, -Xsyslog and the new -XX option
- For HotSpot customers, add to table about similar functioning options
- Add notes here and there for existing users about -Xlog superseded
- Some edits in the version topic for consistency etc.

### DIFF
--- a/docs/cmdline_migration.md
+++ b/docs/cmdline_migration.md
@@ -39,7 +39,7 @@ You can use the following command-line options in OpenJ9, just as you did in Hot
 | [`-Xcheck:jni`](xcheck.md)                                       | Runs additional checks for JNI functions during VM startup.                                                                                  |
 | [`-Xfuture`](xfuture.md)                                         | Turns on strict class-file format checks.                                                                                                    |
 | [`-Xint`](xint.md)                                               | Runs an application in interpreted-only mode.                                                                                                |
-| [`-Xlog`](xlog.md)                                                | Some forms of `-Xlog` that enable GC logging are recognized. (Equivalent to [`-Xverbosegclog`](xverbosegclog.md)).
+| [`-Xlog`](xlog.md)                                               | Some forms of `-Xlog` that enable garbage collection logging are recognized. (Equivalent to [`-Xverbosegclog`](xverbosegclog.md)).
 | [`-Xmn`](xmn.md)                                                 | Sets the initial and maximum size of the new area when using -Xgcpolicy:gencon.                                                              |
 | [`-Xms`](xms.md)                                                 | Sets the initial size of the heap. (Equivalent to `-XX:InitialHeapSize`)                                                                     |
 | [`-Xmx`](xms.md)                                                 | Specifies the maximum size of the object memory allocation pool. (Equivalent to `-XX:MaxHeapSize`)                                           |
@@ -73,11 +73,11 @@ You can use the following command-line options in OpenJ9, just as you did in Hot
 
 These HotSpot command-line options have equivalents in OpenJ9 that are not specified in the same way, but perform a related function:
 
-| HotSpot Option          | OpenJ9 Option                                    | Usage                                                            |
-|-------------------------|--------------------------------------------------|------------------------------------------------------------------|                                                                        
-| `-Xcomp`                | [`-Xjit:count=0`](xjit.md#count)**<sup>1</sup>** | `-Xcomp` disables interpreted method invocations.                |
-| `-Xgc`                  | [`-Xgcpolicy`](xgcpolicy.md)**<sup>2</sup>**     | Configuring your garbage collection policy.                      |
-| `-XX:+UseNUMA`          | [`-Xnuma:none`](xnumanone.md)**<sup>3</sup>**    | Controls non-uniform memory architecture (NUMA) awareness.       |
+| HotSpot Option          | OpenJ9 Option                                    | Usage                                                     |
+|-------------------------|--------------------------------------------------|-----------------------------------------------------------|                                                                        
+| `-Xcomp`                | [`-Xjit:count=0`](xjit.md#count)**<sup>1</sup>** | `-Xcomp` disables interpreted method invocations.         |
+| `-Xgc`                  | [`-Xgcpolicy`](xgcpolicy.md)**<sup>2</sup>**     | Configuring your garbage collection policy.               |
+| `-XX:+UseNUMA`          | [`-Xnuma:none`](xnumanone.md)**<sup>3</sup>**    | Controls non-uniform memory architecture (NUMA) awareness.|
 
 <i class="fa fa-pencil-square-o" aria-hidden="true"></i> **Notes:**
 
--- a/docs/interface_jvmti.md
+++ b/docs/interface_jvmti.md
@@ -595,7 +595,7 @@ You can query VM log options by using the `QueryVmLogOptions()` API:
 
     jvmtiError QueryVmLogOptions(jvmtiEnv* jvmti_env, jint buffer_size, void* options, jint* data_size_ptr)
 
-This extension returns the current log options as an ASCII string. The syntax of the string is the same as the [`-Xsyslog`](xsyslog.md) command-line option, with the initial `-Xsyslog:` omitted. For example, the string "error,warn" indicates that the VM is set to log error and warning messages only. For more information, see [`-Xsyslog`](xsyslog.md).
+This extension returns the current log options as an ASCII string. The syntax of the string is the same as the `-Xsyslog` command-line option, with the initial `-Xsyslog:` omitted. For example, the string "error,warn" indicates that the VM is set to log error and warning messages only. For more information, see [`-Xsyslog`](xsyslog.md).
 
 **Parameters**
 
@@ -627,7 +627,7 @@ You can set VM log options by using the `SetVmLogOptions()` API:
 
     jvmtiError SetVmLogOptions(jvmtiEnv* jvmti_env, char* options_buffer)
 
-The log option is passed in as an ASCII character string. Use the same syntax as the [`-Xsyslog`](xsyslog.md) command-line option, with the initial `-Xsyslog:` omitted. For example, to set the VM to log error and warning messages, pass in a string containing "error,warn". For more information, see [`-Xsyslog`](xsyslog.md).
+The log option is passed in as an ASCII character string. Use the same syntax as the `-Xsyslog` command-line option, with the initial `-Xsyslog:` omitted. For example, to set the VM to log error and warning messages, pass in a string containing "error,warn". For more information, see [`-Xsyslog`](xsyslog.md).
 
 **Parameters**
 
--- a/docs/messages_intro.md
+++ b/docs/messages_intro.md
@@ -61,6 +61,9 @@ These messages can help you with problem determination.
 
 By default, all error and some information messages are routed to the system log and also written to `stderr` or `stdout`. The specific information messages are `JVMDUMP039I`, `JVMDUMP032I`, and `JVMDUMP033I`, which provide valuable additional information about dumps produced by the VM. To route additional message types to the system log, or turn off message logging to the system log, use the [`-Xsyslog`](xsyslog.md) option. The `-Xsyslog` option does not affect messages written to the standard error stream (stderr). See [OpenJ9 command-line options](cmdline_specifying.md).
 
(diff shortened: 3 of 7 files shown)


SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

docs/interface_jvmti.md
```diff
@@ -595,7 +595,7 @@ You can query VM log options by using the `QueryVmLogOptions()` API:
 
     jvmtiError QueryVmLogOptions(jvmtiEnv* jvmti_env, jint buffer_size, void* options, jint* data_size_ptr)
 
-This extension returns the current log options as an ASCII string. The syntax of the string is the same as the [`-Xsyslog`](xsyslog.md) command-line option, with the initial `-Xsyslog:` omitted. For example, the string "error,warn" indicates that the VM is set to log error and warning messages only. For more information, see [`-Xsyslog`](xsyslog.md).
```
pshipton, line 598: What's wrong with having links? It's easier to use a link than to lookup the option.

docs/cmdline_migration.md
```diff
@@ -73,19 +73,22 @@ You can use the following command-line options in OpenJ9, just as you did in Hot
...
+| `-XX:+UseNUMA`          | [`-Xnuma:none`](xnumanone.md)**<sup>4</sup>**    | Controls non-uniform memory architecture (NUMA) awareness.|
 
 <i class="fa fa-pencil-square-o" aria-hidden="true"></i> **Notes:**
 
 1. HotSpot uses `-Xcomp` to force compilation of methods on first invocation. However, this option is deprecated. Whilst it can be used for compatibility, using `-Xjit:count=0` is preferred.
 
 2. HotSpot uses `-Xgc` to both select policies and configure them; OpenJ9 uses `-Xgcpolicy` to select policies, reserving `-Xgc` for configuration.
 
-3. In HotSpot, NUMA awareness is turned off by default and is turned on by using the `-XX:+UseNUMA` option. Conversely, the OpenJ9 VM automatically enables NUMA awareness and uses `-Xnuma:none` to turn it *off*.
+3. To avoid confusion with the HotSpot `-Xlog` option, the OpenJ9 `-Xlog` option is superseded by the [`-Xsyslog`](xsyslog.md) option.
```
pshipton, line 89: I'm thinking that most people won't know or care there used to be an OpenJ9 -Xlog option that did something different. Perhaps "the OpenJ9 `-Xlog` option" should be quantified with "prior to the OpenJ9 0.24 release".

docs/messages_intro.md
```diff
@@ -61,6 +61,9 @@ These messages can help you with problem determination.
 
 By default, all error and some information messages are routed to the system log and also written to `stderr` or `stdout`. The specific information messages are `JVMDUMP039I`, `JVMDUMP032I`, and `JVMDUMP033I`, which provide valuable additional information about dumps produced by the VM. To route additional message types to the system log, or turn off message logging to the system log, use the [`-Xsyslog`](xsyslog.md) option. The `-Xsyslog` option does not affect messages written to the standard error stream (stderr). See [OpenJ9 command-line options](cmdline_specifying.md).
 
+<i class="fa fa-pencil-square-o" aria-hidden="true"></i> **Note:** The `-Xsyslog` option supersedes the `-Xlog` option.
```
pshipton, line 64: "the `-Xlog` option" should be quantified.

docs/xlog.md
```diff
@@ -22,21 +22,35 @@
...
+## Compatibility processing
 
-Other forms of `-Xlog` which should enable GC logging are also recognized, such as `-Xlog` or `-Xlog:all`, but since these options are intended to enable other forms of logging which are not supported, a warning is output `JVMJ9VM007W Command-line option unrecognised: <option>` in addition to enabling verbose GC logging . Similarly `-Xlog:gc:stdout` enables `-Xverbosegclog` but also outputs the warning. The `-Xverbosegclog` output goes to `stderr`, logging to `stdout` is not supported.
+Any garbage collection (GC) options passed to the `-Xlog` option are mapped to the equivalent OpenJ9 verbose GC command line options as described in the following list:
 
-<!-- ==== END OF TOPIC ==== xsyslog.md ==== -->
+- `-Xlog:gc` or `-Xlog:gc:stderr` is equivalent to [`-Xverbosegclog`](xverbosegclog.md)
+- `-Xlog:gc:file=<filename>` is equivalent to `-Xverbosegclog:<filename>`, where `<filename>` can contain the following tokens:
```
pshipton, line 40: Maybe `-Xverbosegclog:<filename>` should be `-Xverbosegclog:<updated filename>`, as `%p` and `%t` are processed before passing on the filename to `-Xverbosegclog`.

docs/interface_jvmti.md
SueChaplain: We don't need 2 links to the same topic in the same paragraph. This is not ideal from a UX perspective.

docs/cmdline_migration.md
SueChaplain: OK. Will qualify.

SueChaplain:
Jenkins doc stage

pshipton (approved):

SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

docs/xlog.md
```diff
@@ -22,21 +22,63 @@
...
+This option controls VM logging and is supported for compatibility with the reference implementation. However, only forms of `-Xlog` that enable garbage collection (GC) logging are recognized.
+
+From Eclipse OpenJ9 version 0.24.0, the [`-Xsyslog`](xsyslog.md) option replaces the existing OpenJ9 `-Xlog`  option for message logging to avoid conflicts with the reference implementation. For backward compatibility, you can control the behavior of the `-Xlog` option with the [`-XX:[+|-]LegacyXlogOption`](xxlegacyxlogoption.md) option.
 
 ## Syntax
 
-        -Xlog:gc[:stderr|:file=<filename>]
+        `-Xlog[:<parameters>]`
+
+Without any parameters, this command runs the `-Xsyslog` option and [`-Xverbosegclog`](xverbosegclog.md). Because the reference implementation of `-Xlog` is intended to enable other forms of logging that are not supported by the OpenJ9 VM, the following warning is printed in addition to enabling verbose GC logging:
```
pshipton, line 35: -Xlog doesn't "runs the `-Xsyslog` option" when `-XX:-LegacyXlogOption` is set (by default).

docs/xlog.md
```diff
@@ -22,21 +22,63 @@
...
+
+Setting `-Xlog:all` runs the `-Xsyslog` option and the `-Xverbosegclog` option. Because the reference implementation of `-Xlog` is intended to enable other forms of logging that are not supported by the OpenJ9 VM, the following warning is printed in addition to enabling verbose GC logging:
+
+```
+JVMJ9VM007W Command-line option unrecognised: <option>
+```
+
+### `-Xlog:gc[stderr]`
+
+Setting `-Xlog:gc` on the command line runs the `-Xsyslog` option and `-Xverbosegclog`.
```
pshipton, line 56: If `-XX:-LegacyXlogOption` then `-Xlog:<whatever>` has no relation to `-Xsyslog`.
If `-XX:+LegacyXlogOption` then `-Xlog` is the same as `-Xsyslog` where `-Xlog:gc` is not a valid option.

docs/xxlegacyxlogoption.md
```diff
@@ -24,14 +24,21 @@
...
+| `-XX:+LegacyXlogOption` | Enable legacy `-Xlog` behavior             |                                                                               |
+| `-XX:-LegacyXlogOption` | Process `-Xlog` requests for GC logging    | <i class="fa fa-check" aria-hidden="true"></i><span class="sr-only">yes</span>|
 
-<!-- ==== END OF TOPIC ==== xxdisableexplicitgc.md ==== -->
\ No newline at end of file
+## Explanation
+
+From Eclipse OpenJ9 0.24.0, the `-Xlog` option is replaced by the [`-Xsyslog`](xsyslog.md) option. The `-XX:[+|-]LegacyXlogOption` controls how the `-Xlog` option is processed.
+
+- When `-XX:-LegacyXlogOption` is set, the `-Xlog` option is not recognized unless garbage collection (GC) options are specified (for example, `-Xlog:gc[:stderr|:file=<filename>]`). For more information, see [`-Xlog`](xlog.md).
```
pshipton, line 40: Not strictly true, `-Xlog` and `-Xlog:all` are recognized as forms that enabled gc logging.

docs/xxlegacyxlogoption.md
```diff
@@ -24,14 +24,21 @@
 
 # -XX:[+|-]LegacyXlogOption
 
-Controls processing of the [`-Xlog`](xlog.md) option. When `-XX:-LegacyXlogOption` is set, which is the default, the `-Xlog` option is not recognized.
-When `-XX:+LegacyXlogOption` is set, the `-Xlog` option is equivalent to the [`-Xsyslog`](xsyslog.md) option. 
+Controls processing of the `-Xlog` option.
```
doveye, line 28: I suggest we start with the following content (or similar), then just have the table. I don't think we need to talk about what the "-" option does as that's already explained in the -Xlog topic (and if there's a change in future, we'd have to update two topics instead of one), plus the only users who need to know about this legacy option are those who want to turn it on.

> Controls the behavior of the -Xlog option. In Eclipse OpenJ9 0.24.0, the -Xlog option changed for better compatibility with the reference implementation. The old -Xlog behavior was transferred to the new -Xsyslog option. If you cannot use -Xsyslog for some reason, you can make -Xlog behave as it did in earlier releases by specifying -XX:+LegacyXlogOption.

docs/xlog.md
SueChaplain: >-Xlog doesn't "runs the -Xsyslog option" when -XX:-LegacyXlogOption is set (by default).

From the release topic:

"Setting -XX:-LegacyXlogOption (default) causes the VM to process the -Xsyslog option instead and process the -Xlog option only for garbage collection (GC) log requests (-Xlog:gc[:stderr|:file=<filename>])."

Are you objecting to "runs" instead of "processes" or is this deeper than that please?

docs/xxlegacyxlogoption.md
SueChaplain: >Not strictly true, -Xlog and -Xlog:all are recognized as forms that enable gc logging.

Maybe(?):

"When `-XX:-LegacyXlogOption` is set, the `-Xlog` option is not recognized unless a form of this option is run that requests garbage collection (GC) logging (for example, `-Xlog:gc[:stderr|:file=<filename>]`)."

docs/xlog.md
SueChaplain: Judging by your next comment, I think we need to update the release topic! What is in there currently is ambiguous. I took this to mean that it will attempt to generate log messages for you (by running -Xsyslog) but also ensure verbose GC logs are included if you are interested in that component. 

I'll have to cascade this change through other topics.

docs/xxlegacyxlogoption.md
SueChaplain: @pshipton - please see Esther's comment as an approach for explaining what we are achieving here. If we went with this, I'd have to update some of the other topics to adopt this. Would you support this?

In this case, we'd not talk too much about the `-XX` option in the `-Xlog` topic other than to reference it as a means of changing `-Xlog` behavior.

docs/xxlegacyxlogoption.md
pshipton: Yes, this is better.

docs/xxlegacyxlogoption.md
pshipton: Works for me.

docs/xlog.md
pshipton: -Xlog is either equivalent to -Xsyslog, or it can enable gc logging, it never does both. The change is incompatible, so anybody using the legacy -Xlog will be broken by default on OpenJ9 builds.

SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

pshipton (approved):

SueChaplain:
Jenkins doc stage

pshipton:
Thanks Sue!

PR #685, opened 2020-12-09
Align GC policy terminology and structure

SueChaplain:
- Apply agreed terminology changes to all policies
- Make structure consistent across areas
- Move "when to use" information from xgcpolicy option to main topic

### DIFF
--- a/docs/gc.md
+++ b/docs/gc.md
@@ -39,7 +39,7 @@ OpenJ9 provides several GC policies that are designed around different applicati
 
 Different policies require a Java heap that is configured in different ways in order to achieve different goals. The simplest configuration consists of a single area of memory, often referred to as a *flat* heap. Other configurations divide the heap into different areas or regions, which might contain objects of different ages (*generations*) or sizes.
 
-A GC cycle is a repeatable process that involves a set of GC operations. These operations process all or parts of the Java heap to complete a discrete function. For example, a *mark* operation traces all objects in the heap to determine which ones are reachable. A *sweep* operation runs to clear away unreachable objects. Together, a *mark* and *sweep*  operation are capable of reclaiming used memory as part of a GC cycle.
+A GC cycle is a repeatable process that involves a set of GC operations. These operations process all or parts of the Java heap to complete a discrete function. For example, a *mark* operation traces all objects in the heap to determine which ones are reachable. A *sweep* operation runs to clear away unreachable objects. Together, a *mark* and *sweep*  operation are capable of reclaiming used memory as part of a GC cycle. Not all GC cycles include operations to reclaim memory. For example, the `balanced` policy involves a global cycle that includes only a *mark* operation; reclaiming the memory with a *sweep* operation occurs as part of a separate partial GC cycle.
 
 A GC operation might complete in one step, or it might involve phases. For example, a *mark* operation consists of the following 3 phases:
 
@@ -55,28 +55,28 @@ The following table shows the heap configuration and the GC cycles and operation
 
 | Policy         | Heap configuration |  GC cycles / operations |
 |----------------|--------------------|-------------------------|
-|`gencon`        | Two areas: *nursery* and *tenure* <br> Two generation groups: new/older |Global GC cycle: concurrent *mark-sweep* operations, optionally followed by a *compact* operation <br> Partial GC cycle: STW *scavenge* operation or concurrent *scavenge* operation (if `-Xgc:concurrentScavenge` is set) |
-|`balanced`      | Multiple regions of equal size <br> Multiple generations: 24 |Global GC cycle: incremental concurrent *mark* operation (*global mark phase*), followed by a global STW *sweep* operation <br> Partial GC cycle: STW *copy forward* operation  or STW *mark-compact* operation |
-|`optavgpause`   | Single area: *flat* | Global GC cycle: concurrent *mark-sweep* operations, optionally followed by a *compact* operation |
-|`optthruput`    | Single area: *flat*  | Global GC cycle: STW *mark-sweep* operations, optionally followed by a *compact* operation|
-|`metronome`     | Multiple regions <br> Segregated by size class | Global GC cycle: incremental STW *mark-sweep* operation in small interruptible steps |
-|` nogc`         | Single area: *flat* | No GC cycles |
+|`gencon`        | Two areas: *nursery* and *tenure* <br> Two generation groups: new/older |Global GC cycle: concurrent *mark-sweep* operations, optionally followed by a *compact* operation <br> Partial GC cycle: STW *scavenge* operation or concurrent *scavenge* operation (if optionally enabled) |
+|`balanced`      | Multiple regions of equal size <br> Multiple generations |Global GC *mark* cycle: incremental concurrent *mark* operation (*global mark phase*) <br> Partial GC cycle: STW *copy forward* operation and optional *mark*, *sweep* or *compact* operations |
+|`optavgpause`   | One area: *flat* <br> One generation | Global GC cycle: concurrent *mark-sweep* operations, optionally followed by a *compact* operation |
+|`optthruput`    | One area: *flat* <br> One generation  | Global GC cycle: STW *mark-sweep* operations, optionally followed by a *compact* operation|
+|`metronome`     | Multiple regions by size class <br> One generation | Global GC cycle: incremental STW *mark-sweep* operation in small interruptible steps |
+|` nogc`         | One area: *flat* | No GC cycles |
 
 
-<i class="fa fa-pencil-square-o" aria-hidden="true"></i> **Note:** All OpenJ9 GC policies support compressed references on 64-bit platforms, which compresses heap pointers to 32 bits if the total heap size does not exceed the theoretical upper bound of 64 GB. Applications that require more heap space can select any heap size within the bounds imposed by the operating system and available system RAM, without using compressed references. For more information, see [compressed references](allocation.md#compressed-references).
+<i class="fa fa-pencil-square-o" aria-hidden="true"></i> **Note:** All OpenJ9 GC policies support compressed references on 64-bit platforms, which compresses heap pointers to 32 bits if the total heap size does not exceed the theoretical upper bound of 64 GB. Applications that require more heap space can select any heap size within the bounds imposed by the operating system and available system RAM, without using compressed references. For more information, see [compressed references](allocator.md#compressed-references).
 
 
 ### Policy selection and tuning
 
-The default GC policy is `gencon`, which suits a broad spectrum of applications. Choosing a different GC policy should be guided by the application dynamics and an observation of how the application interacts with the heap during startup and at steady state. To help with this analysis, all OpenJ9 GC policies are instrumented to collect a wide range of GC-related metric data for reporting in a GC log file.
+The default policy is the Generational Concurrent (`gencon`) GC policy, which suits a broad spectrum of applications. Choosing a different GC policy should be guided by the application dynamics and an observation of how the application interacts with the heap during startup and at steady state. To help with this analysis, all OpenJ9 GC policies are instrumented to collect a wide range of GC-related metric data for reporting in a GC log file.
 
-To enable GC logging for the OpenJ9 Java runtime, include the `-verbose:gc` option on the command line. By default, this option prints output to `stderr` but you can send the output to a log file by using [`-Xverbosegclog`](xverbsoegclog). You can then visualize the output by loading the GC log into the [Garbage Collector and Memory Visualizer (GCMV) plugin for Eclipse](https://marketplace.eclipse.org/content/ibm-monitoring-and-diagnostic-tools-garbage-collection-and-memory-visualizer-gcmv). OpenJ9 Java GC logs can also be analyzed by some online services, such as [GCEasy](https://gceasy.io/).
+To enable GC logging for the OpenJ9 Java runtime, include the `-verbose:gc` option on the command line. By default, this option prints output to `stderr` but you can send the output to a log file by using [`-Xverbosegclog`](xverbsoegclog). You can then visualize the output by loading the GC log into the [Garbage Collector and Memory Visualizer (GCMV)](https://marketplace.eclipse.org/content/ibm-monitoring-and-diagnostic-tools-garbage-collection-and-memory-visualizer-gcmv) plugin for the Eclipse IDE. OpenJ9 Java GC logs can also be analyzed by some online services, such as [GCEasy](https://gceasy.io/).
 
-To select a different GC policy, specify the [`-Xgcpolicy`](xgcpolicy.md) option on the command line. To adjust the initial and maximum size of the Java heap, use the [`-Xms` and `-Xmx`](xms.md) command line options. For generational GC policies, you can also set the [`-Xmn`, `-Xmns`, and `-Xmnx`](xmn.md) options.
+The following sections provide more information about each policy and when you might choose it for your application. To select a GC policy other than `gencon`, specify the [`-Xgcpolicy`](xgcpolicy.md) option on the command line. To adjust the initial and maximum size of the Java heap, use the [`-Xms` and `-Xmx`](xms.md) command line options. For generational GC policies, you can also set the [`-Xmn`, `-Xmns`, and `-Xmnx`](xmn.md) options.
 
 ### `gencon` policy (default)
 
-The Generational Concurrent GC policy ([`-Xgcpolicy:gencon`](xgcpolicy.md#gencon)) is probably best suited if you have a transactional application, with many short lived objects. This policy aims to minimize GC pause times without compromising throughput. This is the default policy employed by the VM, so if you want to use it you don't need to specify it on the command line when you start your application.
+The Generational Concurrent GC policy ([`-Xgcpolicy:gencon`](xgcpolicy.md#gencon)) is probably best suited if you have a transactional application, with many short-lived objects. This policy aims to minimize GC pause times without compromising throughput. This is the default policy employed by the VM, so if you want to use it you don't need to specify it on the command line when you start your application.
 
 If your application requires the allocation of objects of very different sizes and liveness on the Java heap, you might experience heap fragmentation, which in turn might lead to global heap compaction. In these circumstances, the [Balanced GC policy](xgcpolicy.md#balanced-policy) might be more appropriate.
 
@@ -103,7 +103,7 @@ The local GC scavenge reduces pause times by freqently reclaiming memory in the 
 
 #### Concurrent Scavenge
 
-A special mode of the `gencon` policy is known as *Concurrent Scavenge*. This mode aims to further reduce the average time spent in stop-the-world pauses by collecting nursery garbage in parallel with running application threads. Whilst aiming to reduce the average time, this mode does not improve the worst case pause time when compared to running `gencon` without Concurrent Scavenge enabled.
+A special mode of the `gencon` policy is known as *Concurrent Scavenge*. This mode aims to further reduce the average time spent in *stop-the-world* (STW) pauses by collecting nursery garbage in parallel with running application threads. Whilst aiming to reduce the average time, this mode does not improve the worst case pause time when compared to running `gencon` without Concurrent Scavenge enabled.
 
 To enable Concurrent Scavenge, see [-Xgc:concurrentScavenge](xgc.md#concurrentscavenge).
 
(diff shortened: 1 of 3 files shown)


SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

docs/gc.md
```diff
@@ -141,12 +141,14 @@ Object ages are tracked for each region with a maximum of 24 possible generation
...
+ The *eden* space is a set of regions of age 0, which contain the newest objects allocated. Regions of intermediate age serve as survivor regions to receive live objects moved from younger regions. When the eden space is full, a partial garbage collection (PGC) runs to reclaim space, typically by using a *copy forward* operation. In specific cases, a *mark-compact* operation might be used, for example, when there are not enough free survivor regions available. The PGC is a *stop-the-world* (STW) operation, after which new regions are allocated to create a new eden space. A PGC always includes the eden space, but might include older regions. Objects within collectible regions of intermediate age N are evacuated to survivor regions of age N+1. Regions of age 24 are included in PGC collection sets in order to defragment them.
```
amicic, line 144: Those couple of sentences about intermediate ages are not quite true. 
Firstly, survivor are both semi-populated regions (of ages from 1 to max age 24) and empty regions, if needed (and often are needed).
Secondly, objects from a region age N are evacuated into another region of same age N (or empty region) and then ages of all regions across heap (whether used as survivor or not) are incremented by 1. An exceptions is that the age of max age regions is never incremented (even though the may be part of PGC collection set).

docs/xgcpolicy.md
```diff
@@ -42,32 +42,31 @@ Controls the behavior of the garbage collector by specifying different garbage c
...
 ### `balanced`
 
         -Xgcpolicy:balanced
 
-: The balanced policy policy uses mark, sweep, compact and generational style garbage collection. The concurrent mark phase is disabled; concurrent garbage collection technology is used, but not in the way that concurrent mark is implemented for other policies. The `balanced` policy uses a region-based layout for the Java&trade; heap. These regions are individually managed to reduce the maximum pause time on large heaps and increase the efficiency of garbage collection. The policy tries to avoid global collections by matching object allocation and survival rates.
+: The Balanced policy requires a multi-region heap to manage multiple generations of objects. The policy uses a global GC cycle that involves an incremental concurrent *mark* operation (global mark phase), followed by *stop-the-world* (STW) *sweep* operation. The policy also uses a partial GC cycle to run *copy forward* or *mark-compact* operations. Regions are individually managed to reduce the maximum pause time on large heaps and increase the efficiency of garbage collection. The policy tries to avoid global collections by matching object allocation and survival rates.
```
amicic, line 60: You may add a detail that the global sweep followed by GMP is actually performed within the same increment of the first PGC after GMP is completed (not as as a part of last GMP cycle increment, as it's done in Gencon)

SueChaplain: I'm not sure I understand you completely @amicic, so please correct me if this sentence is wrong:

"With the `balanced` policy, the global and partial GC cycles interleave. That is, the global STW *sweep* operation runs within the same GC increment as the partial GC cycle."

SueChaplain:
Jenkins doc stage

docs/gc.md
SueChaplain: One more query @amicic. We mention that occasionally Balanced needs to run a global GC as an explicit GC. This is not in our table where we discuss heap configuration and GC cycles/ops. I think we ought to add it for completeness or are you happy that we discuss it only in the main Balanced section?

SueChaplain: This item already discussed. We won't put it in the table but we do mention it in the main section as a "rare event".

SueChaplain:
Jenkins doc stage

docs/gc.md
```diff
@@ -55,28 +55,28 @@ The following table shows the heap configuration and the GC cycles and operation
...
-|`optthruput`    | Single area: *flat*  | Global GC cycle: STW *mark-sweep* operations, optionally followed by a *compact* operation|
-|`metronome`     | Multiple regions <br> Segregated by size class | Global GC cycle: incremental STW *mark-sweep* operation in small interruptible steps |
-|` nogc`         | Single area: *flat* | No GC cycles |
+|`gencon`        | Two areas: *nursery* and *tenure* <br> Two generation groups: new/older |Global GC cycle: concurrent *mark-sweep* operations, optionally followed by a *compact* operation <br> Partial GC cycle: STW *scavenge* operation or concurrent *scavenge* operation (if optionally enabled) |
+|`balanced`      | Multiple regions of equal size <br> Multiple generations |Global GC *mark* cycle: incremental concurrent *mark* operation (*global mark phase*) <br> Partial GC cycle: STW *copy forward* operation and optional *mark*, *sweep* or *compact* operations |
+|`optavgpause`   | One area: *flat* <br> One generation | Global GC cycle: concurrent *mark-sweep* operations, optionally followed by a *compact* operation |
```
dmitripivkine, line 60: I am not sure how to improve this without overloading by facts (may be it is acceptable "as is" as outline). 
In reality there are two types of GC cycles on optavgpause:
- Concurrent Mark/STW phase (Final Mark + Partial Sweep)/Concurrent Sweep
- Concurrent Mark/STW phase (Final Mark, Sweep, Compact)

docs/gc.md
```diff
@@ -141,15 +141,17 @@ Object ages are tracked for each region with a maximum of 24 possible generation
...
 
-With the `balanced` policy, a global GC cycle is rare, occurring only in very tight memory conditions when PGC and GMP operations cannot free enough memory on the heap. These global GCs are explicit GCs, like those invoked by an application calling `System.gc()`.
+With the `balanced` policy, a global GC is sometimes required in addition to the global *mark* operations and PGC. This global GC cycle is rare, occurring only in very tight memory conditions when other GC cycles cannot free enough memory on the heap. This third global GC cycle is an explicit GC cycle, like those invoked by an application calling `System.gc()`.
```
dmitripivkine, line 148: This *third* global GC cycle ? Did I miss something here?

docs/xgcpolicy.md
```diff
@@ -42,32 +42,31 @@ Controls the behavior of the garbage collector by specifying different garbage c
...
 
 ### `gencon`
 
         -Xgcpolicy:gencon
 
-: The generational concurrent policy (default) uses a concurrent mark phase combined with generational garbage collection to help minimize the time that is spent in any garbage collection pause. This policy is particularly useful for applications with many short-lived objects, such as transactional applications. Pause times can be significantly shorter than with the `optthruput` policy, while still producing good throughput. Heap fragmentation is also reduced.
+: The generational concurrent policy (default) requires a heap that is divided into two main areas (*nursery* and *tenure*) to manage two generation groups (*new* and *older*). The policy uses a global GC cycle of concurrent *mark-sweep* operations, optionally followed by *compact* operations. The policy also uses a partial GC cycle to run *scavenge* operations on the *nursery* area. The partial cycle helps reduce the frequency and duration of the global GC cycle. Note that *scavenge* is a *stop-the-world* operation, unless `-Xgcpolicy:gencon` is specified with the [`-Xgc:concurrentScavenge`](xgc.md#concurrentscavenge) option (supported only on IBM Z&reg; systems).
```
dmitripivkine, line 51: `(supported only on IBM Z&reg; systems)` This is not accurate any more. IBM Z provides Hardware-based support for Concurrent Scavenge(z14 and higher). There in number of platforms with Software-based support   Please update correspondently (see `#### Concurrent Scavenge` section in gc.md)

docs/gc.md
SueChaplain: Let me remove the word "global" - I was referring to the GMP and PGC as the first 2 :-)

docs/xgcpolicy.md
SueChaplain: Oh yes .... the SW based update must have missed fixing the command ref. Will sort.

SueChaplain:
Jenkins doc stage

docs/gc.md
SueChaplain: Yes, I've had similar discussions with Aleks. We're keeping this table relatively high-level and need to elaborate in the detailed sections. Not sure it is all there yet, but a work in progress. However, I'm keen to get as much content is for 0.24.0 as possible and iterate on it afterwards.

SueChaplain:
Jenkins doc stage

docs/gc.md
amicic: 'is an explicit type' also needs correction. This is more correct:

'This third cycle type is also used with explicit GCs, like those....'

This cycle type when triggered in tight memory condition is considered implicit (since not invoked by user, but by internal GC mechanism).

docs/xgcpolicy.md
```diff
@@ -42,32 +42,31 @@ Controls the behavior of the garbage collector by specifying different garbage c
...
+: The Balanced policy requires a multi-region heap to manage multiple generations of objects. The policy uses a global GC cycle that involves an incremental concurrent *mark* operation (global mark phase), followed by *stop-the-world* (STW) *sweep* operation. The policy also uses a partial GC cycle to run *copy forward* or *mark-compact* operations. Regions are individually managed to reduce the maximum pause time on large heaps and increase the efficiency of garbage collection. The policy tries to avoid global collections by matching object allocation and survival rates. With the `balanced` policy, the global and partial GC cycles interleave. That is, the global STW *sweep* operation runs within the same GC increment as the first partial GC cycle that follows the global mark phase.
```
amicic, line 60: '....global mark phase increments and partial GC cycles interleave.'

or longer variant:

'....GMP cycle overlaps with partial GC cycles, so that GMP increments interleave with PGC cycle/increments'



Where/when global STW sweep runs has no relation with this interleaving behavior. So, the next sentence 'That is, then global STW sweep operation'.... should not lean on the previous one (should not start with 'That is...'). It's an independent statement.

docs/gc.md
```diff
@@ -141,15 +141,17 @@ Object ages are tracked for each region with a maximum of 24 possible generation
...
+Partial GC cycles reclaim free regions by moving objects out of *eden* and older regions. However, moving objects out of *eden* regions can never reclaim all of the consumed memory because some objects from *eden* regions will survive. The partial GC cycle compensates by moving objects to regions that contain older generations so that newer regions can be defragmented. However, over time the overall amount of defragmented memory decreases and records about object liveness in older regions becomes less accurate. Eventually, the work that is done by partial GC cycles to reclaim memory cannot keep pace with memory allocation. Free regions become so scarce that global *mark* operations (sometimes known as the *Global mark phase* (GMP)) are required to build a new record of object liveness across the heap. This record allows a global *sweep* operation to reclaim memory. The global *sweep* operation that is logically associated with the global *mark* operations, runs in the first partial GC cycle after the *mark* operations complete. Because the GC cycle responsible for the global *mark* operations runs concurrently, it might overlap and interleave with a few partial GC cycles.
```
amicic, line 146: This is not correct 

'The partial GC cycle compensates by moving objects to regions that contain older generations so that newer regions can be defragmented.'

what is reworded variant of my sentence:

'This incomplete reclamation can be compensated through evacuation of older regions through their defragmentation. '

The point is that older (not just Eden) regions become the source of reclaimed memory. For example, we include 5 fragmented older regions, evacuate them, and move objects into 2 new (previously empty regions. So the net effect is that we reclaimed 3 empty regions. 

Also this needs to be corrected:
... over time, he overall amount of fragmented (not defragmented) memory decreases....

amicic: I also originally stated this, that you omitted: 

`Additionally, a sweep operation is performed to quantify amount of fragmented memory in older regions based on newly built liveness information'

Sweep has dual functionality. First one is that it reclaims memor -, it may create completely empty regions. For those that are not completely freedd, the reclamation of free memory will occur later in PGCs through their defragmentation, as explained above. However, sweep for them has that the second role: it determines how much free (fragmented) memory there is, so that later PGCs know what/when to pick to defragmenent.

It's nice to mention that second role, but it's not mandatory, one reason being that we have plans in future to measure fragmentation within GMP, rather than sweep.

docs/gc.md
SueChaplain: @amicic I don't understand what you are saying. Is is explicit or implicit? One sentence seems to contradict the other.

docs/gc.md
SueChaplain: I'm still finding your input difficult to understand, but I've had a further pass at trying to get this right. I am still partly troubled that we are introducing a global *sweep* that we don't mention in the table. This *sweep* now does 2 things - measures fragmentation and later reclaims memory? Please take another look at the updated text.

SueChaplain:
Jenkins doc stage

docs/gc.md
amicic: If a GC cycle was triggered by internal mechanisms (allocation failure, taxation threshold reached) then we say it's implicit.
If it's triggered by user (directly by System.gc() or indirectly by for example requesting heap dump, when heap dump will be preceeded by a Global GC) we say it's explicit.

Global GC cycle type can be triggered by other of the two. It can be both explicit or implicit depending how it's triggered.

So, it's not correct to say: 'This third global GC cycle is an explicit GC cycle'. When we say 'is' it implies 1:1 relationship, but to me it's ok to say  'This third cycle type is also used with explicit GCs'. I'm trying to imply 1:n relationship, since this third type is used in implicit scenario, as already stated by 'occurring only in very tight memory conditions '.

docs/gc.md
amicic: Sweep is not a stand alone cycle, it's an operation physically attached to PGC cycle/increment, but logically associated with GMP cycle. So it definitely does not deserve a separate row in the table, but can be mentioned in either PGC or GMP definition, depending if physical or logical association is considered more important. I'm slightly more inclined to mention it as an operation in PGC, since PGC already lists most of operations it does.

docs/gc.md
SueChaplain: In that case perhaps that last sentence is redundant, where 'occurring only in very tight memory conditions' implies exactly what causes it and the user does not need the extra information.

docs/gc.md
```diff
@@ -141,15 +141,20 @@ Object ages are tracked for each region with a maximum of 24 possible generation
...
+The *eden* space is a set of regions of age 0, which contain the newest objects allocated. Regions of intermediate age or empty regions serve as survivor regions to receive live objects moved from younger regions. When the *eden* space is full, a partial GC cycle runs to reclaim space, typically by using a *copy forward* operation. In specific cases, a *mark-compact* operation might be used, for example, when there are not enough free survivor regions available. The partial GC cycle is a *stop-the-world* (STW) operation, after which new regions are allocated to create a new *eden* space. A partial cycle always includes the *eden* space, but might include older regions. Objects from collectible regions of age N are moved into another region of the same age N or to an empty region that is assigned an age of N. Then, the ages of all regions across the heap are incremented by 1, except for the maximum age 24 regions. Regions of age 24 are included in partial GC collection sets in order to defragment them.
```
amicic, line 144: 'Regions of intermediate age or empty regions serve as survivor regions to receive live objects moved from younger regions'

 can now be even omitted. This one states is very similar and is more correct:
 
Objects from collectible regions of age N are moved into another region of the same age N or to an empty region that is assigned an age of N.

docs/gc.md
```diff
@@ -141,15 +141,20 @@ Object ages are tracked for each region with a maximum of 24 possible generation
...
 
-Over time, transient objects drop out of the live set and viable survivor regions become scarce. This problem is solved by introducing a global mark phase (GMP) to mark and sweep the entire heap. GMP is a hybrid *incremental-concurrent* GC cycle. Each GMP increment is interleaved with a PGC cycle so that regions that are left sparse after sweeping the heap can be collected, cleared, and recycled during and after the GMP.
+Partial GC cycles work to reclaim free regions in the heap for allocating new objects. As much as 90% of this memory is reclaimed by moving objects out of *eden* regions to older regions.
```
amicic, line 146: 90% is just a typical ratio. 'as much as' sounds to me it's maximum

amicic: As much as 90% of this memory is reclaimed by moving objects out of *eden* regions to older regions. 

and 

However, because some objects from *eden* regions always survive, 

are tightly couple sentences and possibly some rewording is needed for proper cause-consequence relationship. Because some Eden object survive (~10%), PGC cannot reclaim all Eden worth of region (but only 90%).

For example, to me it would be better to start the second sentence without however, but like:

Since some objects from eden always survive

docs/gc.md
```diff
@@ -141,15 +141,20 @@ Object ages are tracked for each region with a maximum of 24 possible generation
...
 
-With the `balanced` policy, a global GC cycle is rare, occurring only in very tight memory conditions when PGC and GMP operations cannot free enough memory on the heap. These global GCs are explicit GCs, like those invoked by an application calling `System.gc()`.
+A global *sweep* operation also runs to reclaim memory. This operation creates further fragmented memory that partial GC cycles can work on. The global *sweep* operation, while logically associated with the global *mark* operation, runs in the first partial GC cycle after the *mark* operation completes. Because the GC cycle responsible for the global *mark* operation runs concurrently, it might overlap and interleave with a few partial GC cycles.
```
amicic, line 149: Consider expaning this sentence:

A global *sweep* operation also runs to reclaim memory, so that it create completely empty regions.

This sentence 

This operation creates further fragmented memory that partial GC cycles can work on.

is actually more similar to 

A *sweep* operation uses this record to measure the amount of free memory in fragmented older regions, which later partial GC cycles can act upon to move objects and reclaim free regions. 
 
and it can be omitted.

docs/gc.md
amicic: I think it's ok to have that sentence. We state that if explicit GC is invoked it's Global GC cycle that it will be performed. It brings new information, it's just that the current form is not correct. Again:

'global GC cycle is an explicit GC' is not quite correct, becuase global GC cycle can be also an implicit
'explicit GC is global GC cycle' is correct, since it cannot be PGC cycle or GMP cycle.

docs/gc.md
SueChaplain: An alternative, which is more in line with your suggestion:

_Because some objects from *eden* regions always survive, a partial GC cycle can reclaim only about  90% of this memory. To keep up with object allocation, partial GC cycles also reclaim free regions by defragmenting older regions_.

amicic: yes, that's perfect

docs/gc.md
SueChaplain: I think the confusion for me (and potentially users) is that we are talking about a *partial* GC cycle (defined as operating on part of the heap) and then explaining that it does a *global* sweep (defined as operating on the whole heap). I think it helps to explain that the sweep is logically associated with the global mark (phase),  but is triggered(?)/run(?) during a PGC?

SueChaplain:
Jenkins doc stage

docs/gc.md
SueChaplain: I think what we have now is fine.

docs/gc.md
SueChaplain: I think we should discuss explicit/implicit GC in the overview in more depth rather than introducing anything new here. Although it is just extra information, it does not make sense without more words, which are going to detract from the message we are providing in this paragraph. i.e. that this is another global GC that occurs rarely under tight memory restrictions. I've made a note to cover this in the overview topic, which I will be able to construct when this PR finally merges.

docs/gc.md
amicic: yes, table looks reasonable to me know. 

note that beside global sweep, also a local sweep can appear in PGC that operate only on collection set, like compact operation too. so having sweep in PGC in table is definitely reasonable. both sweep operations in VGC are named same though (just 'sweep') so it's not immediately clear which one is whicvh

docs/gc.md
```diff
@@ -141,72 +141,90 @@ Object ages are tracked for each region with a maximum of 24 possible generation
...
+Partial GC cycles work to reclaim free regions in the heap for allocating new objects. Because some objects from *eden* regions always survive, a partial GC cycle can reclaim only about  90% of this memory. To keep up with object allocation, partial GC cycles also reclaim free regions by defragmenting older regions. For example, a partial GC cycle that moves objects from 5 fragmented older regions into 2 empty regions, reclaims 3 regions for new object allocation. However, over time the overall amount of fragmented memory decreases and records about object liveness in older regions becomes less accurate. Eventually, the work done by partial GC cycles to reclaim memory cannot keep pace with memory allocation. Free regions become so scarce that a global *mark* operation (GMP) is required to build a new record of object liveness across the heap. A *sweep* operation uses this record to measure the amount of free memory in fragmented older regions, which later partial GC cycles can act upon to move objects and reclaim free regions.
```
amicic, line 146: grammar: records become (not becomes)

Perhpas it's more generic to say: .... cannot keep pace with memory consumption (rather than allocation). We are not talking just about mutator allocation, but also empty survivor region consumption by PGC or in other words difference between mutator allocation minus the recovered memory. In the example above, we did not really allocate 5 regions, but we also reclaimed 3, so the net is 2 regions consumed.

docs/gc.md
```diff
@@ -141,72 +141,90 @@ Object ages are tracked for each region with a maximum of 24 possible generation
...
 
-With the `balanced` policy, a global GC cycle is rare, occurring only in very tight memory conditions when PGC and GMP operations cannot free enough memory on the heap. These global GCs are explicit GCs, like those invoked by an application calling `System.gc()`.
+A global *sweep* operation also runs to reclaim memory so that it can create empty regions. The global *sweep* operation, while logically associated with the global *mark* operation, runs in the first partial GC cycle after the *mark* operation completes. Because the GC cycle responsible for the global *mark* operation runs concurrently, it might overlap and interleave with a few partial GC cycles.
```
amicic, line 148: instead of 'runs in the first partial GC cycle', this is more correct to say (what may help with your confusion, but then may introduce some new ones?):

runs within same STW increment as first partial GC (cycle)

SueChaplain:
Jenkins doc stage

PR #696, opened 2020-12-23
Add GC operations

SueChaplain:
- Add layout for GC operations and start to build the list
- Add further content to Allocation topic

Signed-off-by: SueChaplain sue_chaplain@uk.ibm.com


SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

docs/allocation.md
```diff
@@ -0,0 +1,152 @@
...
+
+Depending on your application workload or service level agreement, you can choose from a number of OpenJ9 *garbage collection (GC) policies*. Each GC policy uses a different strategy to manage memory on the heap. The structure of the heap also depends on the strategy in force. For more information about choosing a GC policy, see [Garbage collection policies](gc.md).
+
+## The allocator
+
+The allocator manages pools of free memory and how the free memory is consumed. It is also responsible for allocating areas of storage in the Java heap for objects at the request of applications, class libraries, or the VM.
+
+Every allocation requires a *heap lock* to stop concurrent threads trying to access the same area of memory at the same time. When an object is allocated, the heap lock is released. If there is insufficient space to allocate the object, allocation fails, the heap lock is released, and the GC is called. If the GC manages to recover some space on the heap, the allocator can resume operations. If the GC does not recover enough space, it returns an `OutOfMemoryError` exception.
```
amicic, line 40: I know this is old text, but consider 'Generally' instead of ' Every' (because, in the next paragraph we will state it's not wise to do it every time....

Instead of 'to stop',  consider to 'to synchronize'

docs/allocation.md
```diff
@@ -0,0 +1,152 @@
...
+
+Acquiring a heap lock for every allocation would be an intensive operation with a knock on impact to performance. To get around this problem, small objects are allocated to allocation caches.
+
+### Allocation caches
+
+To improve performance, allocation caches are reserved in the heap for different threads. These allocation caches are known as thread local heaps (TLH) and allow each thread to allocate memory from its cache without acquiring the heap lock. A TLH is typically used for small objects of less than 512 bytes (768 bytes on 64-bit VMs) although larger objects can be allocated from the cache if there is sufficient space.
```
amicic, line 46: This only partially true: 'A TLH is typically used for small objects of less than 512 bytes (768 bytes on 64-bit VMs) although larger objects can be allocated from the cache if there is sufficient space.'

This is more correct (and partially covers the aspect of TLH refresh):

Objects are attempted to allocate from TLH regardless of its size. If TLH cannot satisfy the allocation, depending on the size left in TLH (and some internal heuristics), and the size of the allocation, the allocation may proceed as individual allocation directly from the heap (typical for larger objects, certainly true for objects larger than TLH max) or we may proceed to refresh TLH (discard the existing one and create new one) and allocate from the new TLH.

docs/allocation.md
```diff
@@ -0,0 +1,152 @@
...
+A TLH is predefined with an initial default size of 2 KB. On every TLH refresh, the requested size for that thread
+is increased by an increment (default 4 KB). The requested size can grow up to a predefined maximum (default 128 KB).
+
+After every garbage collection cycle, the TLH requested size for each thread is reduced, sometimes by as much as 50%, to take account of threads that
+reduce their allocation rate and no longer need large TLHs.
+
+For very inactive threads, the requested size can even drop below the initial value, down to the predefined minimum (512/768 bytes).
+For very active threads, the maximum TLH requested size might be reached before the next GC occurs.
+
+Larger TLHs can help reduce heap lock contention, but might also reduce heap utilization and increase heap fragmentation.
```
amicic, line 59: I don't see it explicitly stated, since it may be obvious and or implied from previous section, but I'll mention it anyway:

When TLH refresh is done, heap lock is used.

Failing to refresh TLH is treated the same as failing to satisfy an individual allocation: it triggers a GC.

docs/allocation.md
```diff
@@ -0,0 +1,152 @@
...
+Some GC policies subdivide a *flat* heap or an *area* of the heap for object allocation into the Small Object Area (SOA) and the Large Object Area (LOA).
+
+The allocator initially attempts to allocate objects in the SOA, regardless of size. If the allocation cannot be satisfied the following actions are possible, depending on object size:
+
+- If the object is smaller than 64 KB, an allocation failure occurs, which triggers a GC action.
+- If the object is larger than 64 KB, the allocator attempts to allocate the object in the LOA. If the allocation cannot be satisfied, an allocation failure occurs, which triggers a GC action.
+
+The GC action that is triggered by the allocation failure depends on the GC policy in force.
+
+The overall size of the LOA is calculated when the heap is initialized, and recalculated at the end of each global GC cycle. The GC can expand or shrink the LOA, depending on usage, to avoid allocation failures.
```
amicic, line 89: We far more often use (in conversation, in code, VGC....) term 'contract' than 'shrink' (not just for LOA, but any type of heap resizing). There are VGC stanzas that report heap resizing elements, and the possible types are expand and contract. Consider aligning the documentation with it.

docs/allocation.md
```diff
@@ -0,0 +1,152 @@
...
+
+Depending on the memory management strategy in force, the Java heap can be configured in a number of ways. The simplest configuration consists of a single area of memory, often referred to as a *flat* heap. Other configurations divide the heap into different *areas* or *regions*, which might contain objects of different ages (generations) or sizes.
+
+### Area-based heaps
+
+The default GC policy for OpenJ9 uses a heap configuration that is divided into two main areas: the *nursery* area for new object allocation, and the *tenure* area for objects that continue to be reachable for a longer period of time. Most objects have short lifecycles and can be reclaimed by the garbage collector more quickly by focusing only on the nursery area. Global GC cycles that cause application pauses in order to clear and defragment the *tenure* area are less frequent.
+
+### SOA and LOA
+
+Some GC policies subdivide a *flat* heap or an *area* of the heap for object allocation into the Small Object Area (SOA) and the Large Object Area (LOA).
```
amicic, line 80: Actually, but default, all (not some) area based heaps will have LOA.

docs/allocation.md
```diff
@@ -0,0 +1,152 @@
...
+- [`-Xgc:tlhMaximumSize=<bytes>`](xgc.md#tlhmaximumsize)
+- [`-Xgc:tlhInitialSize=<bytes>`](xgc.md#tlhinitialsize)
+- [`-Xgc:tlhIncrementSize=<bytes>`](xgc.md#tlhincrementsize)
+
+Typically, when the maximum TLH size is increased, you should also increase the increment proportionally, so that active threads can
+reach the maximum requested TLH size more quickly.
+
+## Heap configuration
+
+Depending on the memory management strategy in force, the Java heap can be configured in a number of ways. The simplest configuration consists of a single area of memory, often referred to as a *flat* heap. Other configurations divide the heap into different *areas* or *regions*, which might contain objects of different ages (generations) or sizes.
```
amicic, line 72: Consider adding some attributes even this early (but could be emphasized later, too):

....small number of variable sized areas or large number of fixed size regions.

docs/allocation.md
```diff
@@ -0,0 +1,152 @@
...
+
+You can control the size of the LOA by using the [`-Xloainitial`, `-Xloaminimum`, and `-Xloamaximum`](xloaminimum.md) command line options. If the LOA is not used, the GC shrinks the LOA after a few cycles, down to the value of `-Xloaminimum`. You can also specify [`-Xnoloa`](xloa.md) to prevent an LOA being created.
+
+An SOA and LOA are used by the OpenJ9 GC policies: `gencon`, `optavgpause` and `optthruput`. For the `gencon` policy, the LOA and SOA are contained within the *tenure* area, which is designated for ageing objects. For more information about policies, see [Garbage collection policies](gc.md).
+
+### Region-based heaps
+
+The Java heap can also be subdivided into multiple regions. The `balanced` GC policy uses a heap that is divided into thousands of equal size regions in order to manage multiple generations of objects. The `metronome` GC policy also uses multiple regions, but each region is segregated by class size to manage a singe generation of objects. Both of these policies also use structures called *arraylets* to store large arrays in the heap.
```
amicic, line 97: Some facts you may want to add

- region size is determined at JVM startup, and the size remains fixed throughout the run
- regions size is a power of 2
- for metronome, regions size defaults to 64KB, unless otherwise specified (by cmdline option)
- for balanced, regions size is a function of _maximum_ heap size (Xmx value), and is chosen such that there are 1024 and 2047 regions, except for very small heaps where there could be less regions, due to min region size of 512KB.
- for metronome, region count is fixed (heap in metronome is fully expanded from the start and never contracts)
- for balanced, actually number of active regions will depend of actual size of heap, but at most (when/if heap is fully expanded) there will be between 1024-2047 regions (as determined at startup)

amicic: I'm not the originator of term 'segregate' (which indeed internally is used for metronome heap), but I think it's correct to say that 'heap is segregated (into subsets of regions of same size size-class)', not 'region is segregated'.

docs/allocation.md
```diff
@@ -0,0 +1,152 @@
...
+
+## Heap sizing
+
+The overall size of the Java heap is determined by two command-line options, `-Xms`, which sets the initial size of the heap, and `-Xmx`, which sets the maximum size of the heap. Finer tuning of the heap depends on the heap configuration that is being used by a GC policy. For example, an LOA within the heap can be sized by using the [`-Xloainitial`, `-Xloaminimum`, and `-Xloamaximum`](xloaminimum.md) command-line options. A *nursery* area within the heap can be sized by using the [`-Xmn`, `-Xmns`, and `-Xmnx`](xmn.md) command-line options. For more information about policies and the heap configurations that are used, see [GC policies](gc.md). To determine the values that are in use for the Java heap, use the `-verbose:sizes` option when you run your Java application.
+
+### Expansion and shrinkage
```
amicic, line 117: as mentioned previous, consider using 'contraction'

docs/allocation.md
```diff
@@ -0,0 +1,152 @@
...
+- `-Xminf` determines the minimum proportion of the heap that must be free after garbage is removed.
+- `-Xmaxf` determines the maximum proportion of the heap that must be free after garbage is removed.
+
+If expansion is required, the amount of memory that the heap can expand by is controlled by the following command-line option:
+
+- `-Xmine` sets the minimum amount that the heap can expand by.
+- `-Xmaxe` sets the maximum amount that the heap can expand by. The default is unlimited expansion up to the maximum heap size (`-Xmx`).
+
+Expansion can also be triggered if more time is being spent on GC processing than is specified by the `-Xmaxt` option. In this case, the heap expands by an amount that provides 17% more free space, within the limits imposed by the `-Xmine` and `-Xmaxe` values.
+
```
amicic, line 134: The above sets of options apply to:
- total heap in flat configuration (optthruput/optavgpause) and balanced
- only tenure area in generational heap (gencon)

Nursery (in gencon) resizing has it's own/rules and options to control (https://www.eclipse.org/openj9/docs/xgc/#dnssexpectedtimeratiomaximum). 

Eden (in balanced) has it's own rules , too. We are actively changing this behavior (both eden and total in balanced) - I'd wait to finish (in a couple of months) before documenting it.

docs/allocation.md
```diff
@@ -0,0 +1,152 @@
...
+When the heap shrinks, physical memory is not released unless paging is supported by the underlying operating system.
+
+
+### Compressed references
+
+On 64-bit systems, the VM can use compressed references to decrease the size of Java objects and make better use of the available space in the Java heap. By storing objects in a 32-bit representation, the object size is identical to a 32-bit object, which occupies a smaller memory footprint. These 4 byte (32-bit) compressed references are converted to 64-bit values at runtime with minimal overhead. Smaller objects enable larger heap sizes that result in less frequent garbage collection and improve memory cache utilization. Overall, the performance of 64-bit applications that store compressed rather than uncompressed 64-bit object references is significantly improved.
```
amicic, line 142: Wording does not seem right here:
'By storing objects in a 32-bit representation, the object size is identical to a 32-bit object, which occupies a smaller memory footprint.'

Should be (i'm maybe being a bit too verbose, though):

By storing _references_ in a 32-bit representation, the total object size is identical to that in a 32-bit VM, which occupies a smaller memory footprint than that in 64-bit VM.

Also I don't know if it's in nature of English to say 'occupies footprint'. It's more natural to me 'creates footprint', or 'occupies memory' (without word footprint).

docs/gc.md
```diff
@@ -83,17 +76,17 @@ The *nursery* area is subdivided into two further areas, the *allocate* space an
 3. A local GC scavenge process runs and reachable objects are either copied into the *survivor* space or into the *tenure* area if they have reached *tenure age*. Any objects that can't be reached are left untouched and subsequently cleared.
```
amicic, line 76: This is mostly terminology thing, but allocate space during GC time is called evacuate. It may not be that interesting for STW variant to mention (while we still consider it evacuate, it's not really exposed), but when/if we elaborate concurrent variant, some part of VGC logs will actually expose it. Just for you, here is an example (and don't ask me about allocate/surivor hybrid now - it's very CS specific)

'''    
<mem type="nursery" free="401962224" total="805306368" percent="49">
      <mem type="allocate/survivor" free="401962224" total="402653184" percent="99" />
      <mem type="evacuate" free="0" total="402653184" percent="0" />
</mem>
```

All that to say, you may want to expand this paragraph with a few words to introduce the notion of evacuate, but we can also defer it and expose it only in CS specific documentation.

btw, we also sometime say 'evacuate' instead of 'copied'.

 3. A local GC scavenge process runs and reachable objects are either copied  from *allocate* space (what is referred as evacuate during scavenge time) into the *survivor* space or into the *tenure* area if they have reached *tenure age*.

File: docs/gc.md [hunk 12]
```diff
@@ -83,17 +76,17 @@ The *nursery* area is subdivided into two further areas, the *allocate* space an
...
+The relative sizes of the *allocate* and *survivor* spaces are dynamically adjusted by a technique called *tilting*. When the *nursery* area is first created, it is evenly divided between the *allocate* and *survivor* spaces. If, after a GC *scavenge* operation runs, the amount of space required for the *survivor* area is comparatively small, the boundary between the two spaces is adjusted by *tilting*. For example, if the *survivor* space requires only 10% of the *nursery* area, the *tilt ratio* is adjusted to give 90% of the *nursery* area to the *allocate* space. With more space available for new objects, a global GC cycle can be delayed.
```
amicic, line 79: Conclusion is not quite right: With more space available for new objects, a global GC cycle can be delayed.

With tilting, we are not distributing the space between tenure and nursery, but only within nursery. 

The right conclusion is:
With more space available to allocate (out of whole nursery), we decrease frequency of scavenge cycles.

The original conclusion is indirectly correct, but we would need 2 things to happen: both tilt to increase allocate within nursrey, and contract nursery (to reduce allocate to its pre-tilt size) to allow tenure to grow (and only if heap is already expanded) for how much nursery decreased. But I don't feel we need to elaborate indirect/potential consequences.

docs/gc.md
```diff
@@ -83,17 +76,17 @@ The *nursery* area is subdivided into two further areas, the *allocate* space an
...
+The *tenure age* of an object is determined by the VM and reflects the number of times that an object has been copied between the *allocate* space and the *survivor* space. The age is in the range 1 - 14 and is adjusted dynamically by the VM depending on the overall amount of space that is used in the *nursery area*. For example, if an object has a tenure age of `5`, it has been copied backwards and forwards between *allocate* and *survivor* spaces five times. If the VM sets a *tenure age* of `5` based on the percentage of space remaining in the *nursery* area, the next *scavenge* moves the object from the *nursery* to the *tenure* area. You can set an initial *tenure age* with the [`-Xgc:scvTenureAge`](xgc.md#scvtenureage) option. You can also prevent the VM dynamically adjusting the *tenure age* by setting the [`Xgc:scvNoAdaptiveTenure`](xgc.md#scvnoadaptivetenure) option so that the initial age is maintained throughout the run time of the VM.
```
amicic, line 81: Actual heuristics for determining tenure age are rather complicated, but it would be more correct to say:

'... adjusted dynamically by the VM depending on the object survival rate at particular ages'

or more precisely (but probably overkill):

'... adjusted dynamically by the VM based on changes in object survival rates from one age to the next one'
(in math terms we are looking into first derivative).

docs/gc.md
```diff
@@ -83,17 +76,17 @@ The *nursery* area is subdivided into two further areas, the *allocate* space an
...
 
-The local GC scavenge reduces pause times by freqently reclaiming memory in the *nursery* area which, for a transactional application with many short-lived objects, has the most recyclable space. However, over time the *tenure* area might become full. So, whilst a local GC scavenge process is operating on the *nursery* area, a concurrent global GC process also runs alongside normal program execution to mark and remove unreachable objects from the *tenure* area. These two GC approaches combine to provide a good trade-off between shorter pause times and consistent throughput.
+The partial GC cycle (*scavenge*) reduces pause times by frequently reclaiming memory in the *nursery* area which, for a transactional application with many short-lived objects, has the most recyclable space. However, over time the *tenure* area might become full. So, whilst a partial GC cycle is operating on the *nursery* area, a concurrent global GC cycle also runs alongside normal program execution to mark and remove unreachable objects from the *tenure* area. These two GC approaches combine to provide a good trade-off between shorter pause times and consistent throughput.
```
amicic, line 85: There is a missing sentence to connect the flow.

... has the most recyclable space. While most of the objects will stay in nursery after scavenge is complete, a small fraction (above tenure age) will be moved to tenure.  Over time the *tenure* area might become full....

docs/gc.md
```diff
@@ -131,32 +124,19 @@ Object ages are tracked for each region with a maximum of 24 possible generation
...
+The *eden* space is a set of regions of age 0, which contain the newest objects allocated. When the *eden* space is full, a partial GC cycle runs to reclaim space, typically by using a *copy forward* operation. In specific cases, *mark* and *compact* operations might be used, for example, when there are not enough free survivor regions available. The partial GC cycle is a STW operation, after which new regions are allocated to create a new *eden* space. A partial cycle always includes the *eden* space, but might include older regions. Objects from collectible regions of age N are moved into another region of the same age N or to an empty region that is assigned an age of N. Then, the ages of all regions across the heap are incremented by 1, except for the maximum age 24 regions. Regions of age 24 are included in partial GC collection sets in order to defragment them.
```
amicic, line 127: This part seems redundant: ', after which new regions are allocated to create a new *eden* space."

One thing to realize is that eden space is not created just after a PGC (as it is the case with allocate in scavenge&nursery). Eden  is just a collection of regions (that could be scattered all over heap) that have recently allocated objects. Those regions are picked from the pool of all empty regions as allocation progresses. PGC just reclaims memory and create empty regions that are put to that pool.

PGC is therefore triggered when eden size threshold (region count) is reached, not when specific predetermined area (what one would assume is eden space) of heap is full, because eden space has not been predetermined. 

Similar thing could be said for survivor regions. They don't really exist - they are not predetermined. They are just consumed from the same global pool of free regions, but just during PGC time. As said in next paragraph, PGC at the end of cycle will return some (most, but not all) of those consumed regions back to the pool.

This is one important advantage of Balanced over Gencon (generally, region based GCs over area based), and could be emphasis: survivor and allocate/eden areas are not  predetermined and share free memory from the global pool of free regions, what in turn reduces probability of (too) late concurrent mark cycles and aborted partial GCs.

docs/gc_overview.md
```diff
@@ -23,81 +23,183 @@
 -->
 
 
-# Memory management overview
+# Garbage collection
 
-The process of managing memory in the VM is handled by the Allocator and the Garbage Collector. These components operate on an area of memory that is reserved for VM processing called the Java&trade; heap.
+To prevent applications running out of memory, objects in the Java heap that are no longer required must be reclaimed. This process is known as *garbage collection* (GC). When garbage is collected, the garbage collector must obtain exclusive access to the heap, which causes an application to pause while the clean up is done. This pause is often referred to as a *stop-the-world* (STW) pause because an application must halt until the process completes. In general, the first step in the GC process is to *mark* the objects that are reachable, which means they are still in use. The next step is to *sweep* away the unmarked objects to reclaim memory. The final step is to *compact* the heap if the heap is badly fragmented.
```
amicic, line 28: clean up (verb) -> cleanup (noun)

docs/gc_overview.md
```diff
@@ -23,81 +23,183 @@
...
 
+A GC cycle is a repeatable process that involves a set of GC operations. These operations process all or parts of the Java heap. When operating on the
+whole of the Java heap, the cycle is referred to as a *global GC cycle*; When operating on part of the heap, the cycle is referred to as a *partial GC cycle*.
 
-## The Allocator
+A global GC cycle can be triggered explicitly or implicitly according to the following rules:
 
-The Allocator assigns areas of the Java heap for Java objects. Objects are considered as *live* when they have a chain of references to them that start from root references, such as those found in thread stacks. When that reference or pointer no longer exists, the objects are considered as *garbage*.
+- A global GC cycle is triggered implicitly if it occurs because of internal mechanisms, such as an allocation failure or a taxation threshold being reached.
```
amicic, line 35: You may some examples now, or in elsewhere (in corresponding GC operations). This is a nice comparison:
- taxation threshold being reached is the one use to trigger PGC (since there is not predetermined eden space, we would end up consuming ALL empty regions before hitting AF - we have to trigger PGC earlier using the threshold),
-  allocation failure is used for scavenge (since allocate is predetermined area of heap where allocation threads allocate from, this mechanism works).
 
Some other examples:
The former is also used to trigger GMP increments (between interleaving PGCs)
The latter is used to trigger global GC in optthruput.

docs/gc_overview.md
```diff
@@ -23,81 +23,183 @@
...
+## GC operations
 
-Acquiring a heap lock for every allocation would be an intensive operation with a knock on impact to performance. To get around this problem, small objects are allocated to thread local heaps (TLH).
+GC operations run discrete functions on the Java heap. For example, a *mark* operation traces all objects in the heap to determine which ones are reachable. A *sweep* operation runs to clear away unreachable objects. Together, a *mark* and *sweep* operation are capable of reclaiming used memory as part of a GC cycle. Not all GC cycles include operations to reclaim memory. For example, the `balanced` GC policy involves a global cycle that includes only a *mark* operation; reclaiming the memory with a *sweep* operation occurs as part of a separate partial GC cycle.
```
amicic, line 42: You can expand that last sentence:

.... which can also reclaims memory through evacuation of young regions and de-fragmenation of older regions.

docs/gc_overview.md
```diff
@@ -23,81 +23,183 @@
...
-Larger TLHs can help reduce heap lock contention, but might also reduce heap utilisation and increase heap fragmentation.
+    A root object is an object that is accessible from outside of the heap such as a stack, class static field, or JNI reference. For other objects in the heap to be reachable, they must retain a connection from a root object. In the initial phase, tracing identifies all root objects by running a root scan. Root objects are pushed onto a work stack for processing in the next phase.
 
-The following options control the requested TLH size:
+2. Main phase
 
-- [`-Xgc:tlhMaximumSize=<bytes>`](xgc.md#tlhmaximumsize)
-- [`-Xgc:tlhInitialSize=<bytes>`](xgc.md#tlhinitialsize)
-- [`-Xgc:tlhIncrementSize=<bytes>`](xgc.md#tlhincrementsize)
+    The list of reachable root objects in the work stack is recursively traced for references to other objects in the heap. If new objects are found, they are added to the work stack. If an object is reachable, the appropriate bit is set in the *mark map*.
```
amicic, line 60: Maybe not be obvious to everyone, so you may clarify how we know that _new_ object is found: it's if not marked yet.
So, we will add an object to work stack, only if its bit in mark map is not set (otherwise we would not never complete the process, if there are cycles in the object graph).

docs/gc_overview.md
```diff
@@ -23,81 +23,183 @@
...
 
-Typically, when the maximum TLH size is increased, you should also increase the increment proportionally, so that active threads can
-reach the maximum requested TLH size more quickly.
+3. Final phase
 
-### SOA and LOA
+    The final phase processes weakly reachable roots such as finalizable objects, weak references, monitor sets, and string sets. If reachable objects are found, the appropriate bit is set in the *mark map*. For more information about the processing of *soft*, *weak*, and *phantom* references, see [Weak reference processing](#weak-reference-processing).
```
amicic, line 64: While this is generally correct statement, this is not what defines this processing step:
If reachable objects are found, the appropriate bit is set in the *mark map*.

GC maintains special sets of root objects called weakly reachable roots such as finalizable objects, weak references..... Those roots do not determine liveness of objects. Those objects may or may not survive by virtue of having direct or indirect references from standard (strong) roots, as found by the main tracing phase.
If such object survives, nothing special is done. But if does not survive, it is removed from its weak root set, and optionally some special action is performed (such as reference enqueuing or execution of finalization method).

docs/allocation.md
SueChaplain: All instances of updated to use the term *contract* or *contraction*.

docs/allocation.md
SueChaplain: Adapted to "By storing objects in a 32-bit representation, the object size is identical to that in a 32-bit VM, which creates a smaller memory footprint."

docs/allocation.md
SueChaplain: Changed to "All area-based heaps subdivide part of the heap into the Small Object Area (SOA) and the Large Object Area (LOA)."

docs/allocation.md
SueChaplain: Adapted, which I hope gives the user enough. If not, I can expand:

"Objects are allocated from the TLH unless there is insufficient space remaining in the TLH to satisfy the request. In this situation, the allocation might proceed directly from the heap by using a heap lock or the TLH might be refreshed to increase the size of the cache."

docs/allocation.md
SueChaplain: Adapted:

"In general, allocation requires a *heap lock* to synchronize concurrent threads that try to access the same area of memory at the same time."

docs/allocation.md
SueChaplain: >Some facts you may want to add

I think we capture most of these in the policy topic, but I'll double check and make sure we aren't missing anything. For this topic, I will add a link so that users can read more about the specifics for each policy.

docs/gc.md
SueChaplain: Modified:

"3. A local GC scavenge process runs and reachable objects are either *evacuated* (copied) into the *survivor* space or into the *tenure* area if they have reached *tenure age*."

docs/gc.md
SueChaplain: Corrected: " With more space available for new objects, the frequency of *scavenge* cycles is reduced."

docs/gc.md
SueChaplain: Hopefully, this is enough to fix:

"The *eden* space is a set of regions of age 0, which contain the newest objects allocated. The size of the *eden* space is determined by the number of regions that it contains. When the region count for the *eden* space reaches a predetermined threshold, a partial GC cycle runs to reduce the number of regions, typically by using a *copy forward* operation. Empty regions can then be assigned to the *eden* space from the pool."

docs/gc_overview.md
SueChaplain: Expanded:

"For example, the `balanced` GC policy involves a global cycle that includes only a *mark* operation; reclaiming the memory with a *sweep* operation occurs as part of a separate partial GC cycle that *evacuates* younger regions and defragments older regions."

docs/gc_overview.md
SueChaplain: Updated:

"Objects that are not marked are new objects and these are added to the work stack."

docs/gc_overview.md
SueChaplain: OK, I removed that sentence in each operation that had it listed.

docs/allocation.md
SueChaplain: Updated the following a couple of paragraphs up:

"If a TLH refresh fails to complete, a GC cycle is triggered."

docs/allocation.md
SueChaplain: As this is the opening paragraph I'd like to keep it simple and work these complexities in a little later.

docs/allocation.md
SueChaplain: Double checked and all those facts are in the policy descriptions.

docs/allocation.md
SueChaplain: I've saved this comment to an issue we are working on to overhaul the options ... seemed appropriate if we are likely to have a change in behavior soon.

docs/gc.md
SueChaplain: Went with:

"The age is in the range 1 - 14 and is adjusted dynamically by the VM depending on the object survival rate at particular ages."

docs/gc.md
SueChaplain: Changed:

"...has the most recyclable space. While most of the objects stay in the *nursery* area after the *scavenge* operation is complete, a small fraction are moved to the *tenure* area. Over time...."

Previous para explains tenure age, so I don't think we need to spell this out here.

docs/gc_overview.md
SueChaplain: - Added taxation threshold as the trigger for the balanced PGC in "copy-forward operations" section
- Added allocation failure as the trigger in the "scavenge operation" section
- Added taxation threshold as the trigger to the GMP section of balanced in policy topic
- Added allocation failure as the trigger for optthruput in policy topic.

SueChaplain:
Jenkins doc stage

SueChaplain:
Thanks for the review Aleks. Updates are now staged at https://eclipse.github.io/openj9-docs-staging/696/allocation/

SueChaplain:
I've taken the **WIP** prefix off of this PR now as I believe all review comments have been addressed and this should be ready to merge. @amicic - please let me know if you are happy with the changes. Thanks.

docs/gc.md
amicic: runs to reduce the number of _used (or occupied)_ regions

docs/allocation.md
amicic: ' to increase the size of the cache.' is not quite correct and probably can be dropped altogether (if you are already trying to keep it simple). If we refresh TLH it's just simply to get a new empty TLH (to be able to satisfy ongoing allocation request). TLH increase is just a secondary effect that may occur, if we have not reached TLH max size already. If we did reach, the refresh will occur with that same TLH size as before.

I'd still consider adding 'for larger objects' somewhere  for the case of 'the allocation might proceed directly from the heap by using a heap lock' (what would imply that TLH will be refreshed for the other case of smaller objects).

docs/openj9_newuser.md
```diff
@@ -36,9 +36,9 @@ For a list of compatible options, equivalent options, and options that need to b
...
 
-Eclipse OpenJ9 has a number of GC policies designed around different types of applications and workloads. By default, OpenJ9 uses the Generational Concurrent (`gencon`) GC policy, which is best suited for transactional applications that have many short-lived objects. The policy aims to minimize GC pause times without compromising throughput. If you are using Java 8, the `gencon` policy is similar to the `ParallelGC` policy, which is the default HotSpot policy. If you are using Java 11, the OpenJ9 balanced (`balanced`) policy is most similar to the default HotSpot policy.
+Eclipse OpenJ9 has a number of garbage colletion (GC) policies designed around different types of applications and workloads. By default, OpenJ9 uses the Generational Concurrent (`gencon`) GC policy, which is best suited for transactional applications that have many short-lived objects. The policy aims to minimize GC pause times without compromising throughput. If you are using Java 8, the `gencon` policy is similar to the `ParallelGC` policy, which is the default HotSpot policy. If you are using Java 11, the OpenJ9 balanced (`balanced`) policy is most similar to the default HotSpot policy.
```
amicic, line 39: I believe gencon is more similar to HotSpot CMS (concurrent mark sweep), although it does have have similarity with ParallelGC too, due to also being parallel.

docs/gc_overview.md
```diff
@@ -23,81 +23,183 @@
...
+
+Weak reference processing includes *soft* references, *weak* references, and *phantom* references. These references are created by the user for specific use cases and allow
+some level of interaction with the garbage collector. For example, a *soft* reference to an object allows that object to persist in memory for a longer period of time before being
+cleared. For example, a software cache. The garbage collector handles the reference types in the order shown and with the behavior detailed in the following table:
+
+
+| Reference type | Class                            | Garbage collector behavior |
+|----------------|----------------------------------|----------------------------|
+| *soft*         | `java.lang.ref.SoftReference`    | A *soft* reference is cleared only when its *referent* has not been marked for a number of GC cycles or if space on the heap is likely to cause an out of memory error. Use the [-Xsoftrefthreshold](xsoftrefthreshold.md) option to control the collection of *soft* reference objects. |
+| *weak*         | `java.lang.ref.WeakReference`    | A *weak* reference is cleared as soon as the object it references is removed. |
```
amicic, line 199: Try to align the terminology....

In Soft you use 'its referent' and weak you use 'the object it references' - indeed same thing, but maybe not obvious to every one.
In Soft you say 'has not been marked' and in Weak you say 'removed' The former is more correct for both. We don't really remove it, although we'll likely recycle its space by the end of this cycle/ at the moment of reference processing, we just know that it has not been marked as live (or more specifically not copy-forwarded or not scavenged for evacuating style of GCs).

In s

Note also that weak refs behave almost like Soft with threshold 0.

SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

Issue #709, opened 2021-01-12
Broken links in footer


keithc-ca:
**Page containing the error:**
  e.g. *www.eclipse.org/openj9/docs/*

**Describe the error and the update needed:**
The links under 'Docs' in the footer are broken.

PR #710, opened 2021-01-12
Fix broken links in footer

keithc-ca:
Fixes: #709.

### DIFF
--- a/theme/partials/footer.html
+++ b/theme/partials/footer.html
@@ -34,17 +34,17 @@
       <div class="above-line-links">
         <div class="column">
           <h3 >Docs</h3>
-          <span class="footer-link"><a href="/openj9_newuser/" >New to OpenJ9?</a></span>
-          <span class="footer-link"><a href="/introduction/">Getting started</a></span>
-          <span class="footer-link"><a href="/cmdline_specifying/">Command-line options</a></span>
+          <span class="footer-link"><a href="openj9_newuser/" >New to OpenJ9?</a></span>
+          <span class="footer-link"><a href="introduction/">Getting started</a></span>
+          <span class="footer-link"><a href="cmdline_specifying/">Command-line options</a></span>
         </div>
-  
+
         <div class="column">
           <h3>Resources</h3>
           <span class="footer-link"><a href="https://www.eclipse.org/openj9/about/" rel="noopener noreferrer" target="_blank">About</a></span>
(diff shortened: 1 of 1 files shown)


SueChaplain:
Jenkins doc stage

theme/partials/footer.html
```diff
@@ -34,17 +34,17 @@
       <div class="above-line-links">
         <div class="column">
           <h3 >Docs</h3>
-          <span class="footer-link"><a href="/openj9_newuser/" >New to OpenJ9?</a></span>
-          <span class="footer-link"><a href="/introduction/">Getting started</a></span>
-          <span class="footer-link"><a href="/cmdline_specifying/">Command-line options</a></span>
+          <span class="footer-link"><a href="openj9_newuser/" >New to OpenJ9?</a></span>
```
SueChaplain, line 37: Good spot @keithc but I don't think this is going to fix it. In production the link is broken in this way: `https://www.eclipse.org/openj9_newuser/` ... so we'd have to add `href="/openj9/docs/openj9_newuser/"`. However, this won't work on the draft site because of the way that GitHub pages works. 

There's probably another way to fix this. Looking ....

keithc-ca: An example link would be in https://www.eclipse.org/openj9/docs/index.html, so I would expect the relative href to refer to https://www.eclipse.org/openj9/docs/openj9_newuser/.

keithc-ca: The staging site looks good.

SueChaplain: I think the underlying problem is that we are building this for one site (ghpages) and then pushing the built site to an Eclipse web server (via another git repo). I think that we should probably hard code the links for the Eclipse site and accept that these do not resolve locally in the draft documentation.

If you agree, please could you update your commit to use the full URLs:

`<a href="https://www.eclipse.org/openj9/<docs_page>/"`

keithc-ca: I would expect relative links to continue to work on the Eclipse web server.
Can we try merging this to see? If it doesn't work, then I'll open a new pull request to use full URLs (which we expect will be broken on the staging site).

SueChaplain: Actually maybe it will. The `/` might be forcing it to go back to resolve from root (which is why it works on staging). Merging.

SueChaplain: Looks good - thanks @keithc!

PR #713, opened 2021-01-18
Correct introduction to exploiting GPUs

### LINKED ISSUE #712: GPU code page conversion
**Page containing the error:**

*https://www.eclipse.org/openj9/docs/introduction/*

**Describe the error and the update needed:**

The text says 'take advantage of graphics processing unit ... such as code page conversion', but code page conversion never became a reality (it didn't show sufficient benefit).

keithc-ca:
* code page conversion was never a supported use-case; mention array sorting instead
* GPU use is also supported on Linux aarch64; in fact they're supported on all little-endian architectures

Fixes #712.


pmhayward:
Jenkins doc stage

PR #719, opened 2021-01-25
Use absolute paths in footer links

keithc-ca:
The footer is included in multiple pages at differing depths in the tree, so absolute paths are required.

### DIFF
--- a/theme/partials/footer.html
+++ b/theme/partials/footer.html
@@ -34,9 +34,9 @@
       <div class="above-line-links">
         <div class="column">
           <h3 >Docs</h3>
-          <span class="footer-link"><a href="../openj9_newuser/" >New to OpenJ9?</a></span>
-          <span class="footer-link"><a href="../introduction/">Getting started</a></span>
-          <span class="footer-link"><a href="../cmdline_specifying/">Command-line options</a></span>
+          <span class="footer-link"><a href="/openj9/docs/openj9_newuser/" >New to OpenJ9?</a></span>
+          <span class="footer-link"><a href="/openj9/docs/introduction/">Getting started</a></span>
+          <span class="footer-link"><a href="/openj9/docs/cmdline_specifying/">Command-line options</a></span>
         </div>
 
         <div class="column">


doveye:
The links work fine for me as they are?

keithc-ca:
I went to https://www.eclipse.org/openj9/docs/ earlier today (and again just now).
Each of the links at the bottom left of the page ('New to OpenJ9?', 'Getting started' and 'Command-line options') is broken.
For example, the first should go to https://www.eclipse.org/openj9/docs/openj9_newuser/ instead of https://www.eclipse.org/openj9/openj9_newuser/.

keithc-ca:
BTW, same results with brave, chrome, edge and firefox.

doveye:
Ah, I didn't realise you were looking at the live site. I was looking at the draft one, although I have now checked the links on the live site and they also work fine for me. What do you see for the openj9_newuser link if you use the Chrome "Inspect" function on it? I see <a href="../openj9_newuser/">New to OpenJ9?</a>, which is what I'd expect (ie my browser hasn't cached an old page).

keithc-ca:
I see the same:
```
  <h3 >Docs</h3>
  <span class="footer-link"><a href="../openj9_newuser/" >New to OpenJ9?</a></span>
  <span class="footer-link"><a href="../introduction/">Getting started</a></span>
  <span class="footer-link"><a href="../cmdline_specifying/">Command-line options</a></span>
  </div>
```
But the problem is that those links are relative to https://www.eclipse.org/openj9/docs/ which loses the `docs` part of the path.

doveye:
I should have paid more attention to the link you started off with - https://www.eclipse.org/openj9/docs/. There, I *do* see the same behaviour as you. The problem is that that root topic is in a different place in the folder structure from all the others - it's one level above. All the other topics are in subfolders according to the filename of the original markdown file, which gets renamed by the build to index.html. So for example, docs/jit.md gets converted to docs/jit/index.html. That's the way MkDocs works, so I'm not sure we can change it (well, there was another link option that MkDocs introduced later, but changing to it means changing all the links and breaking people's bookmarks and I'm not sure it would fix this issue anyway).
So, looks like what's needed is a different footer just for that one topic. I'll defer to @pmhayward as he knows more about MkDocs than me :).

SueChaplain:
I don't see an easy way of fixing it except to hard code the links in the footer and accept that they will link to the production site from our draft gh-pages doc set. As it stands, the only page that has broken links is the landing page https://www.eclipse.org/openj9/docs/, which isn't great, but it would be a lot worse if it was every other page in the docs.

keithc-ca:
Ok, since the footer is included in multiple pages at differing depths in the tree, I think we have to use absolute paths: I'll update this accordingly.

SueChaplain:
Jenkins doc stage

Issue #726, opened 2021-02-01
List features in OpenJDK 16 that apply to OpenJ9


SueChaplain:
**Subject:**
http://openjdk.java.net/projects/jdk/16/ lists the features targeted for this release.

**Suggestion:**
List the ones that would apply to an OpenJDK that is built with OpenJ9 in the release topic. Call out important change https://openjdk.java.net/jeps/396, which might affect users apps.

pshipton:
The following jdk16 JEPs apply to OpenJ9.

338: Vector API (Incubator) - the OpenJDK impl works as-is, but OpenJ9 is adding optimization. TBD what makes it into 0.25

390: Warnings for Value-Based Classes
395: Records
397: Sealed Classes (Second Preview)

The following apply to OpenJ9 but are deferred to a future OpenJ9 release.

389: Foreign Linker API (Incubator)
393: Foreign-Memory Access API (Third Incubator)

The following don't apply directly to OpenJ9 but are part of OpenJDK so included in OpenJ9 builds.

347: Enable C++14 Language Features
380: Unix-Domain Socket Channels
394: Pattern Matching for instanceof
396: Strongly Encapsulate JDK Internals by Default

392: Packaging Tool - see https://github.com/eclipse/openj9/issues/11028 and https://github.com/eclipse/openj9-docs/issues/664

The following don't apply.

357: Migrate from Mercurial to Git
369: Migrate to GitHub
376: ZGC: Concurrent Thread-Stack Processing
386: Alpine Linux Port
387: Elastic Metaspace
388: Windows/AArch64 Port

pshipton:
@tajila fyi

tajila:
https://github.com/eclipse/openj9-docs/issues/726#issuecomment-770847916 looks good

PR #729, opened 2021-02-02
Add list of JDK 16 features that apply to OpenJ9

SueChaplain:
Explain which features apply/don't apply and whether they are implemented yet.
Add note about JEP 396, which might break user apps.

Also add note about general support for Java 8 & 11. No testing in this release and remove Windows compiler update that was copied through from 0.24.

Closes: #726

### DIFF
--- a/docs/version0.25.md
+++ b/docs/version0.25.md
@@ -27,6 +27,7 @@
 The following new features and notable changes since v 0.24.0 are included in this release:
 
 - [New binaries and changes to supported environments](#binaries-and-supported-environments)
+- ![Start of content that applies to Java 16](cr/java16.png) [New JDK 16 features](#new-jdk-16-features)
 - ![Start of content that applies to Java 11 (LTS) and later](cr/java11plus.png) [Support for the `-verbose:module` option](#support-for-the-verbosemodule-option)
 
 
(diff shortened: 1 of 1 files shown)


SueChaplain:
Jenkins doc stage

SueChaplain:
@pshipton - staged at https://eclipse.github.io/openj9-docs-staging/729/version0.25/#new-jdk-16-features
Thanks!

docs/version0.25.md
```diff
@@ -36,10 +37,46 @@ The following new features and notable changes since v 0.24.0 are included in th
...
+Although it might be possible to build an OpenJDK 8 or OpenJDK 11 with OpenJ9 release 0.25.0, testing at the project is not complete and therefore support for new features that apply to these Java versions is not available.
 
 To learn more about support for OpenJ9 releases, including OpenJDK levels and platform support, see [Supported environments](openj9_support.md).
 
+### ![Start of content that applies to Java 16](cr/java16.png) New JDK 16 features
+
+The following features are implemented in OpenJ9:
+
+- [JEP 338](https://openjdk.java.net/jeps/338): Vector API (incubator)
+    - The OpenJ9 project is extending this feature with optimizations.
```
pshipton, line 49: OpenJ9 isn't extending the feature. I expect Hotspot has optimizations that make this fast. OpenJ9 needs the same.

docs/version0.25.md
```diff
@@ -36,10 +37,46 @@ The following new features and notable changes since v 0.24.0 are included in th
...
+- [JEP 390](https://openjdk.java.net/jeps/390): Warnings for value-based classes
+- [JEP 395](https://openjdk.java.net/jeps/395): Records
+- [JEP 397](https://openjdk.java.net/jeps/397): Sealed Classes (second preview)
+
+The following features apply to OpenJ9 but are deferred to a future release:
+
+- [JEP 389](https://openjdk.java.net/jeps/389): Foreign linker API (incubator)
+- [JEP 393](https://openjdk.java.net/jeps/393): Foreign-memory access API (third incubator)
+
+The following features do not apply directly to OpenJ9 but are available in any builds of OpenJDK 16 with OpenJ9:
```
pshipton, line 59: Even though I wrote "do not apply directly" I don't like this wording, it can likely be improved. I wanted to say that OpenJDK implements these, their operation doesn't require any changes/support in OpenJ9. This probably isn't important to the end user, perhaps they should be grouped with the "features implemented by OpenJ9" which could be re-worded to "supported by OpenJ9".

docs/version0.25.md
```diff
@@ -36,10 +37,46 @@ The following new features and notable changes since v 0.24.0 are included in th
...
+- [JEP 347](https://openjdk.java.net/jeps/347): Enable C++ 14 language features
+- [JEP 380](https://openjdk.java.net/jeps/380): Unix-domain socket channels
+- [JEP 394](https://openjdk.java.net/jeps/394): Pattern matching for `instanceof`
+- [JEP 396](https://openjdk.java.net/jeps/396): Strongly encapsulate JDK internals by default
+- [JEP 392](https://openjdk.java.net/jeps/392): Packaging tool
+
+You can find the full list of features for JDK 16 at the [OpenJDK project](http://openjdk.java.net/projects/jdk/16/). Any remaining features that are listed by the OpenJDK project do not apply to OpenJ9.
+
+<i class="fa fa-pencil-square-o" aria-hidden="true"></i> **Note:** Applications might be adversely affected by
+JEP 396 if they make use of internal APIs other than `sun.misc.Unsafe`. You should update your application to use standard APIs. To temporarily work around this problem, set `--illegal-access=permit` on the command line, which prints a warning that is similar to the following example when an illegal access call is made:
```
pshipton, line 70: It's more than just sun.misc.Unsafe. Classes in packages exported from modules are still accessible. Related to Unsafe there are the jdk.unsupported and jdk.unsupported.desktop modules, which are expected to be removed in some future version of Java.

jdk.unsupported exports:
- com.sun.nio.file (ExtendedCopyOption, ExtendedOpenOption, ExtendedWatchEventModifier, SensitivityWatchEventModifier)
- sun.misc (Signal, SignalHandler, Unsafe)
- sun.reflect (ReflectionFactory)

jdk.unsupported.desktop exports:
- jdk.swing.interop (6 classes)

SueChaplain: I'm not sure if we really want to expand all this here? Could we just say the following and leave users to read the spec changes themselves?

Applications might be adversely affected by JEP 396 if they make use of internal APIs ~~other than `sun.misc.Unsafe`~~. You should update your application to use standard APIs. To temporarily work around this problem, set `--illegal-access=permit` on the command line, which prints a warning that is similar to the following example when an illegal access call is made:

pshipton: Sounds good.

pshipton (approved):

PR #735, opened 2021-02-17
Add content for configuring systems

SueChaplain:
- environment variables
- system resource limits
- large page memory allocation
- DLPAR support

Add missing content following contribution


SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

AdamBrousseau:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

docs/configuring.md
```diff
@@ -0,0 +1,180 @@
...
+* [1] https://www.gnu.org/software/classpath/license.html
+* [2] http://openjdk.java.net/legal/assembly-exception.html
+*
+* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH
+* Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+-->
+
+# Configuring your system
+
+When you install a Java&trade; runtime environment on your system you must set the **PATH** environment variable so that the operating system can find the Java programs and utilities to run your application. If your application requires a different library or class file that is not on the **PATH**, you must also set the **CLASSPATH** environment variable. How to set these environment variables is covered in many publications about Java, such as [The Java Tutorials: PATH and CLASSPATH](https://docs.oracle.com/javase/tutorial/essential/environment/paths.html).
```
pshipton, line 27: I'm objecting to "must". You don't have to set PATH if apps use the full path or start via a script that uses the JAVA_HOME environment variable or other mechanisms, but you can if you want. 

Setting CLASSPATH is also optional and the side affects should be understood before using it. If set globally it affects all invocations of Java, which may not be desirable.

docs/configuring.md
```diff
@@ -0,0 +1,180 @@
...
+
+On some systems, a further environment variable is required if your application requires shared libraries and these libraries are located in a different directory to the one that is specified in the application's header section. Set the following environment variables to specify the directory location of the shared libraries:
```
pshipton, line 29: I think this is only actually required on z/OS, and maybe AIX. Other platforms can load libraries when a full path is used. Only when the app is loading libraries without specifying their location do these variables need to be set. Just like CLASSPATH, setting these globally has side affects. Apps should avoid needing these to be set, or have a launcher that sets them just for the process.

docs/configuring.md
```diff
@@ -0,0 +1,180 @@
...
+- **LD_LIBRARY_PATH** (Linux&reg;)
+- **DYLD_LIBRARY_PATH** (macOS&reg;)  
+
+![Start of content that applies to Java 14+](cr/java14plus.png) On AIX systems, you must have the [XL C++ Runtime](https://www.ibm.com/support/pages/xl-cc-runtime-aix-v16101-fix-pack-december-2018) installed.
+
+## Setting resource limits (AIX, Linux, and macOS)
+
+The operating system sets resource limits per user for running processes to ensure that a single process cannot consume all available resources. However, these limits can affect certain operations that might need to run for a Java application, such as producing a dump file.
+
+Limits are controlled by the **ulimit** command and are referred to as a *hard* limit when set by the **ROOT** user, and a *soft* limit when set by other users. To see the *soft* limits set for a system, run `ulimit -a`. The output is similar to the following example:
```
pshipton, line 41: This information is incorrect. Copying from some man page "An unprivileged process may only set its soft limit to a value in the range from 0 up to the hard limit, and (irreversibly) lower its hard limit. A privileged process can make arbitrary changes to either limit value.".

docs/configuring.md
```diff
@@ -0,0 +1,180 @@
...
+|**MAXCPUTIME**	    | 2147483647 |
+|**MAXMMAPAREA** 	  | 40960      |
+|**IPCSEMNIDS**	    | 500        |
+|**IPCSEMNSEMS**	  | 1000       |
+
+
+<i class="fa fa-pencil-square-o" aria-hidden="true"></i> **Notes:**
+
+1. The number of threads that can be created by a Java process is limited by the lower of the two values for **MAXTHREADS** and **MAXTHREADSTASKS**.
+2. The number of 4K pages that are available for memory-mapped `*.jar` files is limited by **MAXMAPAREA** through the **JAVA_MMAP_MAXSIZE** environment variable.
```
pshipton, line 91: I don't find any reference to JAVA_MMAP_MAXSIZE in OpenJ9/OMR or OpenJDK code. I expect this is an IBM Java 8 thing.

docs/configuring.md
```diff
@@ -0,0 +1,180 @@
...
+HugePages_Free:     
+Hugepagesize:     
+```
+
+If these lines do not exist, update your Linux kernel. If `HugePages_Total` has a value of `0`, large pages are available, but not enabled.
+
+Where large page support is available, the following default sizes apply for the object heap:
+
+- Linux on x86: 2 MB large pages
+- Linux on IBM POWER: ??
```
pshipton, line 131: 16MB

docs/configuring.md
```diff
@@ -0,0 +1,180 @@
...
+
+To configure large page memory allocation on Linux, the kernel must support large pages. If large pages are supported, the following lines are present in the **proc/meminfo** file:
+
+```
+HugePages_Total:    
+HugePages_Free:     
+Hugepagesize:     
+```
+
+If these lines do not exist, update your Linux kernel. If `HugePages_Total` has a value of `0`, large pages are available, but not enabled.
```
pshipton, line 126: I expect all OSes supported today have huge page support.

SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

SueChaplain:
@pshipton. Joran has agreed to review the z/OS-specific content. I've made updates based on your earlier comments and now added more content for large pages. Please can you review or assign developers to review. I think we have what we need based on earlier docs, but let me know if anything is missing.

docs/configuring.md
```diff
@@ -0,0 +1,241 @@
...
+For more information, see [LFAREA](https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.4.0/com.ibm.zos.v2r4.ieae200/lfareaxx.htm) in the z/OS product documentation.
+
+For usage information, including examples, see [`-Xlp:objectheap`](xlpobjectheap.md#pageablenonpageable).
+
+
+## Configuring Dynamic LPAR support (AIX only)
+
+*Dynamic logical partitioning* (DLPAR) provides a mechanism to add or remove system resources, such as memory or CPU, to or from the operating system in a logical partition without rebooting. Changing these resources dynamically can have an impact on Java applications that are running on the LPAR.
+
+To enable an application to respond to DLPAR events, you can use OpenJ9 extensions to the `java.lang.management` API. These extensions can listen for events and trigger any necessary adjustments to the runtime environment. For example, if a Java VM is running in an LPAR with 2GB of memory, but the available memory might be adjusted between 1GB and 8GB, you might set the following options for the Java heap at run time:
```
pshipton, line 229: This should refer to the actual extensions in the API docs.
AvailableProcessorsNotificationInfo
ProcessingCapacityNotificationInfo
TotalPhysicalMemoryNotificationInfo

docs/configuring.md
```diff
@@ -0,0 +1,241 @@
...
+
+```
+–Xms1g –Xsoftmx2g –Xmx8g
+```  
+
+This command-line string sets an initial heap size of 1 GB, a *soft* (adjustable) maximum heap size of 2 GB, and a maximum heap size of 8 GB. You can then use the API to adjust the size of [`-Xsoftmx`](xsoftmx.md) dynamically in response to changes in memory resources.
```
pshipton, line 235: softmx support is unrelated to AIX DLPAR support, and is supported across platforms. For softmx, pls refer to MemoryMXBean getMaxHeapSize(), isSetMaxHeapSizeSupported(), setMaxHeapSize().

pshipton:
@zl-wang can you please take a look at the AIX/plinux information and see if there is anything that needs to be changed.

pshipton:
@vijaysun-omr do you have any advice about using THP vs huge pages on Linux. i.e. is there any advantage configuring for huge pages using -Xlp if THP is used?

pshipton:
@dmitripivkine fyi

docs/configuring.md
```diff
@@ -0,0 +1,241 @@
...
+```
+LDR_CNTRL=TEXTPSIZE=4K@STACKPSIZE=64K@DATAPSIZE=64K
+```
+
+The OpenJ9 VM supports large page usage only to back shared memory segments. Large pages are allocated by using `shmget()` with the **SHM_LGPG** and **SHM_PIN** flags.
+
+For more information, including support considerations, see [Large pages](https://www.ibm.com/support/knowledgecenter/ssw_aix_72/performance/large_page_ovw.html) and [Multiple page size support](https://www.ibm.com/support/knowledgecenter/ssw_aix_72/performance/multiple_page_size_support.html) in the AIX documentation.
+
+The 16 MB and 16 GB page sizes, which are intended for very high performance environments, require special user permissions. You must also configure the number of pages that you require, which cannot be adjusted on demand. For 16 MB large pages, you set the number of large pages by using the `vmo` command. For 64 GB huge pages you must define the number of pages by using the hardware management console. For more information, see [Page sizes for very high-performance environments](https://www.ibm.com/support/knowledgecenter/ssw_aix_72/performance/page_sizes_very_high-perf_env.html) in the AIX documentation.
+
```
zl-wang, line 156: typo:  64GB huge pages ... it should be 16GB huge pages

docs/configuring.md
```diff
@@ -0,0 +1,241 @@
...
+- **STACKPSIZE**: Page size to use for stacks
+- **DATAPSIZE**: Page size to use for native data or HEAP64
+
+The following example sets 4 KB for text and 64 KB for stack, native data, and heap areas:
+
+```
+LDR_CNTRL=TEXTPSIZE=4K@STACKPSIZE=64K@DATAPSIZE=64K
+```
+
+The OpenJ9 VM supports large page usage only to back shared memory segments. Large pages are allocated by using `shmget()` with the **SHM_LGPG** and **SHM_PIN** flags.
```
zl-wang, line 151: Not sure if we need to talk about these implementation details here. 
Plus, I didn't see SHM_LGPG but SHM_LGPAGE.

docs/configuring.md
```diff
@@ -0,0 +1,241 @@
...
+- **SHMALL**: The total amount of shared memory in the system (bytes or pages).
+
+The user running the Java process must either be ROOT or have permissions to use huge pages. For the appropriate permissions, the user must be a member of a group that has its group identifier (gid) stored in **/proc/sys/vm/hugetlb_shm_group**. The locked memory limit must also be increased to at least the size of the Java heap by using the **ulimit -l** command.
+
+To promote all memory allocated to huge pages, set the `-XX:+TransparentHugePage` option on the command line when you start your application and ensure that the Transparent Huge Page option on your system is set to `madvise`.
+
+Where huge page support is available, the following default sizes apply for the object heap:
+
+- Linux on x86: 2 MB
+- Linux on IBM POWER: 16 MB
```
zl-wang, line 187: starting POWER9 and later, there are two types of huge pages on Linux:  16MB and 2MB.  They are available exclusively (i.e. 16MB page availability doesn't co-exist with 2MB page availability). As currently, PowerVM LPAR(s)/machines will offer 16MB huge pages, while bare-metal machines will offer 2MB huge pages.  In the future, PowerVM environment may support 2MB huge pages (but not now).

docs/configuring.md
zl-wang: the question is whether JVM is notified of memory resource changes without DLPAR enabled.  i guessed JVM can keep polling the OS to query the available memory resources.

dmitripivkine: I believe softmx should not be mentioned in this context as an option for dynamic object heap size control. And yes, I agree it should be referred to MemoryMXBean

docs/configuring.md
```diff
@@ -0,0 +1,241 @@
...
+* [1] https://www.gnu.org/software/classpath/license.html
+* [2] http://openjdk.java.net/legal/assembly-exception.html
+*
+* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH
+* Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+-->
+
+# Configuring your system
+
+When you install a Java&trade; runtime environment on your system you can set the **PATH** environment variable so that the operating system can find the Java programs and utilities to run your application. If your application requires a different library or class file that is not on the **PATH**, you can also set the **CLASSPATH** environment variable. However, if you set **CLASSPATH** globally, all invocations of Java are affected. How to set these environment variables is covered in many publications about Java, such as [The Java Tutorials: PATH and CLASSPATH](https://docs.oracle.com/javase/tutorial/essential/environment/paths.html).
```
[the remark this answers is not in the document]
SueChaplain: This time I read the tutorial ;-) ... 

*To tell your application where to find user classes, you can use the **-cp** option or set the **CLASSPATH** environment variable.*

docs/configuring.md
```diff
@@ -0,0 +1,241 @@
...
+Where `<number>` is the number of huge pages required.
+
+Configure the number of huge pages that you require at boot time to ensure that the VM has access to sufficient contiguous pages. The following kernel parameters must be set appropriately for your system:
+
+- **SHMMAX**: The maximum size of the shared memory segment (bytes).
+- **SHMALL**: The total amount of shared memory in the system (bytes or pages).
+
+The user running the Java process must either be ROOT or have permissions to use huge pages. For the appropriate permissions, the user must be a member of a group that has its group identifier (gid) stored in **/proc/sys/vm/hugetlb_shm_group**. The locked memory limit must also be increased to at least the size of the Java heap by using the **ulimit -l** command.
+
+To promote all memory allocated to huge pages, set the `-XX:+TransparentHugePage` option on the command line when you start your application and ensure that the Transparent Huge Page option on your system is set to `madvise`.
```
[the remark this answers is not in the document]
SueChaplain: I got rid of the current paragraph and created a new one at the end of the Linux section:

*Transparent HugePage Support (THP) is an automated mechanism of using huge pages to back virtual memory. On Linux kernels that support THP, it is typically enabled by default with the **madvise** option and can be relied on to provide huge pages as required without any user configuration. To disable THP for your application, use the OpenJ9 [`-XX:-TransparentHugePage`](xxtransparenthugepage.md) option on the command line. To disable THP system-wide, change the **sysfs** boot time defaults with the command `transparent_hugepage=never `. For more information about THP see [Transparent HugePage Support](https://www.kernel.org/doc/html/latest/admin-guide/mm/transhuge.html).*

Partly my interpretation of reading the Linux doc, so please put me straight!

docs/configuring.md
```diff
@@ -0,0 +1,241 @@
...
+- See [z/OS Language Environment Programming Reference](https://www.ibm.com/support/knowledgecenter/SSLTBW_2.4.0/com.ibm.zos.v2r4.ceea300/abstract.htm) for a full list of the available runtime options.
+- See [z/OS Language Environment Debugging Guide](https://www.ibm.com/support/knowledgecenter/SSLTBW_2.4.0/com.ibm.zos.v2r4.ceea100/abstract.htm) for tuning guidance by using **RPTSTG (ON)**.
+
+<i class="fa fa-exclamation-triangle" aria-hidden="true"></i> **Warning:** Changing the runtime options can often degrade performance.
+
+## Configuring large page memory allocation
+
+If your application allocates a large amount of memory and frequently accesses that memory, you might be able to improve performance by enabling large page support on your system.
+
+Large page support is enabled by setting the [`-Xlp:objectheap`](xlpobjectheap.md) and [`-Xlp:codecache`](xlpcodecache.md) options on the command line when you start your application. These options have the following effects:
```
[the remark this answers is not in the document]
SueChaplain: Updated:

*Some Linux kernels implement Transparent HugePage Support (THP), which automates the provision of large pages to back virtual memory, as described in [Linux systems](#linux-systems). Alternatively, you can enable large page support by setting the [`-Xlp:objectheap`](xlpobjectheap.md) and [`-Xlp:codecache`](xlpcodecache.md) options.....*

docs/configuring.md
SueChaplain: Thanks .... clearer. I've modified:

>You can then use the `MemoryMXBean` API to dynamically respond to changes in memory resources. The following classes can be used:

- `getMaxHeapSize()`: Query the maximum heap size.
- `isSetMaxHeapSizeSupported()`: Query whether the VM can support dynamic updates.
- `setMaxHeapSize()`: Adjust the maximum heap size.

>For more information about the `com.ibm.lang.managment` package, which extends the `jdk.management` module, see the [API documentation](api-overview.md#monitoring-and-management).

docs/configuring.md
SueChaplain: Removed the sentence. We used to talk about this in the J9 docs, but I did wonder why.

SueChaplain:
Jenkins doc stage

docs/configuring.md
pshipton: @zl-wang the doc is updated to state "bare-metal machines support only 2 MB huge pages", is "only" correct?

docs/configuring.md
pshipton: The softmx doc is still in the wrong place. It doesn't belong
a) in Configuring your system, since it's a JVM feature unrelated to the OS
b) under Configuring Dynamic LPAR support (AIX only), it's not an AIX specific feature and it's not related to DLPAR

docs/configuring.md
zl-wang: it is correct as of after-a-certain-version of kernel.  such that, as a forward-looking document, it is correct.  if this document meant to describe older kernel situations as well, i would suggest changing the wording on bare-metal case to be "either 16MB or 2MB huge pages, depending on the exact version of the Linux kernel".  Too mouthful maybe. of course, a better more concise description (to the same meaning) would be very welcome.

docs/configuring.md
SueChaplain: It's explaining a scenario where you might want to change the settings dynamically. i.e. For example, ....... you've created an environment with limits x,y,z ... and then when notified that things have changed, you can react. I accept it is not an OS feature or related to DLPAR. I'll remove if you think it confuses people .... or maybe I'm still misunderstanding your point.

docs/configuring.md
SueChaplain: So: 

- IBM POWER9 support 16 MB or 2 MB
- PowerVM LPARs support 16 MB
- Bare-metal systems support 16 MB or 2 MB depending on your Linux kernel version

If they can get the settings from `/proc/meminfo `, should we just get them to check that instead and say something like:

- IBM Power Systems: Varies depending on kernel version, check `/proc/meminfo`

docs/configuring.md
pshipton: Ok, I see the point.

vijaysun-omr:
@pshipton I have not really seen a benefit from using -Xlp/manually configured large pages when THP is configured on linux (x86 is where I have most experience to make that statement). 

I believe there is some difference in how the RSS is reported where the accounting happens more accurately with THP than with manually configured huge pages. Documented at: https://bugzilla.linux.ibm.com/show_bug.cgi?id=181776

docs/configuring.md
```diff
@@ -0,0 +1,254 @@
...
+
+Language Environment&reg; runtime options affect performance and storage usage. These options can be optimized for your application.
+
+Runtime options are typically embedded in programs by using **#pragma runopts** settings. In many cases, these options provide suitable default values that are known to produce good performance results. However, these options can be overridden to tune the runtime environment of your application.
+
+On 64-bit z/OS systems, the following runtime options affect Java applications:
+
+- **HEAP64**: Controls allocation of the user heap. A suggested starting point for an override is **HEAP64(512M,4M,KEEP,16M,4M,KEEP,0K,0K,FREE)**.
+- **HEAPPOOLS64**: Used to manage Java heap storage above the 2 G bar. The Java USS launcher sets **HEAPPOOLS64(ALIGN)** for more optimal management of multi-threaded applications. Other Java launchers might have different settings. Before you set an override for **HEAPPOOLS64**, use **RPTOPTS(ON)** or **RPTSTG(ON)** and check the result of **#pragma runopts**. Check this result because the host product might have already set cell sizes and numbers that are known to produce good performance.
```
joransiu, line 104: > Before you set an override for **HEAPPOOLS64**, use **RPTOPTS(ON)** or **RPTSTG(ON)** , and check the result of **#pragma runopts**.

For the following statement, I think we can clarify further by by explaining the use of the two report options.  Something like:

> Before you set an override for **HEAPPOOLS64**, use **RPTOPTS(ON)** to confirm the active settings for your environment, and **RPTSTG(ON)** to review storage usage and tuning recommendations.

docs/configuring.md
```diff
@@ -0,0 +1,254 @@
...
+- **STACK64**: Controls the allocation and management of stack storage. A suggested default is **STACK64(1M,1M,128M)**.
+- **THREADSTACK64**: Controls the allocation of thread-level stack storage for both the upward and downward-growing stack. A suggested default is **THREADSTACK64(OFF,1M,1M,128M)**.
+
+A suitable **MEMLIMIT** value is also required. The OpenJ9 VM requirement is the sum of the following amounts:
+
+- **HEAP64** (minimum 512 M), which should include **HEAPPOOLS64**.
```
joransiu, line 110: HEAP64 dictates an allocation policy and HEAPPOOLS64 controls the management of the storage.  I'm not too sure if the latter is meaningful towards MEMLIMIT.   I suggest we update to:

- User Heap Storage required by the JVM and native libraries, as controlled by **HEAP64** (minimum 512 M) settings.

docs/configuring.md
```diff
@@ -0,0 +1,254 @@
...
+- **STACK64** (3 MB multiplied by the expected number of concurrent threads)
```
joransiu, line 111: Similar to above:
- User Stack Storage (3MB multiplied by the expected number of concurrent threads), as controlled by **STACK64** settings.

docs/configuring.md
```diff
@@ -0,0 +1,254 @@
...
+- `-Xmx` largest expected VM heap size.
```
joransiu, line 112: @dmitripivkine : I recall we had to over-reserve in the past due to lack of alignment support for IARV64 (at least under CS.  Do you know if that's still the case?

joransiu (commented):
The BPXPRM* / OMVS minimum settings seem reasonable to me, but I will review with the POK performance team to validate. 

Left misc suggestions on some of the HEAP*/STACK* related wording.

docs/configuring.md
```diff
@@ -0,0 +1,254 @@
...
+- [`GetLargePageMinimum` function (`memoryapi.h`)](https://docs.microsoft.com/en-gb/windows/win32/api/memoryapi/nf-memoryapi-getlargepageminimum)
+
+### z/OS systems  
+
+When available, 1 MB pageable pages are the default size for the object heap and the code cache. Other page sizes are available for the object heap, depending on the system architecture as shown in the following table:
+
+| Large page size  | System architecture required              | `-Xlp:codecache` | `-Xlp:objectheap`          |
+|------------------|-------------------------------------------|------------------|----------------------------|
+| 2 GB nonpageable | IBM zEnterprise EC12 processor or later   | Not supported    | Supported (64-bit VM only) |
+| 1 MB nonpageable | System z10 processor or later             | Not supported    | Supported (64-bit VM only) |
```
joransiu, line 207: While accurate, I believe these docs will only cover 64-bit JVM only, so not sure if the `(64-bit VM only)` is necessary.

docs/configuring.md
```diff
@@ -0,0 +1,254 @@
...
+| 1 MB pageable    | IBM zEnterprise EC12 processor or later (see Note) |Supported| Supported                  |
+
+<i class="fa fa-pencil-square-o" aria-hidden="true"></i> **Note:** The Flash Express feature (#0402) helps avoids demoting 1 MB pageable pages to 4 KB pages during periods of high memory usage.
```
joransiu, line 210: I am not sure if Flash Express is still relevant nowadays.  Have sent an email to David Hom from RSM team to confirm.

SueChaplain: I've updated the content with the suggestions/corrections from Dave.

docs/configuring.md
SueChaplain: Yes, Peter suggested we don't bring in any 31-bit z/OS content. However, I've got some cleaning up to do as some cmdline options were already in place. E.g. -XCEEHDLR. I'll sweep the docs in a separate PR to ensure we're only indicating 64-bit for z/OS. We do have some 31-bit content for Linux on Z, so I'll have to be careful not to accidentally remove anything in the process.  :-/ - I'll involve you in the review there if that's ok with you?

SueChaplain:
Jenkins doc stage

docs/configuring.md
dmitripivkine: > @dmitripivkine : I recall we had to over-reserve in the past due to lack of alignment support for IARV64 (at least under CS. Do you know if that's still the case?

Yes, for Z platforms for Concurrent Scavenger running with HW Guarded Storage support for Full 64-bit JVM actual heap memory reservation might be larger than provided in `-Xmx` (over-reserved). It is necessary to guarantee proper alignment of Guarded Storage memory range.

joransiu: Thanks @dmitripivkine.  @SueChaplain, we had the following statement around this point in the ConcurrentScavenge docs.   Should we consider adding a reference to that?

> Note: On z/OS, the virtual storage used might exceed the Java maximum heap size. Set the z/OS memory limit, specified by ulimit -M, to a larger value than the maximum heap size.

https://www.ibm.com/support/knowledgecenter/en/SSYKE2_8.0.0/openj9/xgc/index.html#concurrentscavenge

SueChaplain: Thanks, adding:

**Note:** If you intend to use the Concurrent Scavenge mode of the default Generational Concurrent (`gencon`) garbage collection policy, the virtual storage used might exceed the Java maximum heap size. Set the z/OS memory limit to a larger value than the maximum heap size. For more information, see [`-Xgc:concurrentScavenge`](xgc.md#concurrentscavenge).

SueChaplain: Oh. Does that need clarifying as H/W not S/W enabled?

joransiu: I agree we should clarify this applies only under H/W usage, which would be default with concurrent scavenge on IBM z14 or newer.

SueChaplain:
Jenkins doc stage

docs/configuring.md
SueChaplain: @zl-wang - please see my last comment. Would you recommend getting the user to check `/proc/meminfo` in order to keep this content current going forward please? If not, is the list I printed ok?

SueChaplain:
Jenkins doc stage

SueChaplain:
@pshipton @joransiu - updates now added for shared classes cache resources.

- I believe the info for AIX/Linux/macOS is accurate based on other docs. 
- z/OS is as accurate as we know until further feedback is received. 

I'm keen to get this content into the OpenJ9 docs as the best available info for now and fix up later if we get further feedback. Please can you each verify that SHRC looks accurate from your perspectives.

Docs being staged here: https://eclipse.github.io/openj9-docs-staging/735/configuring/

SueChaplain:
Jenkins doc stage

SueChaplain:
P.S. The layout still troubles me. "Large page config" is really just another system resource that needs configuring, same as "shared memory" or "ulimits". Please feel free to suggest a restructure for what you think makes sense. We could just cover each of those config changes per platform.

pshipton:
In the "On some systems, a further environment variable might be required", pls add `PATH` for Windows.

pshipton:
"Getting shared memory settings accurate" is not required for the java heap. Java 11+ also uses shared memory to allocate the `lib/modules` file (containing all the modules .class files and resources) which is ~120MB+, although it will only share the `modules` file index when not enough shared memory is available.

pshipton:
I see, setting SHMMAX/SHMALL is mentioned in the Configuring large page memory allocation Linux section, so it's important when explicitly using large pages and not Transparent HugePage Support.

SueChaplain:
Jenkins doc stage

docs/configuring.md
```diff
@@ -0,0 +1,285 @@
...
+The OpenJ9 class data sharing feature is implemented by using shared memory segments on z/OS. Special consideration should be given to the following parameters that relate to the shared memory and IPC semaphore settings:
+
+<!-- - **MAXSHAREPAGES** (This no longer seems to do anything, per https://www.ibm.com/support/knowledgecenter/SSLTBW_2.4.0/com.ibm.zos.v2r4.ieae200/spb.htm). Also removed from table below. -->
+- **IPCSHMSPAGES**
+- **IPCSHMMPAGES**
+- **IPCSHMNSEGS**
+
+Incorrect or suboptimal settings might prevent shared classes from working or impact performance. By default, the VM attempts to create a 16 MB cache. If you set a cache size for your application by specifying the [`-Xscmx`](xscmx.md) option on the command line, the VM rounds the value up to the nearest megabyte. Ensure that the value set for **IPCSHMMPAGES** takes this adjustment into consideration.
+
+To see the current settings, enter the following z/OS operator command:
```
joransiu, line 93: I'd suggest this paragraph on how to see current OMVS settings, and the table below with suggested min values should be bumped up and placed right after the initial paragraph of this section.   i.e. placed ahead of the shared memory options above.

SueChaplain:
Jenkins doc stage

SueChaplain:
@pshipton - I've commented out the z/OS system resource information until we have confirmation of the settings from an SME. The remainder of this content has been reviewed, so if you are happy, please merge. Currently staged here: https://eclipse.github.io/openj9-docs-staging/735/configuring/

docs/configuring.md
pshipton: @SueChaplain I think you should go ahead with your suggested change until @zl-wang has further input.

SueChaplain:
Jenkins doc stage

SueChaplain:
@pshipton - update made to the Linux large page section as suggested.

pshipton (approved):

PR #742, opened 2021-03-03
-Xconcurrentbackground default value correction

RSalman:
Documentation for `Xconcurrentbackground` default behaviour is not up-to-date, the default value 1 is on all platforms. Obsolete limitation on zLinux was removed some time back - https://github.com/eclipse/omr/pull/3617.

Signed-off-by: Salman Rana <salman.rana@ibm.com>

### DIFF
--- a/docs/xconcurrentbackground.md
+++ b/docs/xconcurrentbackground.md
@@ -32,9 +32,11 @@ Specifies the number of low-priority background threads attached to assist the m
 
 ## Default behavior
 
-The default is `0` for Linux&reg; on IBM Z&reg; and `1` on all other platforms.
+The default value is `1`.
 
+<i class="fa fa-pencil-square-o" aria-hidden="true"></i> **Note:** This value is reported in the header section of a verbose GC log with an entry 
+
+        <attribute name="gcthreads Concurrent Mark" value="1" />
  
-
 <!-- ==== END OF TOPIC ==== xconcurrentbackground.md ==== -->
 


RSalman:
fyi @dmitripivkine @amicic

amicic:
I'm fine with removing traces that in past for zLinux for some period it was 0, but I wonder if we should add a generic statement that would help avoid confusions that the current value is reported by VGC initialized stanza

RSalman:
Added note to documentation: 
_**This value is reported in the initialized header of Verbose GC logs as "gcthreads Concurrent Mark" attribute.**_

_Will look something like this_ 
image

SueChaplain:
Jenkins doc stage

docs/xconcurrentbackground.md
```diff
@@ -30,9 +30,11 @@ Specifies the number of low-priority background threads attached to assist the m
 
         -Xconcurrentbackground<n>
 
+<i class="fa fa-pencil-square-o" aria-hidden="true"></i> **Note:** This value is reported in the initialized header of Verbose GC logs as "gcthreads Concurrent Mark" attribute.
```
SueChaplain, line 33: @RSalman - I was a little confused by the note until I looked at a log. Perhaps you can add the actual output. Suggest:

This value is reported in the header section of a verbose GC log with an entry `<attribute name="gcthreads Concurrent Mark" value="1" />`.

RSalman: Thanks - updated with the suggested text

SueChaplain:
Jenkins doc stage

PR #752, opened 2021-03-17
Improve class data sharing topic

SueChaplain:
- Add more content as an overview
- Add more content for Enabling a cache
- Add a class data sharing ops section with a subsection on AOT code and JIT data
- Repurpose the Benefits section to a Creating section (largely unchanged), add Creating a layered cache from shareclasses topic
- Add a Saving a cache section, with some content pulled in from shareclasses topic
- Add a Housekeeping topic
- Add Support for Custom class loaders, plus info about Java Helper API
- Removed the Cache utilities section (for now) as most of it had been covered in other sections.

Topic might need splitting, reorganising.
More content from xshareclasses should live here, with xshareclasses being more reference.

Signed-off-by: Sue-Chaplain <sue_chaplain@uk.ibm.com>


SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

SueChaplain:
@pshipton - I'm still fiddling with a couple of code blocks that won't behave themselves inside a list, but please could you start to take a look. There is much more to add in general, including BCI, but I thought this might be a good start. I'll do another wave post 0.26 GA.

docs/shrc.md
```diff
@@ -66,43 +140,138 @@ The [-Xshareclasses](xshareclasses.md) option is highly configurable, allowing y
...
+    It is good practice to explicitly set permissions for the cache directory when the defaults are not appropriate. Access is controlled by operating system permissions and Java security permissions; read/write access is the default and gives all users equal rights to update the cache. On Unix systems, you can use the [`-Xshareclasses:groupAccess`](xshareclasses.md#groupaccess) suboption to allow read/write permissions for groups as well as users.
+
+    On z/OS, a cache can be accessed only by a VM that is running in the same storage key as the VM that created the cache. If the keys do not match, permission to access the cache is denied.
 
 - Set the [`-Xshareclasses:nonfatal`](xshareclasses.md#nonfatal) option.
 
-    This option means that your application can start even if there is a problem opening or creating the shared cache, in which case, the VM might be able to start *without* class data sharing.
+    Setting this option allows your application to start even if there is a problem opening or creating the shared cache. In this situation, the VM might be able to start *without* class data sharing.
```
pshipton, line 151: We should note here or in the `nonfatal` doc that this option doesn't address all cases, just certain scenarios such as not having read access to an existing cache, or detectable corruption. i.e. it doesn't help with bugs or all cache corruption. Cache corruption (generally as a result of a bug in the OS, VM, user code) isn't necessarily detected when opening the cache, the cache will be used and the application may crash.

pshipton: I don't think this should say "might be able", but that the VM will continue to start without the shared cache.

docs/shrc.md
```diff
@@ -66,43 +140,138 @@ The [-Xshareclasses](xshareclasses.md) option is highly configurable, allowing y
...
+
+When the cache is active, you can adjust the soft maximum size by using the [`-Xshareclasses:adjustsoftmx`](xshareclasses.md#adjustsoftmx) cache utility.
```
pshipton, line 153: This should be another bullet point. It seems sparse, as soft max hasn't been introduced.

docs/shrc.md
```diff
@@ -66,43 +140,138 @@ The [-Xshareclasses](xshareclasses.md) option is highly configurable, allowing y
...
+
+Using the [`-Xshareclasses:listAllCaches`](xshareclasses.md#listallcaches) cache utility allows you to identify all the caches that exist on a system, including compatible and incompatible caches.
```
pshipton, line 155: This should be another bullet point. Also it doesn't find all caches on a system, only the ones in the default directory, or the ones in the directory specified by adding a `cacheDir` suboption.

docs/shrc.md
```diff
@@ -66,43 +140,138 @@ The [-Xshareclasses](xshareclasses.md) option is highly configurable, allowing y
...
+
+Normally, writing to an existing shared cache in a lower image layer results in Docker duplicating the shared cache to the top layer (following the Docker [copy-on-write strategy](https://docs.docker.com/storage/storagedriver/#the-copy-on-write-cow-strategy)). With a layered cache, you can instead write into a new cache in the top layer. The new cache builds on the existing cache, so space is saved in the image.
 
+The following example shows a Docker container with four layers:
+
+: ![This diagram is explained in the surrounding text](./cr/shrc_layers.jpg "Docker container with layered caches")
+
+
+1. The lowest layer is a Ubuntu Docker image.
+2. The next layer is an OpenJ9 Docker image that is built on the Ubuntu image. As part of this image, the `-Xshareclasses:name=Cache1` suboption is used to create a cache called `Cache1`. The layer number assigned to this cache is `0`. The `listAllCaches` suboption shows the cache and the layer number:
```
pshipton, line 169: We can use a different cache name for the example, such as `MyCache`, so the 1 in Cache1 isn't confused with the layer number.

docs/shrc.md
```diff
@@ -66,43 +140,138 @@ The [-Xshareclasses](xshareclasses.md) option is highly configurable, allowing y
...
+        ...
+        Cache name              level         cache-type      feature         layer       OS shmid       OS semid       last detach time
+
+        Compatible shared caches
+        Cache1                  Java8 64-bit  persistent      cr              0                                         Mon Sep 23 11:41:04 2019
+        ```
+
+3. The next Docker layer up is an Open Liberty image that is built on the OpenJ9 image. As part of this image, the `-Xshareclasses:name=Cache1,layer=1` suboption is used to create another cache called `Cache1`. Because the `layer=1` suboption is specified, this new cache is a layered cache, which builds on `Cache1` in the previous container layer. (Open Liberty starts two VMs, so if you instead use the `createLayer` suboption here, two layered caches are created, with layer numbers of 1 and 2.) Note that cache layers are different from, and independent of, container layers.
+
+4. In the same way, another Docker layer is added for an Open Liberty Java application, and another layered cache is created to add to `Cache1`. The `listAllCaches` suboption now shows all the caches and their layers:
```
pshipton, line 182: "Open Liberty" doesn't need to be mentioned here, "application" is sufficient.

docs/shrc.md
```diff
@@ -66,43 +140,138 @@ The [-Xshareclasses](xshareclasses.md) option is highly configurable, allowing y
...
+
+        Compatible shared caches
+        Cache1                  Java8 64-bit  persistent      cr              0                                         Mon Sep 23 11:41:04 2019   
+        Cache1                  Java8 64-bit  persistent      cr              1                                         Mon Sep 23 11:46:25 2019
+        Cache1                  Java8 64-bit  persistent      cr              2                                         In use
+        ```  
+
+The caches are created in the same directory.
+
+When you use the `-Xshareclasses:name=Cache1` suboption in future Java commands, all the caches are started. The top-layer cache is started in read/write mode, and lower-layer caches are started in read-only mode. Modifying a lower-layer cache would invalidate all the caches in the layers above.
```
pshipton, line 197: would -> will

docs/shrc.md
```diff
@@ -66,43 +140,138 @@ The [-Xshareclasses](xshareclasses.md) option is highly configurable, allowing y
...
+The following options and cache utilities are available for creating, managing, and removing layered caches:
+
+- [`-Xshareclasses:createLayer`](xshareclasses.md#createlayer)
+- [`-Xshareclasses:layer`](xshareclasses.md#layer)
+- [`-Xshareclasses:printTopLayerStats`](xshareclasses.md#printtoplayerstats) (for example output, see [printTopLayerStats](shrc_diag_util.md#printtoplayerstats))
+- [`-Xshareclasses:destroyAllLayers`](xshareclasses.md#destroyalllayers)
+
+## Saving a shared classes cache
+
+As described in an earlier section, a shared classes cache can be persistent or non-persistent. Non-persistent caches are stored in shared memory and are removed when a system is rebooted. If you want to save a cache beyond a reboot you might want to consider making your cache persistent or taking a cache snapshot.
```
pshipton, line 208: Persistent caches are memory mapped files.

> If you want to save a cache beyond a reboot you might want to consider making your cache persistent

Seems obsolete, persistent caches are the default except on z/OS where persistent isn't an option.

docs/shrc.md
```diff
@@ -66,43 +140,138 @@ The [-Xshareclasses](xshareclasses.md) option is highly configurable, allowing y
...
+
+To use a persistent cache, see the [`-Xshareclasses:persistent`](xshareclasses.md#persistent) suboption.
```
pshipton, line 210: As above, this is the default on platforms that support it.

docs/shrc.md
```diff
@@ -66,43 +140,138 @@ The [-Xshareclasses](xshareclasses.md) option is highly configurable, allowing y
...
+
+## Housekeeping
+
+Caches can be deleted if they contain many stale classes or if the cache is full and you want to create a bigger cache. Use one of the following utilities to remove unwanted caches:
+
+- [`-Xshareclasses:destroy`](xshareclasses.md#destroy): Removes specific caches when used with the `name`, `cacheDir`, and `nonpersistent` suboptions.
+- [`-Xshareclasses:destroyAll`](xshareclasses.md#destroyall): Removes all the caches that are specified by the `cacheDir` and `nonpersistent` suboptions.
+- [`-Xshareclasses:destroySnapshot`](xshareclasses.md#destroysnapshot): Removes a cache snapshot from disk that is specified by `name` and `cacheDir` suboptions.
+- [`-Xshareclasses:destroyAllSnapshots`](xshareclasses.md#destroyallsnapshots): Removes all cache snapshots from disk that are found by specifying the `cacheDir` suboption.
+- [`-Xshareclasses:destroyAllLayers`](xshareclasses.md#destroyalllayers): Removes all shared cache layers that are specified by the `name` suboption.
```
pshipton, line 233: `cacheDir` should be mentioned here as well.

docs/shrc.md
```diff
@@ -66,43 +140,138 @@ The [-Xshareclasses](xshareclasses.md) option is highly configurable, allowing y
...
+
+Caches can also be removed if they are unused for a specified amount of time. To configure time-based housekeeping, use the [-Xshareclasses:expire](xshareclasses.md#expire) option.
+
+If you want to remove a cache but allow it to be re-created when the VM restarts, use the [`-Xshareclasses:reset`](xshareclasses.md#reset) option.
 
 
 ## Support for custom class loaders
 
-Classes are shared by the bootstrap class loader internally in the OpenJ9 VM. If you are using a custom class loader, you can use the Java Helper API to find and store classes in the shared classes cache.
+Classes are shared by the bootstrap class loader internally in the VM. The OpenJ9 implementation of `java.net.URLClassLoader` is modified to use `SharedClassURLClasspathHelper` and any class loaders that extend `java.net.URLClassLoader` inherit this behavior. If you are using a custom class loader that does not extend `java.net.URLClassLoader`, you can use the Java Helper API to find and store classes in a shared classes cache.
```
pshipton, line 243: may inherit this behavior.

docs/shrc.md
```diff
@@ -66,43 +140,138 @@ The [-Xshareclasses](xshareclasses.md) option is highly configurable, allowing y
...
 
-For more information, see [Using the Java Helper API](https://www.ibm.com/support/knowledgecenter/SSYKE2_8.0.0/com.ibm.java.vm.80.doc/docs/shrc_pd_helper.html).
+If a running application uses its own class loaders, you must grant these class loaders shared class permissions before they can share classes. To grant permission, add shared class permissions to the `java.policy` file by specifying the `ClassLoader` class name. Permissions can be set for `read`, `write`, or `read,write`.
```
pshipton, line 245: As does the next point, this only applies if a `SecurityManager` is being used. `SharedCachePermisssion` is the name of the permission.

docs/shrc.md
```diff
@@ -66,43 +140,138 @@ The [-Xshareclasses](xshareclasses.md) option is highly configurable, allowing y
...
+
+The following main functions are available from the `SharedClassHelper` API:
+
+- `findSharedClass`: Used to check whether a class is already in the cache before looking for the class on the filesystem.
+- `storeSharedClass`: Used to store a class in the cache.
+- `setSharingFilter`: A filter that can be used to decide which classes are found and stored in the cache. This filter can be applied to a particular package. To apply a filter to all non-bootstrap class loaders that share classes, specify the `-Dcom.ibm.oti.shared.SharedClassGlobalFilterClass` system property on the command line.
+
+The following APIs are available for obtaining information about a shared classes cache:
```
pshipton, line 261: This is a separate sub-section than the `SharedClassHelper` stuff. I'm thinking there should be 2 headers, "The Java Shared Classes Helper API", and "The Java Shared Classes Utility API".

docs/shrc.md
```diff
@@ -66,43 +140,138 @@ The [-Xshareclasses](xshareclasses.md) option is highly configurable, allowing y
...
 
-- Adjust the minimum and maximum amount of cache space reserved for AOT or JIT data.
-- Adjust the soft maximum size of the cache.
-- Create snapshots of a cache.
-- Create a cache from a snapshot.
-- Remove caches and cache snapshots.
-- List all the compatible and incompatible caches and snapshots.
-- For problem determination, invalidate and revalidate AOT methods that cause a failure in an application.
-- For problem determination, provide information to analyze the contents of a shared classes cache.
+Each class loader that wants to share data must get a `SharedDataHelper` object from a `SharedDataHelperFactory`.
```
pshipton, line 267: This belongs with the `SharedClassHelper` stuff.

docs/shrc.md
```diff
@@ -66,43 +140,138 @@ The [-Xshareclasses](xshareclasses.md) option is highly configurable, allowing y
...
+A `SharedDataHelperFactory` provides an interface that can be used to create `SharedDataHelpers`, which are used for storing Java byte array data. A `SharedDataHelper` also has a one to one relationship with a class loader, although a class loader can exist without a `SharedDataHelper`.
 
-For more information, see [-Xshareclasses](xshareclasses.md).
 
 ## See also
 
 - [AOT compiler](aot.md)
-- [Class Sharing](https://www.ibm.com/developerworks/java/library/j-ibmjava4/index.html)
+- [Class sharing article](https://www.ibm.com/developerworks/java/library/j-ibmjava4/index.html)
```
pshipton, line 274: This link doesn't work any more. I think https://developer.ibm.com/languages/java/tutorials/j-class-sharing-openj9/ is the latest.

pshipton:
@hangshao0 @mpirvu fyi as well.

docs/shrc.md
```diff
@@ -22,33 +22,107 @@
...
-Memory footprint is reduced by sharing common classes between applications that run in separate Java VMs.
+When you enable class data sharing, Ahead-of-time (AOT) compilation is also enabled by default, which dynamically compiles certain methods into AOT code at runtime. By using these features in combination, startup performance is further improved because the cached AOT code can be used to quickly enable native code performance for subsequent runs of your application. For more information about AOT, see [AOT Compiler](aot.md).
+
+The contents of a shared classes cache can include the following artifacts:
+
+- Bootstrap classes
+- Application classes
+- Metadata that describes the classes
+- AOT-compiled code
+- Java byte arrays
```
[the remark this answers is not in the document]
hangshao0: I see when talking about SharedDataHelpers:
"`SharedDataHelpers`, which are used for storing Java byte array data".

I guess Java byte arrays here means `TYPE_BYTE_DATA`.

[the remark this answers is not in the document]
pshipton: Understood, but SharedDataHelpers isn't used by default, and I think listing it here may be confusing. Before OpenJ9 was created we didn't expose the documentation for SharedDataHelpers. Although it's possible to store data, I don't know that anyone does.

docs/shrc.md
```diff
@@ -22,33 +22,107 @@
...
+
+For a set of best practices when using class data sharing, see [Creating a shared classes cache](#creating-a-shared-classes-cache).
+
 
-- You can change the name and location of the default shared classes cache.
-- You can enable messages about the default shared classes cache by using the default command line option without the `silent` suboption: `-Xshareclasses:bootClassesOnly,nonFatal`.
+## Class data sharing operations
 
-:fontawesome-solid-pencil-alt:{: .note aria-hidden="true"} **Note:** If you have multiple VMs and you do not change the default shared classes behavior, the VMs will share a single default cache, assuming that the VMs are from a single Java installation. If the VMs are from different Java installations, the cache might be deleted and recreated; for more information, see the following section about best practices.
+When a VM loads a class, the VM looks in the shared classes cache to see if the class is already present. If it is, it loads the class from the cache. Otherwise, it loads the class from the filesystem and writes it into the cache.
```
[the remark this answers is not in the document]
SueChaplain: Updated, hopefully without changing the meaning:

*When a VM loads a class, the VM looks in the shared classes cache to see if the class is already present. If the class is present and the classpath or URL to load the class is a match, the VM loads the class from the cache. Otherwise, it loads the class from the filesystem and writes it into the cache.*

docs/shrc.md
```diff
@@ -22,33 +22,107 @@
...
 
-## Best practices for using `-Xshareclasses`
+Caches are not compatible between VMs that are using different object storage modes. For example, a 64-bit VM that uses compressed references to store 64-bit objects in a 32-bit representation, cannot share a cache with a 64-bit VM that is not using compressed references. For more information about object storage options, see [Compressed references](allocation.md#compressed-references).
+
+In the OpenJ9 implementation of `java.net.URLClassLoader`, classes are read from and written to the cache using the public Helper API. Therefore, any class loader that extends `java.net.URLClassLoader` gets class sharing support for free. Custom  class loaders that do not extend `java.net.URLClassLoader` must be adapted to share class data as described in [Support for custom class loaders](#support-for-custom-class-loaders).
```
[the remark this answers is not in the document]
SueChaplain: Using:

*In the OpenJ9 implementation of `java.net.URLClassLoader`, classes are read from and written to the cache using the public Helper API. Therefore, any class loader that extends `java.net.URLClassLoader` gets class sharing support for free provided that it continues to use the methods in `java.net.URLClassLoader` to load classes.*

docs/shrc.md
```diff
@@ -66,43 +140,138 @@ The [-Xshareclasses](xshareclasses.md) option is highly configurable, allowing y
 
     In addition, if you have VMs from different Java installations, of the same Java release and installed by the same user, each VM checks whether the existing default shared cache in the cache directory is from the same Java installation as the VM. If not, the VM deletes that shared cache, then creates a new one. Specifying a different cache directory for each Java installation avoids this situation.
 
-
-
 - Ensure that the cache directory permissions are set appropriately ([`-Xshareclasses:cacheDirPerm`](xshareclasses.md#cachedirperm)).
 
-    It is good practice to explicitly set permissions for the cache directory when the defaults are not appropriate.
+    It is good practice to explicitly set permissions for the cache directory when the defaults are not appropriate. Access is controlled by operating system permissions and Java security permissions; read/write access is the default and gives all users equal rights to update the cache. On Unix systems, you can use the [`-Xshareclasses:groupAccess`](xshareclasses.md#groupaccess) suboption to allow read/write permissions for groups as well as users.
```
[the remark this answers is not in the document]
SueChaplain: Changing:

*Access is controlled by operating system permissions and Java security permissions; read/write access is the default only for the current user. On Unix systems, you can use the [`-Xshareclasses:groupAccess`](xshareclasses.md#groupaccess) suboption to allow read/write permissions for groups as well as users*

docs/shrc.md
SueChaplain: Changed to:

*In most cases, setting this option allows your application to start even if there is a problem opening or creating the shared cache. The VM will continue to start *without* class data sharing.*

I added the extra info you provided into the `xshareclasses.md` topic as a note to the `nonfatal` suboption.

docs/shrc.md
SueChaplain: I've moved it into a bullet about setting a soft maximum size, and linked to the `-Xscmx` topic where we have more information.

docs/shrc.md
SueChaplain: I think I worked this in ok by moving it to the top, but let me know if you have an alternative suggestion: 

*-  Before starting your application, use the [`-Xshareclasses:listAllCaches`](xshareclasses.md#listallcaches) cache utility to identify a suitable cache that your application might share.*
- *This option lists all the caches that exist in the default directory, including compatible and incompatible caches. You can also specify the `cacheDir` suboption to look for caches in a specified directory.*

Then changed the lead in to the next bullet:

*- If you are creating a new cache, .....*

docs/shrc.md
SueChaplain: I altered (4). I also went to alter (3), which has 2 mentions, but am unsure whether you intended this or not. Please confirm:

*3. The next Docker layer up is an **application** image that is built on the OpenJ9 image. As part of this image, the `-Xshareclasses:name=MyCache,layer=1` suboption is used to create another cache called `MyCache`. Because the `layer=1` suboption is specified, this new cache is a layered cache, which builds on `MyCache` in the previous container layer. (**For example**, Open Liberty starts two VMs, so if you instead use the `createLayer` suboption here, two layered caches are created, with layer numbers of 1 and 2.) Note that cache layers are different from, and independent of, container layers.*

docs/shrc.md
SueChaplain: used *can*, rather than *may*. If you meant *might*, please let me know. 
(we tend not to use *may* as it is not precise)

docs/shrc.md
```diff
@@ -22,33 +22,107 @@
...
-## Best practices for using `-Xshareclasses`
+Caches are not compatible between VMs that are using different object storage modes. For example, a 64-bit VM that uses compressed references to store 64-bit objects in a 32-bit representation, cannot share a cache with a 64-bit VM that is not using compressed references. For more information about object storage options, see [Compressed references](allocation.md#compressed-references).
+
+In the OpenJ9 implementation of `java.net.URLClassLoader`, classes are read from and written to the cache using the public Helper API. Therefore, any class loader that extends `java.net.URLClassLoader` gets class sharing support for free. Custom  class loaders that do not extend `java.net.URLClassLoader` must be adapted to share class data as described in [Support for custom class loaders](#support-for-custom-class-loaders).
+
+### AOT code and JIT data
+
+OpenJ9 can automatically store small amounts of AOT code and JIT data, which helps improve performance in the following ways:
+
+- The VM dynamically compiles certain methods into AOT code at runtime. Subsequent VMs that attach to the cache can take advantage of the compiled code to start faster.
```
mpirvu, line 106: Nitpick: It's the JIT compiler that "compiles certain methods into AOT", therefore I would change `VM` into `JIT compiler`

docs/shrc.md
SueChaplain: I'm not clear what you are meaning here. *As does the next point, ...*? Do you mean that you only need to grant permission to `SharedCachePermission` *if* you are using a security manager? 

i.e. 

*If a running application uses its own class loader **and you are using a `SecurityManager`**, you must grant the class loader permission to `SharedCachePermission` before they can share classes. To grant permission, add shared class permissions to the `java.policy` file by specifying the `ClassLoader` class name. Permissions can be set for `read`, `write`, or `read,write`.*

Then remove the sentence *"If a `SecurityManager` is being used, ...."*

mpirvu:
@dsouzai FYI, should you want to make small amends to the AOT text

docs/dcomibmotisharedsharedclassglobalfilterclass.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Syntax
+
+        -Dcom.ibm.oti.shared.SharedClassGlobalFilterClass=<filter_class_name>
+
+This property is not set by default.
+
+## Explanation
+
+A filter can be used to decide which classes are found and stored by a custom class loader in a shared classes cache. The filter is applied to a particular package.
```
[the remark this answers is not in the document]
SueChaplain: I added some more clarity about what is needed:

*- `setSharingFilter`: A filter that can be used to decide which classes are found and stored in the cache. This filter can be applied to a particular package by implementing the `SharedClassFilter` interface.*

At the top of the Helper section there is a link to the `com.ibm.oti.shared` API. With the interface they need to use mentioned, I think users will find it ok without linking to the source.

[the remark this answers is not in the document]
SueChaplain: Oops - updated in the `-D` topic too and added the API doc as a *See also* link.

docs/introduction.md
```diff
@@ -79,7 +79,7 @@ For more information, see the [API documentation](api-overview.md).
 ### Cloud optimizations
 To improve the performance of applications that run in containers, try setting the following tuning options:
 
-- Use a shared classes cache (`-Xshareclasses -XX:SharedCacheHardLimit=200m -Xscmx60m`) with Ahead-Of-Time (AOT) compilation to improve your startup time. For more information, see [Class Data Sharing](shrc.md) and [AOT Compiler](aot.md).
+- Use a shared classes cache (`-Xshareclasses -XX:SharedCacheHardLimit=200m -Xscmx60m`) with Ahead-Of-Time (AOT) compilation to improve your startup time. For more information, see [Inroduction to class data sharing](shrc.md) and [AOT Compiler](aot.md).
```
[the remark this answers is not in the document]
SueChaplain: Adding in a one liner for now:

*- Use a shared classes cache (`-Xshareclasses -XX:SharedCacheHardLimit=200m -Xscmx60m`) with Ahead-Of-Time (AOT) compilation to improve your startup time. For persistence, store the cache in a volume that you map to your container.*

SueChaplain:
Jenkins doc stage

dsouzai:
AOT text looks ok to me.

hangshao0 (approved):

docs/shrc.md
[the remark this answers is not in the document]
pshipton: Please also add, "and the class loader is enabled for class sharing". After "When a VM loads a class" seems a good place.

docs/shrc.md
pshipton: Yes.

docs/shrc.md
pshipton: I don't think we should say " to identify a suitable cache that your application might share". It's best if every application has it's own cache. There should be a point for that somewhere. `listAllCaches` is for maintenance, such as to find a delete obsolete caches.

docs/shrc.md
pshipton: "Open Liberty starts two VMs" is a valid warning, I think it should stay. Otherwise we can be generic in point 4.

pshipton: If the .jpg can be changed, it also mentions Open Liberty.

docs/shrc.md
pshipton: After " shared classes cache can be persistent or non-persistent." we can add "Persistent caches are memory mapped files.".

docs/xshareclasses.md
```diff
@@ -439,11 +409,10 @@ behavior, which can improve the performance of class loading from the shared cla
 
         -Xshareclasses:nonfatal
 
-:   Allows the VM to start even if class data sharing fails. Normal behavior for the VM is to refuse to start if class data sharing fails. If you select `nonfatal` and the shared classes cache fails to initialize, the VM attempts to connect to the cache in read-only mode. If this attempt fails, the VM starts without class data sharing. See also [`fatal`](#fatal).
+:   Allows the VM to start, in most cases, even if class data sharing fails. Normal behavior for the VM is to refuse to start if class data sharing fails. If you select `nonfatal` and the shared classes cache fails to initialize, the VM attempts to connect to the cache in read-only mode. If this attempt fails, the VM starts without class data sharing. See also [`fatal`](#fatal).
```
pshipton, line 412: I still think we should mention: The `nonfatal` suboption doesn't protect against cache corruption caused by bugs, when the corruption isn't detected at the time the shared cache is opened.

docs/shrc.md
SueChaplain: Updated as follows:

*- Before starting your application, use the [`-Xshareclasses:listAllCaches`](xshareclasses.md#listallcaches) cache utility **to review and maintain the existing caches on your system**.*

 - *This option lists all the caches that exist in the default directory, including compatible and incompatible caches. You can also specify the `cacheDir` suboption to look for caches in a specified directory. **Remove any obsolete caches, as described in [Housekeeping](#housekeeping)**.*

docs/xshareclasses.md
SueChaplain: There is a note underneath that you might have missed. See:

***Note:** Unless it is important that your application runs with class data sharing, it is good practice to set this parameter. See [Creating a shared classes cache](shrc.md#creating-a-shared-classes-cache). However, cache corruption as a result of a bug in the operating system, VM, or user code might not be detected when opening the cache. In this situation, the cache is used and the application might crash.*

Is this ok?

SueChaplain:
Jenkins doc stage

SueChaplain:
@pshipton - please merge if you are happy or suggest further changes.

docs/xshareclasses.md
pshipton: Looks good.

docs/shrc.md
pshipton: We can put middleware at cache layer 1 in the picture.

docs/shrc.md
```diff
@@ -22,37 +22,116 @@
...
+# Introduction to class data sharing
 
-Sharing class data between JVMs improves start up performance and reduces memory footprint.
+Sharing class data between OpenJ9 VMs improves start up performance and reduces memory footprint. Consider the following outcomes for two VMs that are running similar Java applications but sharing class data:
 
-Start up performance is improved by placing classes that an application needs when initializing into a shared classes cache. The next time the
-application runs, it takes much less time to start because the classes are already available. When you enable class data sharing, AOT compilation is also enabled by default, which dynamically compiles certain methods into AOT code at runtime. By using these features in combination, startup performance can be improved even further because the cached AOT code can be used to quickly enable native code performance for subsequent runs of your application.
+- Start up performance is improved by placing classes that each application needs when initializing into a shared classes cache. The next time the application runs, it takes much less time to start because the classes are already available.
```
pshipton, line 29: The performance is actually due to AOT code, JIT hints and profiles.

docs/shrc.md
```diff
@@ -22,37 +22,116 @@
...
+
+For a set of best practices when using class data sharing, see [Creating a shared classes cache](#creating-a-shared-classes-cache).
+
+
+## Class data sharing operations
+
+When a VM loads a class and the class loader is enabled for class sharing, the VM looks in the shared classes cache to see if the class is already present. If the class is present and the classpath or URL to load the class is a match, the VM loads the class from the cache. Otherwise, it loads the class from the filesystem and writes it into the cache.
+
+The VM detects filesystem updates by storing timestamp values into the cache and comparing the cached values with actual values on each class load operation. In this way, the VM detects when a class might be invalidated and can mark the class as *stale*. These operations happen transparently when classes are loaded, so users can modify and update as many classes as they like during the lifetime of a shared class cache, knowing that the correct classes are always loaded. Stale classes are *redeemed* if the same class is subsequently fetched by the class loader from another VM and checked against the stale class in the cache.
```
pshipton, line 95: "on each class load operation" isn't correct, unless loading classes from a directory. For a jar file, it's checked once when opened, unless https://www.eclipse.org/openj9/docs/xshareclasses/#checkurltimestamps is used.

docs/shrc.md
```diff
@@ -66,43 +145,141 @@ The [-Xshareclasses](xshareclasses.md) option is highly configurable, allowing y
...
+A `SharedDataHelperFactory` provides an interface that can be used to create `SharedDataHelpers`, which are used for storing Java byte array data. A `SharedDataHelper` also has a one to one relationship with a class loader, although a class loader can exist without a `SharedDataHelper`.
+
+
+### The Java shared classes utility API
+
+The following APIs are available for obtaining information about a shared classes cache:
+
+- `com.ibm.oti.shared.SharedClassStatistics`: Obtains information about cache size, including free space, soft maximum limit, and the limits enforced for AOT and JIT data.
+- `com.ibm.oti.shared.SharedClassUtilities`: Obtains detailed information about a shared classes cache, including its size, name, type, and status.
+- `com.ibm.oti.shared.SharedClassCacheInfo`: Stores information about a shared classes cache and provides API methods to retrieve the information and remove caches. For example, you can use the `IterateSharedCaches` and `DestroySharedCache` [JVMTI extensions](interface_jvmti.md#iteratesharedcaches).
```
pshipton, line 274: This shouldn't contain "For example, ". You can use the jvmti extensions as an alternate to SharedClassCacheInfo.

docs/shrc.md
SueChaplain: I toned down "much less" to "less". Note that credit is given to AOT improving performance a couple of paragraphs down. I've also added that further improvements are gained by storing JIT data and profiles.

pshipton:
There was information about running with a SecurityManager that seems to be gone now.

docs/shrc.md
```diff
@@ -66,43 +145,141 @@ The [-Xshareclasses](xshareclasses.md) option is highly configurable, allowing y
...
+
 
 ## Support for custom class loaders
 
-Classes are shared by the bootstrap class loader internally in the OpenJ9 VM. If you are using a custom class loader, you can use the Java Helper API to find and store classes in the shared classes cache.
+Classes are shared by the bootstrap class loader internally in the VM. The OpenJ9 implementation of `java.net.URLClassLoader` is modified to use `SharedClassURLClasspathHelper` and any class loaders that extend `java.net.URLClassLoader` can inherit this behavior. If you are using a custom class loader that does not extend `java.net.URLClassLoader`, you can use the Java Helper API to find and store classes in a shared classes cache.
 
-For more information, see [Using the Java Helper API](https://www.ibm.com/support/knowledgecenter/SSYKE2_8.0.0/com.ibm.java.vm.80.doc/docs/shrc_pd_helper.html).
+If a running application uses its own class loader, you must grant the class loader permission to `SharedCachePermission` before they can share classes. To grant permission, add shared class permissions to the `java.policy` file by specifying the `ClassLoader` class name. Permissions can be set for `read`, `write`, or `read,write`.
```
pshipton, line 249: This is only required when the SecurityManager is enabled, which I thought was already fixed. Also I though there was a second paragraph about permissions, which seems to be gone now.

SueChaplain: I'm confused. I made some changes because of this thread, which I thought was ok? +1 https://github.com/eclipse/openj9-docs/pull/752#discussion_r604094290

pshipton: There was a sentence "If a SecurityManager is being used, ....", which I guess I accidently indicated could be removed, but I think the details where useful because they were different from the above. Possibly the "If a SecurityManager is being used" part could be removed, since the above paragraph should contain this.

SueChaplain: I looked back at that comment and saw that I missed adding the SecurityManager part of the agreed sentence. However, I thought you agreed to me removing the 2nd paragraph. I've put part of that back but I'll have to figure out what disappeared when I force pushed.

pshipton: Looking at the knowledgecenter I only find the following which isn't included. Not sure if this is the sentence that was removed.

If a SecurityManager is being used with class data sharing and the running application is calling the com.ibm.oti.shared.SharedClassUtilities APIs getSharedCacheInfo() or destroySharedCache(), you must grant the code calling these APIs the appropriate SharedClassesNamedPermission. For example:

pshipton: I also note there are examples in the knowledgecenter which weren't brought over. 
```
permission com.ibm.oti.shared.SharedClassPermission
        "com.abc.customclassloaders.*", "read,write";
```

SueChaplain:
Jenkins doc stage

SueChaplain:
@pshipton - I think I got those last few changes that you wanted. Hopefully ready now. More to come in a second wave for SHRC.

pshipton (approved):

PR #759, opened 2021-03-23
Fix staging and draft banner

SueChaplain:
We need to pull the base.html file from the material theme in order
to insert our staging/draft docs banner. It is not needed as a
permanent artifact in the repo, we can just bring it in at build time.

Signed-off-by: Sue-Chaplain <sue_chaplain@uk.ibm.com>

### DIFF
--- a/JenkinsFile_build_doc.groovy
+++ b/JenkinsFile_build_doc.groovy
@@ -107,13 +107,15 @@ timeout(time: 6, unit: 'HOURS') {
                             if (params.ghprbPullId) {
                                 // If a pull request, use staging site banner
                                 sh """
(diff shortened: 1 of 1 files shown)


SueChaplain:
@AdamBrousseau - please can you check the groovy syntax looks ok?

AdamBrousseau:
Jenkins doc stage

AdamBrousseau:
Jenkins doc stage

SueChaplain:
Jenkins doc stage

AdamBrousseau (approved):

pshipton:
Ideally it should be in the repo so it's tracked. Can it be put in a different location in the repo and copied to the right place at build time? I'll merge this for now.

pshipton (approved):

AdamBrousseau:
If it was put in the repo twice we could avoid the sed too. Not that it's been a problem yet but it could avoid unexpected changes? But maybe we'd prefer it kept in line with upstream?

SueChaplain:
We used to modify and use this file in our builds but it always changes upstream and has caused a lot of issues during upgrades. This time we dropped our dependency on it but forgot that it was used to inject the banner. I didn't spot it because I couldn't use the standard staging job to host the newer flavour of docs for review .... I had to build it locally in the new container  and push directly to the staging server. 

We could put it back but would need to overwrite it at each upgrade. 

I'd rather we avoid storing 2 versions of `base.html` for each banner. It's worked this way for a long time without an issue.

PR #787, opened 2021-05-10
Improve mapping of GC policies to GC options

SueChaplain:
- Add clarity for GC options regarding which policies can and can't use them.
- Move a few metronome `-Xgc` options from `-Xgcpolicy` to their correct location.
- Some tidying up/editing improvements.

Signed-off-by: SueChaplain <Sue_chaplain@uk.ibm.com>


doveye (approved):
LGTM

PR #790, opened 2021-05-18
Updates for Material theme 7.1.4

davehobbs:
Includes updates to:
- Switch search off for the offline build . The material theme does not support search on local sites
- Correct the table layout in Supported environments which is broken by the material upgrade
- Use the `mkdocs-macros-plugin` to allow correct links  to offline api doc
- Fix relative urls to api docs due to path changes caused by use_directory_urls
- Explain how to reference non-Markdown files in the contributions guidelines


davehobbs:
Jenkins build container

davehobbs:
Jenkins build container

Issue #766, opened 2021-04-08
Adaptive GC Threading Documentation Updates


RSalman:
**Issue or pull request number:**
https://github.com/eclipse/openj9-docs/pull/765

**Overview:**
With the introduction of Adaptive GC Threading optimization (eclipse/openj9#11999 &
eclipse/omr#5830), new options need to documented and some of the existing options need to be updated for clarification.

Discussion on the behaviour, naming and interaction with existing options can be found at eclipse/openj9#12119.

**Release target:**
Eclipse OpenJ9 0.27.0

**Applies to the following JDK versions:**
8 and Later

**Applies to the following platforms:**
All platforms (Only applicable to the default GC GENCON Policy)

**For new command line options:**
XX:
-XX:[+-]AdaptiveGCThreading to enable/disable the GC adaptive threading optimization.
-XX:ParallelGCMaxThreads= Mapped to Xgcmaxthreads (see below)

Xgc:
-Xgcmaxthreads The max number of threads available to Adaptive Threading (Similar to Xgcthreads, except it doesn’t set enforce thread count). for default value/range of values see existing `Xgcthreads` option

Existing Xgcthreads & XX:ParallelGCThreads= option pages updated with notes to compare with new options.

RSalman:
@pshipton could you please confirm that the changes are Applies to JDK versions *8 and Later*

pshipton:
Yes, the same OpenJ9 code stream applied to all jdk8, jdk11, jdk16, and the future jdk17+

SueChaplain:
Closed by #765.

PR #765, opened 2021-04-07
Adaptive GC Threading Documentation

RSalman:
With the introduction of Adaptive GC Threading optimization  (https://github.com/eclipse/openj9/pull/11999 & 
 https://github.com/eclipse/omr/pull/5830), new options need to  documented and some of the existing options need to be updated for  clarification.

Discussion on the behaviour, naming and interaction with existing  options can be found at https://github.com/eclipse/openj9/issues/12119.

New options:
**XX:**
`-XX:[+-]AdaptiveGCThreading` to enable/disable the GC adaptive threading  optimization.
`-XX:ParallelGCMaxThreads=` Mapped to Xgcmaxthreads (see below)

**Xgc:**
`-Xgcmaxthreads` The max number of threads available to Adaptive Threading  (Similar to Xgcthreads, except it doesn’t set forced thread count flag)

Existing `Xgcthreads` & `XX:ParallelGCThreads=` option pages updated  with notes to compare with new options.

Signed-off-by: Salman Rana <salman.rana@ibm.com>

PR #791, opened 2021-05-19
Update AdaptiveGCThreading release note

pshipton:
AdaptiveGCThreading is enabled by default.

Issue https://github.com/eclipse-openj9/openj9-docs/issues/766

This updates https://github.com/eclipse-openj9/openj9-docs/pull/765

### DIFF
--- a/docs/version0.27.md
+++ b/docs/version0.27.md
@@ -34,7 +34,11 @@ The following new features and notable changes since v 0.26.0 are included in th
 
 ### New `-XX:[+|-]AdaptiveGCThreading` option added
 
-To optimize performance, you can now enable adaptive threading to automatically tune the number of active parallel garbage collection (GC) threads. When enabled, the GC thread count is dynamically adjusted from collection cycle to cycle to account for changes in the workload and consider the available resources. When parallel workloads decrease, less threads can be used, reducing the overhead and freeing up resources for other processing activities.
+Adaptive threading is enabled by default, which automatically tunes the number of active parallel garbage collection (GC) threads.
+When this feature is enabled, the GC thread count is dynamically adjusted from collection cycle to cycle to account for changes in the the amount
+of time that parallel threads spend doing useful GC work (such as object graph traversal) compared to time spent synchronizing among themselves.
+When GC work decreases, fewer threads are used, which reduces the overhead, effectively reducing GC pause times.
+Resources are freed up for other processing activities.
 
 Use the [`-xgcmaxthreads`](xgcmaxthreads.md) option with the [`-XX:+AdaptiveGCThreading`](xxadaptivegcthreading.md) option to specify a thread count limit.
 


pshipton:
Jenkins doc stage

pshipton:
The changes aren't obvious since I modified the text to be on multiple lines instead of one long line. I modified the first sentence, and added ", effectively reducing GC pause times`" to the last sentence.

pshipton:
Staged to https://eclipse-openj9.github.io/openj9-docs-staging/791/version0.27/

docs/version0.27.md
```diff
@@ -34,7 +34,10 @@ The following new features and notable changes since v 0.26.0 are included in th
 
 ### New `-XX:[+|-]AdaptiveGCThreading` option added
 
-To optimize performance, you can now enable adaptive threading to automatically tune the number of active parallel garbage collection (GC) threads. When enabled, the GC thread count is dynamically adjusted from collection cycle to cycle to account for changes in the workload and consider the available resources. When parallel workloads decrease, less threads can be used, reducing the overhead and freeing up resources for other processing activities.
+Adaptive threading is enabled by default, which automatically tunes the number of active parallel garbage collection (GC) threads.
+When enabled, the GC thread count is dynamically adjusted from collection cycle to cycle to account for changes in the workload and consider
+the available resources. When parallel workloads decrease, less threads can be used, reducing the overhead, freeing up resources for other
```
SueChaplain, line 39: The last sentence in this paragraph is a little long and would read better if broken up. Something like: 

When parallel workloads decrease, less threads are used, which reduces the overhead. Resources are freed up for other processing activities, which effectively reduces GC pause times.

pshipton: Updated, except I used "fewer threads are used".

amicic: The last sentence isn't really explaining why GC pauses may be lower.... Something like this is more correct:


When enabled, the GC thread count is dynamically adjusted from collection cycle to cycle to account for changes in time parallel threads spend doing useful GC work (like object graph traversal) vs time synchronizing among themselves.

doveye: I'll put my comment here rather than via the review so it's easier to see the thread of comments. I'd just make some tweaks to the latest suggestion:

"When this feature is enabled, the GC thread count is dynamically adjusted from collection cycle to cycle to account for changes in the amount of time that parallel threads spend doing useful GC work (such as object graph traversal) compared to time spent synchronizing among themselves."

pshipton: The new text suggested was already partially there in a previous sentence. I've updated it with the suggested modifications, and re-jigged the last two sentences.

amicic: It's ok, although I was trying to avoid word 'workload'. It may mislead the reader we are talking about application parallel workload, while we are working about GC specific work (parallel roots processing, object graph traversal,...)

amicic (approved):

docs/version0.27.md
pshipton: I've used your words to remove "workload" and better describe the GC work.

amicic: ok. 

you only need once to go into details what GC work is (root processing, object graph traversal....)

pshipton: Going point, fixed it up.

doveye (approved):

PR #797, opened 2021-06-21
Add note about requiring the openj9.dtfj module

kgibm:
Fixes https://github.com/eclipse-openj9/openj9-docs/issues/798

Signed-off-by: Kevin Grigorenko <kevin.grigorenko@us.ibm.com>

### DIFF
--- a/docs/interface_dtfj.md
+++ b/docs/interface_dtfj.md
@@ -47,6 +47,14 @@ DTFJ is implemented in pure Java and tools written using DTFJ can be cross-platf
 
 See the [DTFJ API documentation](api-dtfj.md). <!-- Link to API -->
 
+Note: If the code that loads DTFJ is in a module, the module must require the `openj9.dtfj` module. For example:
+
+```
+module MyModule {
+  requires openj9.dtfj;
+}
+```
+
 ## Using the DTFJ interface
 
 To create applications that use DTFJ, you must use the DTFJ interface. Implementations of this interface have been written that work with system dumps and Java dumps.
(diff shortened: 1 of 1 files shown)


docs/interface_dtfj.md
```diff
@@ -47,6 +47,14 @@ DTFJ is implemented in pure Java and tools written using DTFJ can be cross-platf
 
 See the [DTFJ API documentation](api-dtfj.md). <!-- Link to API -->
 
+Note: If the code that loads DTFJ is in a module, the module must require the `openj9.dtfj` module. For example:
```
pshipton, line 50: The documentation is generic and this only applies to Java 11+ so pls mention that.

kgibm: Does it strictly apply to Java >= 11, or does it also apply to Java >= 9 with modules? I can test if needed.

keithc-ca: Perhaps making it explicit is helpful, but the code can't be in a module if you're not using Java 11+.

keithc-ca: It would apply to Java 9 and Java 10 if there were any supported VMs. Saying Java >= 9 might be technically more correct, but it might also give people the idea they should try finding jdk9, which isn't helpful. On reflection, I think it might be better without a reference to any version: the 'module' concept is the only real consideration here.

keithc-ca (approved):

doveye:
Jenkins doc stage

doveye (changes requested):
Looks good, but for consistency, can you please code the Note with bold font and a font awesome icon as described in the contribution guidelines, thanks :)
https://github.com/eclipse-openj9/openj9-docs/blob/master/CONTRIBUTING.md

doveye:
Jenkins doc stage

doveye (approved):

Issue #803, opened 2021-07-16
-Xmint/-Xmaxt descriptions misleading


cedrichansen:
**Subject:**
*-Xmint/-Xmaxt page*

https://www.eclipse.org/openj9/docs/xmint/ describes the option correctly, but provides a misleading default values.
The default values are displayed as follows
Setting | Effect | Default
-- | -- | --
-Xmint<value> | Set minimum time in GC | 5
-Xmaxt<value> | Set maximum time in GC | 13

Users would then think that they could specify

`-Xmint5 -Xmaxt10` on the command line, but this causes the following JVM initialization error

```
$ ./java -Xmint5 -Xmaxt10 -version
JVMJ9GC033E -Xmaxt value must be between 0.00 and 1.00 (inclusive)
JVMJ9VM015W Initialization error for library j9gc29(2): Failed to initialize; unable to parse command line
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```

Rather, the correct format for 5% & 10% is `-Xmint0.05 -Xmaxt0.10`

**Suggestion:**
Display the defaults as follows

Setting | Effect | Default
-- | -- | --
-Xmint<value> | Set minimum time in GC | 0.05 
-Xmaxt<value> | Set maximum time in GC | 0.13

`-Xminf/-Xmaxf` is consistent with how the option should be specified in this regard (example/default is 0.0-1.0 based)
as is `-Xloainitial / -Xloaminimum / -Xloamaximum`

cedrichansen:
@amicic @dmitripivkine fyi

doveye:
The docs have been updated

PR #810, opened 2021-07-22
Fix default -Xmint/-Xmaxt values

doveye:
The format of the values listed should be a percentage (see https://github.com/eclipse-openj9/openj9-docs/issues/803)


doveye:
@keithc-ca could you please merge this PR. I thought merging wouldn't be possible due to the legal check but looks like it is. I think the legal check fail is due to the PR kicking off a Docker build, which it shouldn't have (see Slack message https://openj9.slack.com/archives/C8312LCV9/p1626937051074700).

keithc-ca (approved):

PR #806, opened 2021-07-20
Openj9 docs 0.27.0

AdamBrousseau:
Built on
https://github.com/eclipse-openj9/openj9/commit/1851b0074f87e20c0007c6190c745dab9760eabe
https://github.com/eclipse-openj9/openj9-omr/commit/9db1c870dfbaf4cf677cb7ba086ea303e83d7d81
https://github.com/ibmruntimes/openj9-openjdk-jdk8/commit/de702c317454d1cf9ad264f15a8fb2d1431d7006
https://github.com/ibmruntimes/openj9-openjdk-jdk11/commit/21849e2ca05db8c5bb1efb8fc335d8931dec81bf
https://github.com/ibmruntimes/openj9-openjdk-jdk16/commit/34df42439f3be06211b15569253d33921591eb1d

PR #811, opened 2021-07-27
Apidoc update july2021

AdamBrousseau:


pshipton:
@doveye we want to merge this head stream update first to pick up all the major changes, and then do another build for 0.27 and see what's changed. Do you need to stage and check things, or can this just be merged?

doveye:
@pshipton we wouldn't be checking technical updates but it would be good to check the general appearance is still ok. Since we have a 0.27 branch for the docs though, I guess we could just merge the head stream, then check that before we merge the 0.27 version. 
BTW, the link to the OpenJ9 repo (gets added to the bottom of each API doc page) should change due to the general project moves. Not sure if that has been done already (looks like there's a redirect in place so not essential) but it's something else we can check.

pshipton:
> Not sure if that has been done already

Yes, it's part of this change.

pshipton:
I've pushed this change to the openj9-docs-0.27.0 branch as well, and we'll update the branch via https://github.com/eclipse-openj9/openj9-docs/pull/806 if necessary to correspond to the 0.27 release.

pshipton:
I was hoping these changes would show up at https://eclipse-openj9.github.io/openj9-docs but they have not so far. @doveye do you need to do something to rebuild the docs?

AdamBrousseau:
Looks like we didn't get an automated build. I kicked one
https://openj9-jenkins.osuosl.org/view/Website-Doc/job/Build-Doc-Push_to_ghpages/13/
Should be up shortly

doveye:
Looks good, thanks Adam and Peter :)

Issue #816, opened 2021-08-05
JITServer -XX options need to be documented


dmitry-ten:
JITServer added a bunch of `-XX` options, such as `+UseJITServer`, `+RequireJITServer`, `JITServerAddress=`, etc.

We need to document each option in preparation for the JITServer GA.

dmitry-ten:
I'll make a pull request once I write the required documentation.
@mpirvu fyi.

doveye:
@dmitry-ten Can you please provide more information. There is a [template](https://github.com/eclipse-openj9/openj9-docs/issues/new?assignees=&labels=&template=new-documentation-change.md) to help when creating issues. For example:

- What's the related code issue?
- Which release is this targeted at? (0.28?)
- Which JDK versions does this apply to? (Currently, the docs say JDK 8 and 11 but 0.28 supports JDK 17 so will JITServer be available on that version too?)
- Which platforms does this apply to? (There is already a list in the docs, are there any changes to that list?)

You mentioned GA - presumably you mean that JITServer will go from a tech preview to a fully-fledged feature? Is that due to happen for 0.28? 

- If yes, then we can go ahead with the separate topics for the -XX options but we'll need to remove "(tech preview)" from the mdkocs.yml file and the title of the jitserver.md topic, and rewrite that topic to move more of the content to the -XX topics. I can probably help with that.
- If no, then we shouldn't create separate topics for the -XX options - any that aren't already mentioned in the jitserver.md topic should be added there, to keep all the tech preview content separate from the rest of the docs.

Changes for 0.28 should be summarized for users in the what's new topic: https://eclipse-openj9.github.io/openj9-docs/version0.28/ (you can see an example at https://eclipse-openj9.github.io/openj9-docs/version0.23/#extended-platform-support-for-the-jitserver-technology-preview).

mpirvu:
**Issue or pull request number:**
There are several PRs that are related to this documentation, most of them quite old.
A couple of the recent ones: https://github.com/eclipse-openj9/openj9/pull/13303 https://github.com/eclipse-openj9/openj9/pull/13288 

**Overview:**
*JITServer tech has a few `-XX:` options that can be used for user tuning. These options need to be properly documented much like the other XX options.*

**Release target:**
*Eclipse OpenJ9 0.28.0*

**Applies to the following JDK versions:**
[...]

PR #830, opened 2021-09-07
Further updates to JITServer docs for 0.28

doveye:
https://github.com/eclipse-openj9/openj9-docs/issues/816

Signed-off-by: Esther <doveye@uk.ibm.com>


doveye:
Jenkins doc stage

doveye:
@mpirvu @dmitry-ten I've made some further doc updates. Can you please review? https://eclipse-openj9.github.io/openj9-docs-staging/830/version0.28/

Overview of changes:
- I removed the list of supported JDK versions, since this is no longer a tech preview (except on z) and the options should work on all JDK versions (I'm assuming that if there was an issue on an non-LTS version, we would still fix it)
- I removed "tech preview" but added a note to say that on z/OS, JITServer is still a tech preview
- I moved the 3 SSL options into one topic (there are still 3 separate toc entries) since they have to be used together so it makes sense for them to be documented in the same place. We already do this for other options, eg https://www.eclipse.org/openj9/docs/xloaminimum/.  I moved the detail of the security setup there too, to avoid duplicating content.
- I removed the table of options from the JITServer topic and instead added a list, so users can get the detail in the individual option topics (so we don't repeat content, which can lead to disparities)
- I added a section to the release notes topic, version0.28.md
- I added links to the JITServer topic from all the option topics
- I made some grammatical updates to some of the content and added some more mentions of JITServer in the options topics so it's clearer to a user who has arrived from outside the docs that the topic relates to JITServer

dmitry-ten:
LGTM

docs/xxjitservertimeout.md
```diff
@@ -26,20 +26,23 @@
...
 | Setting                 | Effect | Default                                                                            |
 |-------------------------|--------|:----------------------------------------------------------------------------------:|
-|`-XX:JITServerTimeout`           | Set the timeout value in milliseconds for socket operations | **Note 1**                                                                                    |
+|`-XX:JITServerTimeout`           | Set the timeout value in milliseconds for socket operations | 30000 ms for the JITServer process and 10000 ms when OpenJ9 is launched as a client VM                                                                                    |
 
-**Note 1:** The timeout default is 30000 ms for the JITServer process, and 10000 ms when OpenJ9 is launched as a client VM.
 
 ## Explanation
 
 The default values should be reasonable for most use cases.
```
mpirvu, line 43: I am thinking of removing this text because it's not really an explanation and does not have much value.

mpirvu (approved):
Looks good to me. I only have a small suggestion to remove the "Explanation" from `xxjitservertimeout.md`

docs/xxjitservertimeout.md
doveye: Thanks, I removed the text.

doveye:
@keithc-ca could you merge this doc issue please? It has technical approval.

mkdocs.yml
```diff
@@ -373,9 +373,9 @@ nav:
             - "-XX:[+|-]JITServerLogConnections"                                 : xxjitserverlogconnections.md
             - "-XX:JITServerPort"                                                : xxjitserverport.md
             - "-XX:JITServerShareROMClasses"                                     : xxjitservershareromclasses.md
-            - "-XX:JITServerSSLCert"                                             : xxjitserversslcert.md
-            - "-XX:JITServerSSLKey"                                              : xxjitserversslkey.md
-            - "-XX:JITServerSSLRootCerts"                                        : xxjitserversslrootcerts.md
+            - "-XX:JITServerSSLCert"                                             : xxjitserversslcert.md             
+            - "-XX:JITServerSSLKey"                                              : xxjitserversslcert.md        #redirect
+            - "-XX:JITServerSSLRootCerts"                                        : xxjitserversslcert.md        #redirect
```
keithc-ca, line 378: Please remove the trailing whitespace from these lines.

doveye: Done, although I don't believe it will affect the docs. Is there another reason why trailing whitespace is undesirable?

keithc-ca: In some file formats it isn't significant, but in some contexts it can lead to hard-to-diagnose problems: in my opinion it's just safer to avoid them everywhere. In fact, openj9 and omr now have a check that pull requests don't introduce any new trailing whitespace.

keithc-ca (approved):

mkdocs.yml
doveye: Interesting, I wasn't aware of that. I'll try to look out for trailing whitespace in future.

doveye:
Jenkins doc stage

PR #867, opened 2021-11-23
Move the build job to alinux

pshipton:
xlinux is slow in good times, and having networking issues atm. If it
works, should be much faster on alinux.


pshipton:
jenkins doc stage

pshipton:
jenkins doc stage

pshipton:
Seems not portable to other platforms.

PR #884, opened 2022-01-24
Remove stale footnote references and add planned releases

### LINKED ISSUE #883: Stale footnote references should be removed
On https://www.eclipse.org/openj9/docs/openj9_support/, there are footnotes for past releases that identify them as future, expected releases.

keithc-ca:
Fixes #883.


pshipton:
Since you are modifying this, can you please add lines for 0.31, 0.32, and remove 0.25 and 0.26.

keithc-ca:
Updated.

pshipton:
I don't think 0.30 and 0.31 need the `(1)` footnote. OpenJDK already released what we need for 0.30, and 0.30 builds will be published when this doc change goes live at the main site (https://www.eclipse.org/openj9/docs/). Java 18 is well in progress at OpenJDK, I don't think there is any question about it being released.

keithc-ca:
Removing the (1) footnote from 0.30.0 seems reasonable, given that release is imminent.

I think the footnote for 0.31.0 should remain: it's likely it will be released around that time, but it's not guaranteed.

I see 0.33 listed at https://github.com/eclipse-openj9/openj9/milestones; you didn't explicitly mention it and I didn't notice it earlier. Do you think we should list it as well?

pshipton:
You can list as many future releases as you want. I don't have a concern with listing 0.33 (or removing 0.27).

keithc-ca:
Updated (footnote removed from 0.30.0; release 0.33.0 added).

keithc-ca:
jenkins doc stage

keithc-ca:
The check-marks for 0.31.0 are the wrong colour: I will fix that.

keithc-ca:
jenkins doc stage

keithc-ca:
That looks better: https://eclipse-openj9.github.io/openj9-docs-staging/884/openj9_support/

pshipton (approved):

PR #898, opened 2022-02-10
Update support to remove CentOS/RHEL 8 from 31/32 bit, RHEL 7 from Z31

pshipton:
RHEL 8 doesn't provide 32-bit support drivers. OpenJ9 only supports
Linux 31/32 bit via IBM Java 8, which doesn't support RHEL 7 on Z31 as
per https://www.ibm.com/docs/en/sdk-java-technology/8?topic=installing-supported-environments


pshipton:
jenkins doc stage

pshipton:
jenkins doc stage

pshipton:
jenkins doc stage

PR #926, opened 2022-04-20
Document OPENJ9_JAVA_COMMAND_LINE and related options

### LINKED ISSUE #925: Update list of environment variables set by VM
https://github.com/eclipse-openj9/openj9/issues/947

**Overview:**
*The set of environment variables defined by the VM has changed.*

**Release target:**
*Eclipse OpenJ9 0.32.0*

*All versions*

*All platforms except z/OS*

Two new options are added:
* `-XX:+OpenJ9CommandLineEnv` (default, except on z/OS)
* `-XX:-OpenJ9CommandLineEnv` (default on z/OS)

keithc-ca:
Fixes #925.


keithc-ca:
jenkins doc stage

pshipton:
@doveye note IBM_JAVA_COMMAND_LINE is removed as it doesn't apply to OpenJ9, however it does apply to IBM Java 8.

docs/xxopenj9commandlineenv.md
```diff
@@ -0,0 +1,40 @@
...
+
+This option controls whether the VM captures the command line in the environment variable `OPENJ9_JAVA_COMMAND_LINE`.
+
+## Syntax
+
+        -XX:[+|-]OpenJ9CommandLineEnv
+
+| Setting                     | Effect  |
+|-----------------------------|---------|
+| `-XX:+OpenJ9CommandLineEnv` | Enable  |
```
pshipton, line 35: Other options put an indication beside the default option (`:fontawesome-solid-check:{: .yes aria-hidden="true"}<span class="sr-only">yes</span>`)

pshipton: I think we should do that and indicate the platform variation in brackets. Also we could mention that enabling it on z/OS may lead to undesirable results.

keithc-ca: Yes, but I wasn't sure how to display that the default varies by platform. How about *two* additional columns (for z/OS and other platforms) like this?

| Setting                     | Effect  | Default on z/OS | Default on Other Platforms |
|-----------------------------|---------|:-:|:-:|
| `-XX:+OpenJ9CommandLineEnv` | Enable  |   | x |
| `-XX:-OpenJ9CommandLineEnv` | Disable | x |   |

pshipton: It's fine with me. @doveye may also have an opinion.

keithc-ca:
jenkins doc stage

docs/xxopenj9commandlineenv.md
keithc-ca: I think using separate columns is clearer, and there's precedence for that approach: https://www.eclipse.org/openj9/docs/xxidletuningcompactonidle/.

keithc-ca:
jenkins doc stage

keithc-ca:
jenkins doc stage

pshipton (approved):

docs/version0.32.md
```diff
@@ -53,6 +54,10 @@ This option specifies whether generated hidden MethodHandle frames are displayed
 
 You can now use the `SharedClassStatistics` API to get the name, path, and directory of a shared classes cache. Depending on the operating system, you can also get the number of attached VMs for a non-persistent cache. This information is available through the following new methods: `cacheDir()`, `cacheName()`, `cachePath()`, and `numberAttached()`. For more information, see the API documentation.
 
+### New `-XX:[+|-]OpenJ9CommandLineEnv` option added
+
+This option specifies whether the command line should be captured in the environment variable `OPENJ9_JAVA_COMMAND_LINE`. For more information, see [`-XX:[+|-]OpenJ9CommandLineEnv`](xxopenj9commandlineenv.md).
```
doveye, line 59: Can we please change "should be" to "is" (use present tense where possible), or even better, use similar text to xxopenj9commandlineenv.md, ie "VM captures the command line" rather than "command line is captured" (active voice). Also, do we mean the generic "command line" or more specifically the "command that was used to start the VM"?
"

docs/xxopenj9commandlineenv.md
```diff
@@ -0,0 +1,40 @@
...
+* [1] https://www.gnu.org/software/classpath/license.html
+* [2] http://openjdk.java.net/legal/assembly-exception.html
+*
+* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0 WITH
+* Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
+-->
+
+# -XX:\[+|-\]OpenJ9CommandLineEnv
+
+This option controls whether the VM captures the command line in the environment variable `OPENJ9_JAVA_COMMAND_LINE`.
```
doveye, line 27: Same comment about generic "command line" as before. Also, I think we could usefully add the info from the env var topic after the first sentence plus we could use some clarification of what enabled/disable means for the behavior (it's pretty obvious but worth clarifying just in case). Eg: "When this option is enabled, the VM captures the command that was used to start the VM in the environment variable `OPENJ9_JAVA_COMMAND_LINE`. You can later use this environment variable to find the parameters that were set when the VM started."

docs/xxopenj9commandlineenv.md
```diff
@@ -0,0 +1,40 @@
...
+## Syntax
+
+        -XX:[+|-]OpenJ9CommandLineEnv
+
+| Setting                     | Effect  | Default on z/OS                                                                      | Default on other platforms                                                           |
+|-----------------------------|---------|:------------------------------------------------------------------------------------:|:------------------------------------------------------------------------------------:|
+| `-XX:+OpenJ9CommandLineEnv` | Enable  |                                                                                      | :fontawesome-solid-check:{: .yes aria-hidden="true"}<span class="sr-only">yes</span> |
+| `-XX:-OpenJ9CommandLineEnv` | Disable | :fontawesome-solid-check:{: .yes aria-hidden="true"}<span class="sr-only">yes</span> |                                                                                      |
+
+It is enabled by default on most platforms. On z/OS, however, it is disabled by default because it is believed to be the cause of at least one crashing scenario. For more information, see [Environment variables](env_var.md).
```
doveye, line 38: I don't think we need "It is enabled by default on most platforms." because that info is covered by the table (and more specifically).
Need to use "this option" instead of the ambiguous "it", z/OS needs a trademark (`&reg;`), and not sure about the problem text. Suggest "This option is currently disabled by default on z/OS&reg; because it might cause the VM to crash." (I used the word "currently" to reassure users that the problem will be fixed, which I assume is the case.)

Rather than "For more information, see <env var topic>", which doesn't really contain much more info (and no more with these review updates), I suggest a "See also" section similar to  https://www.eclipse.org/openj9/docs/xdisablejavadump/#see-also

Presumably there's no way for a user to get the command-line info on z/OS but if IBM_JAVA_COMMAND_LINE wasn't used by OpenJ9 before anyway, I guess there's no change so nothing to document. I assume that's also the reason we're not announcing that IBM_JAVA_COMMAND_LINE is deprecated, or mentioning that it's been removed.

doveye (commented):
Some changes suggested

docs/version0.32.md
keithc-ca: I will update as suggested.

Yes, it's the more generic command line. It applies not just to the `java` launcher, but to any application that uses JNI to invoke `JNI_CreateJavaVM()`.

docs/xxopenj9commandlineenv.md
pshipton: IBM_JAVA_COMMAND_LINE is not deprecated or removed, but it's only supported by IBM Java 8, and was never a feature of OpenJ9 builds.

keithc-ca:
Updated and rebased.

keithc-ca:
jenkins doc stage

Issue #927, opened 2022-04-25
Delete sentence that says JITServer is tech preview on zLinux


mpirvu:
**Issue or pull request number:**
*https://github.com/eclipse-openj9/openj9/pull/14424*

**Overview:**
*Since we fixed the remaining bugs for JITServer on zLinux we can declare full support on this platform as well. Thus we shouls remove the sentence that says: "Note: On Linux on IBM Z systems, this feature is a technical preview and should not be used in production environments."*

**Release target:**
*Eclipse OpenJ9 0.32.0*

**Applies to the following JDK versions:**
*8, 11 and 17 (where JITServer is supported)*

**Applies to the following platforms:**
*Linux Z*

doveye:
Doc updates are in as are updates to the release notes in the code repo, so this issue is done.

PR #928, opened 2022-04-25
Starting with release 0.32.0 JITServer is supported on zLinux as well

mpirvu:
Issue: #927

Signed-off-by: Marius Pirvu <mpirvu@ca.ibm.com>

### DIFF
--- a/docs/jitserver.md
+++ b/docs/jitserver.md
@@ -26,8 +26,6 @@
 
 **Linux&reg; on x86, Linux on IBM Power&reg; systems, and Linux on IBM Z&reg; systems (64-bit only)**
 
-:fontawesome-solid-pencil-alt:{: .note aria-hidden="true"} **Note:** On Linux on IBM Z systems, this feature is a technical preview and should not be used in production environments.
-
 JITServer technology decouples the JIT compiler from the VM and lets the JIT compiler run remotely in its own process. This mechanism prevents your Java&trade; application suffering possible negative effects due to CPU and memory consumption caused by JIT compilation.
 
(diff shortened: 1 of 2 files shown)


pshipton:
Pls update the release notes as well. 
https://www.eclipse.org/openj9/docs/version0.32/

mpirvu:
Created openj9 PR  https://github.com/eclipse-openj9/openj9/pull/14966 to update doc/release-notes/0.32/0.32.md

doveye:
Jenkins doc stage

doveye (approved):
You only need to trademark the first occurrence of a term, in normal text (terms don't need to be trademarked in titles). Plus these days it's just "IBM Z" rather than "IBM Z systems". But they are minor issues and this release has already gone out so I'm going to approve. We can make a note to tweak the text in future.

mpirvu:
> In the openj9 release notes it mentions JIT server (for Z) is only supported on 8, 11, 17. I don't see this limitation mentioned in the User Guide.

There is nothing known that prevents us to declare support for non LTS Java releases. It was a decision to limit exposure to just LTS Java. The same pattern was used with previous releases: the release notes only claimed support for Java8/11 and Java17 when it was released.
I can modify the User Guide to be more specific (i.e. say that only Java8/11/17 are supported)

pshipton:
> I can modify the User Guide to be more specific (i.e. say that only Java8/11/17 are supported)

Up to you, I was just pointing it out. I doubt we can expect user's to be aware of something announced in some old release notes.

Issue #932, opened 2022-05-10
AIX 11.0.16 requires the 16.1 XL C++ Runtime


pshipton:
**Issue or pull request number:**
https://github.com/eclipse-openj9/openj9/issues/14865

**Overview:**
Java 11 requires the 16.1 XL C++ Runtime installed.

**Release target:**
OpenJ9 0.33

**Applies to the following JDK versions:**
Java 11.0.16

**Applies to the following platforms:**
AIX

PR #933, opened 2022-05-10
AIX 11.0.16 requires the 16.1 XL C++ Runtime

pshipton:
Fixes https://github.com/eclipse-openj9/openj9-docs/issues/932

### DIFF
--- a/docs/version0.33.md
+++ b/docs/version0.33.md
@@ -1,0 +1,49 @@
+<!--
+* Copyright (c) 2017, 2022 IBM Corp. and others
+*
+* This program and the accompanying materials are made
+* available under the terms of the Eclipse Public License 2.0
+* which accompanies this distribution and is available at
+* https://www.eclipse.org/legal/epl-2.0/ or the Apache
+* License, Version 2.0 which accompanies this distribution and
+* is available at https://www.apache.org/licenses/LICENSE-2.0.
+*
+* This Source Code may also be made available under the
+* following Secondary Licenses when the conditions for such
+* availability set forth in the Eclipse Public License, v. 2.0
+* are satisfied: GNU General Public License, version 2 with
+* the GNU Classpath Exception [1] and GNU General Public
+* License, version 2 with the OpenJDK Assembly Exception [2].
+*
+* [1] https://www.gnu.org/software/classpath/license.html
+* [2] http://openjdk.java.net/legal/assembly-exception.html
+*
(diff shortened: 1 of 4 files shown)


pshipton:
jenkins doc stage

docs/version0.33.md
```diff
@@ -0,0 +1,49 @@
...
+### Binaries and supported environments
+
+OpenJ9 release 0.33.0 supports OpenJDK 8, 11, 17, and 18.
+
+To learn more about support for OpenJ9 releases, including OpenJDK levels and platform support, see [Supported environments](openj9_support.md).
+
+### ![Start of content that applies to Java 11](cr/java11.png) AIX requires the XL C++ Runtime
+
+AIX requires the [16.1 XL C++ Runtime](https://www.ibm.com/support/pages/fix-list-xl-cc-runtime-aix#161X) to accommodate a security update to the HarfBuzz text shaping library.
+
```
doveye, line 43: There's a mismatch between the link further up the topic and the title of this section, so the link doesn't work. But I think the words could use tweaking anyway to clarify that this is a change in level, to be more consistent with the text in the supported env topic, and to use a more official name for the runtime (which also clarifies that it's a third-party thing). Suggest something like:

> **Later level of XL C++ Runtime required on AIX**
> 
> AIX OpenJ9 builds now require version 16.1 of the IBM XL C++ Runtime for AIX (the previous required level was 13.1.3). This later level is required to accommodate a security update to the HarfBuzz text shaping library.

I also found this topic that mentions XL C++: https://eclipse-openj9.github.io/openj9-docs/configuring/#setting-resource-limits-aix-linux-and-macos. Looks like the Java 14 icon and text needs changing to 11.

doveye (changes requested):

pshipton:
jenkins doc stage

pshipton:
jenkins doc stage

pshipton:
jenkins doc stage

pshipton:
jenkins doc stage

pshipton:
@doveye this is ready for another review.

docs/version0.33.md
doveye: Ok, I saw the change in levels in the build env table but I guess that's different.

doveye (approved):
LGTM

doveye:
Had to prioritise other things yesterday. Checked again today though and the changes look good, thanks.

PR #941, opened 2022-05-30
Excessive GC

amicic:
The logic is reversed. Removing 'not'.

Signed-off-by: Aleksandar Micic <amicic@ca.ibm.com>


pshipton:
@amicic was the doc incorrect, or did the JVM change behavior?

amicic:
It was incorrect. Even back in IBM docs, where it was copied from.

doveye:
Thanks for the clarification Aleks. Merging...

Issue #966, opened 2022-07-15
Update `-XcompilationThreads` option documentation


AlexeyKhrabrov:
**Issue or pull request number:**
https://github.com/eclipse-openj9/openj9/pull/12914
https://github.com/eclipse-openj9/openj9/pull/12926

**Overview:**
The ranges of valid values for the `-XcompilationThreadsN` option have changed on both the JITServer and the JIT client. The server now supports up to 999 compilation threads, and the client (as well as the JVM not using JITServer) supports up to 16. The default number of threads is unchanged: 7 on the client on most platforms, 63 at the server. Increasing the maximum number of client threads can improve performance in high network latency settings since there can be more in-flight concurrent compilation requests. Increasing the number of threads at the server can improve performance if it has many CPU cores available and serves a large number of clients concurrently.

**Release target:**
Eclipse OpenJ9 0.33.0

**Applies to the following JDK versions:**
8 and later

**Applies to the following platforms:**
All JITServer platforms: Linux X|P|Z

AlexeyKhrabrov:
@mpirvu FYI

AlexeyKhrabrov:
In addition to updating the `-XcompilationThreads` documentation, I think it might be helpful to also describe when and how to increase the number of threads from the default in the "JITServer tuning" page.

Sreekala-Gopakumar:
@AlexeyKhrabrov, @mpirvu,  Could you please check the changes made and confirm. Thanks!

pshipton:
Confirmed the changes are in 0.33

Sreekala-Gopakumar:
@mpirvu - I have updated the documents based on your feedback. Please confirm the content is now fine.

doveye:
Thanks, the changes look good. As this item was already approved, I'll go ahead and merge the PR.

doveye:
Actually, can you squash the commits first please @Sreekala-Gopakumar

doveye:
git refused to squash, the PR was merged, and the changes are now appearing on the draft server, so I'll close this issue.

PR #967, opened 2022-07-18
Updated -XcompilationThreads option documentation

Sreekala-Gopakumar:
https://github.com/eclipse-openj9/openj9-docs/issues/966

Updated the ranges of valid values for the -XcompilationThreadsN option on both the JITServer and the JIT client.


docs/xcompilationthreads.md
```diff
@@ -31,7 +31,16 @@ Use this option to specify the number of compilation threads that are used by th
 
         -XcompilationThreads<n>
 
-: where `<n>` is the number of threads, in the range 1-7 inclusive. Any number outside this range is ignored. 
+: where `<n>` is the number of threads, in the range 1-7 inclusive. Any number outside this range is ignored.
```
mpirvu, line 34: We should probably not mention about JVM client here, since if the JVM is not attaching to any client the behavior is the same: for Java8 and above, the maximum number of usable compilation threads is 15 (not 16!).  The default number of threads is automatically determined by the JVM. We can talk about JITServer here, but I don't think we need to explain why somebody might want to use a larger number of threads.

mpirvu (commented):
I suggest the following text:
```
# -XcompilationThreads 
Use this option to specify the number of compilation threads that are used by the JIT compiler.
## Syntax

        -XcompilationThreads<n>

where `<n>` is the number of threads, in the range 1-7 inclusive. Any number outside this range is ignored.
Setting the compilation threads to zero does not prevent the JIT from working. Instead, if you do not want the JIT to work, use the -Xint option.

The default number of compilation threads is determined by internal JIT heuristics that take into consideration the number of CPUs the JVM is allowed to use.


![Start of content that applies to Java 8+](cr/java8plus.png) Starting with OpenJ9 release 0.33.0, the maximum number of compilation threads is 15. When operating in JITServer mode, the maximum number of compilation threads is 999, while the default value is 63.
```

docs/jitserver_tuning.md
```diff
@@ -120,6 +120,11 @@ A value greater than 0 for the `Compilation Queue Size` is a sign that the serve
 
     -XcompilationThreads<N> (default is 63)
 
-More detailed diagnostics can be obtained with the option `-Xjit:verbose={JITServer},verbose={compilePerformance}` which is typically used for debugging server behavior.
+![Start of content that applies to Java 8+](cr/java8plus.png) You can specify a larger number of threads with `-XcompilationThreads<N>` on the JITServer (999) and the JIT client (16). The default number of threads is kept at 63.
```
mpirvu, line 123: I suggest to compact all this information:

-XcompilationThreads<N> (default is 63, maximum is 999)

Increasing the maximum number of client threads (maximum is 15) can improve performance in high network latency settings since there can be more in-flight concurrent compilation requests. Increasing the number of threads at the server can improve performance if it has many CPU cores available and serves a large number of clients concurrently.

Sreekala-Gopakumar:
Jenkins doc stage

mpirvu (approved):
LGTM

docs/xcompilationthreads.md
doveye: @AlexeyKhrabrov the topic was updated in 2019 to say that the maximum was changed to 7 (see https://github.com/eclipse-openj9/openj9-docs/issues/435). Seems pretty clear from that issue that it was a maximum value? And now in 0.33 it's changed again to 15 - as I understand it, this change applies on a JITServer client and when using the JIT compiler "normally", ie without enabling the JITServer feature. Though things are confusing as the issue says that the change applies only to Linux.

Actually, looking at the code, looks like back in 2019 there was just MAX_CLIENT_USABLE_COMP_THREADS (7/63 for client/server) whereas now there's DEFAULT_CLIENT_USABLE_COMP_THREADS with the values from the old MAX_CLIENT_USABLE_COMP_THREADS (ie 7/63), and new values for MAX_CLIENT_USABLE_COMP_THREADS (15/999).

I've asked Sreekala to add words to clarify that the text later on about a max value of 999 applies to a JITServer server only (although I can't see my comments in github for some reason!) but perhaps we should add a table given that there are now 2 default values, depending on scenario.

mpirvu: Let me try to clarify the current state of affairs for Java8 and above (IBM SDK with Java7 does not apply).

The default number of compilation threads is determined by JIT heuristics. For JITServer the default is 63.
The maximum number of compilation threads is 15. For JITServer the maximum value is 999.

JITServer is only available on Linux platforms (X/P/Z).

doveye: Thanks @mpirvu. And what are the default & max values for JIT running without the JITServer feature enabled? Same as for JITServer client presumably, even on non-Linux platforms? eg if you're running JIT on AIX or on Linux in non-JITServer mode.

mpirvu: Indeed, it does not matter if we are running in client mode or vanilla OpenJ9, the values are the same.

Sreekala-Gopakumar:
Jenkins doc stage

Sreekala-Gopakumar:
Jenkins doc stage

docs/xcompilationthreads.md
```diff
@@ -31,9 +31,14 @@ Use this option to specify the number of compilation threads that are used by th
...
-: where `<n>` is the number of threads, in the range 1-7 inclusive. Any number outside this range is ignored. 
+: Where <n\> is the number of threads. The minimum value is 1. The default and maximum values depend on whether the VM is running with the JITServer feature enabled and if so, whether the VM is running as a JITServer server, as shown in the following table:
 
-: Setting the compilation threads to zero does not prevent the JIT from working. Instead, if you do not want the JIT to work, use the [`-Xint`](xint.md) option.
+| JIT compiler scenario| Default value | Maximum value |
+|----------------------|---------------|:-------------:|
+| JIT and JITServer client| Determined by internal JIT heuristics that consider the number of CPUs that the VM is allowed to use                         | 15            |
+| JITServer server        | 63         | 999           |
+
+: Any number outside this range is ignored. Setting the compilation threads to zero does not disable the JIT compiler. If you want to disable the JIT, use the [`-Xint`](xint.md) option.
```
doveye, line 42: The changes look good to me but we could still use a link to the JITServer topic. You could add a "See also" section, which is common to many command line option topics, eg https://www.eclipse.org/openj9/docs/xalwaysclassgc/.

docs/version0.33.md
```diff
@@ -50,6 +51,10 @@ This was already required for OpenJDK 17 and is now also required from 11.0.16 t
 
 Linux builds for all platforms now use gcc 10.3 instead of gcc 7.5. See the list of [build environments](openj9_support.md#build-environments).
 
+### The maximum number of compilation threads that are used by the JIT compiler increased
+
+The maximum number of compilation threads increased from 7 to 15 if the VM is running with the JITServer feature enabled (JIT and JITServer client). The maximum number of compilation threads on the JITServer server is increased to 999.
+
```
doveye, line 57: The text is not quite correct  - the 15 limit also applies when the JITServer feature is not enabled. We could use a link too. I suggest something like the following:

"The maximum number of JIT compilation threads is increased from 7 to 15. If the JITServer feature is enabled, the maximum number of JIT compilation threads is increased further on the server, to 999. You can set the number of compilation threads to be used by specifying the `-XcompilationThreads` command line option[link].”

docs/version0.33.md
```diff
@@ -50,6 +51,10 @@ This was already required for OpenJDK 17 and is now also required from 11.0.16 t
 
 Linux builds for all platforms now use gcc 10.3 instead of gcc 7.5. See the list of [build environments](openj9_support.md#build-environments).
 
+### The maximum number of compilation threads that are used by the JIT compiler increased
```
doveye, line 54: The title is odd because it switches from full English to brief English. I suggest something more like "The maximum number of JIT compilation threads is increased"

Sreekala-Gopakumar:
Jenkins doc stage

PR #979, opened 2022-08-17
Create release notes for 0.33.1

pshipton:


pshipton:
@Sreekala-Gopakumar pls review.

doveye:
Hi @pshipton. As 0.33.1 is for fixing a bug that doesn't affect the rest of the documentation (as far as I'm aware), I think we should modify the existing 0.33.0 What's New topic in the same way we did for [0.18.0](https://www.eclipse.org/openj9/docs/version0.18/), which was retitled for 0.18.1 but mentioned both 0.18.0 and 0.18.1 with a note to say that 0.18.1 contains additional bug fixes.

pshipton:
jenkins doc stage

pshipton:
jenkins doc stage

pshipton:
https://eclipse-openj9.github.io/openj9-docs-staging/979 

@doveye I've made the updates, pls have another look.

doveye:
Ok. I meant it would be a good learning experience for Sreekala to make the changes so she knows what to do in the future, but it's ok :).  I'll take a look...

mkdocs.yml
```diff
@@ -102,7 +102,7 @@ nav:
 
     - "Release notes" :
         - "Overview"                                                             : openj9_releases.md
-        - "Version 0.33.0"                                                       : version0.33.md
+        - "Version 0.33.0, .1"                                                   : version0.33.md
```
doveye, line 105: Hmm, this is not a very readable format and totally different from any previous release. For 15.1 and 18.1, we just changed the text to the point release. But if this situation is different and you really want the text to cover both releases, I suggest "Version 0.33.x"

(Perhaps in the future we should rename all the releases in the table of contents to the first 2 sets of digits, like the release notes in the code repo.)

docs/version0.33.md
```diff
@@ -35,10 +35,12 @@ The following new features and notable changes since version 0.30.0 are included
 
 ### Binaries and supported environments
 
-OpenJ9 release 0.33.0 supports OpenJDK 8, 11, 17, and 18.
+OpenJ9 release 0.33.0 and 0.33.1 supports OpenJDK 8, 11, 17, and 18.
 
 OpenJ9 Windows&reg; builds for OpenJDK 8 are now compiled with Microsoft&reg; Visual Studio 2017. The Visual Studio redistributable files included with the build are updated to match.
 
+:fontawesome-solid-pencil-alt:{: .note aria-hidden="true"} **Note:** Binaries that are labeled 0.33.1 include additional bug fixes. For more information, see the [release notes](https://github.com/eclipse-openj9/openj9/blob/master/doc/release-notes/0.33/0.33.1.md).
+
```
doveye, line 43: Re the links to the release notes here and at the end of the topic - I see that there's a new release note for 0.33.1 but the only difference is that the new features from 0.33.0 aren't in it. So it's not clear to a reader what these 2 releases are about. I suggest keeping just the 0.33 release note, in which we use "0.33.x" (if that's what we use in the doc table of contents, for consistency) when we're talking about both releases and specific release numbers when we're talking about 1 release, eg when linking to 0.33.0 vs 0.33.1 commit history. Then the bug that was fixed can be added in the "Known issues" table so 0.33.0 users are aware of it, with a comment to say it was fixed in 0.33.1. Then it will be clear to readers what the 2 releases are about and they only have to look at 1 file to find the info.

docs/version0.33.md
```diff
@@ -22,9 +22,9 @@
 * Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
 -->
 
-# What's new in version 0.33.0
+# What's new in version 0.33.0 and 0.33.1
```
doveye, line 25: See comment for mkdocs.yml - suggest either just "0.33.1" or "0.33.x"

docs/version0.33.md
```diff
@@ -22,9 +22,9 @@
...
 
-The following new features and notable changes since version 0.30.0 are included in this release:
+The following new features and notable changes since version 0.32.0 are included in this release:
```
doveye, line 28: Good spot! We spotted it recently but you got the fix in first :)

docs/version0.33.md
```diff
@@ -35,10 +35,12 @@ The following new features and notable changes since version 0.30.0 are included
 
 ### Binaries and supported environments
 
-OpenJ9 release 0.33.0 supports OpenJDK 8, 11, 17, and 18.
+OpenJ9 release 0.33.0 and 0.33.1 supports OpenJDK 8, 11, 17, and 18.
```
doveye, line 38: Should be "releases support" not "release supports"

doveye (changes requested):

mkdocs.yml
pshipton: I didn't want to add 0.33.x in case there is another release for which we want different release notes, but I suppose that is very unlikely. I thought about adding two entries, one for 0.33.1 and one for 0.33.0 which both refer to the same web page, but in end used 0.33.x.

docs/version0.33.md
pshipton: Used 0.33.x

docs/version0.33.md
pshipton: Used 0.33.x here as well.

docs/version0.33.md
pshipton: Updating OpenJ9 via https://github.com/eclipse-openj9/openj9/pull/15745

pshipton:
Pushed the update.

jenkins doc stage.

docs/version0.33.md
```diff
@@ -22,9 +22,9 @@
 * Classpath-exception-2.0 OR LicenseRef-GPL-2.0 WITH Assembly-exception
 -->
 
-# What's new in version 0.33.0
+# What's new in version 0.33.x
```
keithc-ca, line 25: This should say "versions".

doveye: I'm not sure I agree. It's a bit of a grey area but we already talk about, for example OpenSSL 1.1.x as singular rather than plural (eg in https://eclipse-openj9.github.io/openj9-docs-staging/979/introduction/#cryptographic-operations) and that would look odd if you made it plural. Similarly, we talk about Java version 8 as one release even though it's actually lots of sub-releases (we don't call it version 8.x but we could do).

keithc-ca: You suggested "OpenJ9 releases 0.33.x support" which considers "0.33.x" to be plural, so it would be inconsistent not to do so here as well.

doveye: I made that suggestion on this sentence: "OpenJ9 release 0.33.0 and 0.33.1 supports OpenJDK 8, 11, 17, and 18." (I think that sentence got lost in the force push?) Now the sentence reads "OpenJ9 releases 0.33.x support OpenJDK 8, 11, 17, and 18.", it sounds a bit odd and I would change it to "OpenJ9 release 0.33.x supports OpenJDK 8, 11, 17, and 18."

doveye (approved):

doveye:
@pshipton I've approved the review changes as I'm signing off now till Mon (I wasn't sure whether leaving the review open would prevent someone from merging the PR - if you don't want to wait till I'm back on Mon). There's still a bit of discussion and another change mentioned but otherwise I don't see anything else that needs changing, thx.

docs/version0.33.md
```diff
@@ -35,10 +35,12 @@ The following new features and notable changes since version 0.30.0 are included
 
 ### Binaries and supported environments
 
-OpenJ9 release 0.33.0 supports OpenJDK 8, 11, 17, and 18.
+OpenJ9 releases 0.33.x support OpenJDK 8, 11, 17, and 18.
```
keithc-ca, line 38: Based on [this comment](https://github.com/eclipse-openj9/openj9-docs/pull/979/files#r949389470), this should say "release 0.33.x supports".

pshipton:
Updated, hopefully I interpreted the comments correctly.

pshipton:
jenkins doc stage

keithc-ca (approved):
No need to re-stage given the nature of the recent changes.

Issue #1017, opened 2022-11-07
JSR 337 Maintenance Release 4: Java SE 8


pshipton:
**Issue or pull request number:**
https://github.com/eclipse-openj9/openj9/pull/15784
https://github.com/eclipse-openj9/openj9/pull/15828

**Overview:**
"JSR 337 Maintenance Release 4: Java SE 8" dictates some changes in Java 8 which are delivered with 0.35 / 8u352.
https://jcp.org/aboutJava/communityprocess/maintenance/jsr337/jsr337-mr4-changes.html

**Release target:**
0.35

**Applies to the following JDK versions:**
Java 8

**Applies to the following platforms:**
All

The doc update should first be delivered in the head stream as usual via Pull Request, and then we can also deliver it to the openj9-docs-0.35.0 branch and re-generate the 0.35 docs. I can help with that part after the Pull Request is merged.

Sreekala-Gopakumar:
@pshipton - After going through the changes specified and the existing OpenJ9 documents, I could identify only one change, Phantom reference, that needs to be updated in https://www.eclipse.org/openj9/docs/gc_overview/#weak-reference-processing. Is there anything else that needs an update in the existing document that you are aware of? 

Thanks!

PR #1018, opened 2022-11-08
JSR 337 Maintenance Release 4 - Phantom reference

Sreekala-Gopakumar:
https://github.com/eclipse-openj9/openj9-docs/issues/1017

Added information about the change in the GC behaviour in handling the phantom reference.

### DIFF
--- a/docs/version0.35.md
+++ b/docs/version0.35.md
@@ -35,6 +35,7 @@ The following new features and notable changes since version 0.33.1 are included
 - [New options added to encrypt the JITServer exported metrics](#new-options-added-to-encrypt-the-jitserver-exported-metrics)
 - ![Start of content that applies to Java 11 and later](cr/java11plus.png) [XL C++ Runtime 16.1.0.7 or later required on AIX](#xl-c-runtime-16107-or-later-required-on-aix)
 - [`-XX:[+|-]JITServerLocalSyncCompiles` enabled by default](#-xx-jitserverlocalsynccompiles-enabled-by-default)
+- ![Start of content that applies to Java 8](cr/java8.png) [Support for changes to the maintenance specifications of JSR 337 in Java 8](#support-for-changes-to-the-maintenance-specifications-of-jsr-337-in-java-8)
 
 ## Features and changes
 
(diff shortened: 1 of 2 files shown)


docs/version0.35.md
```diff
@@ -94,6 +95,12 @@ AIX OpenJ9 builds now require version 16.1.0.7 or later of the [IBM XL C++ Runti
 
 The `-XX:[+|-]JITServerLocalSyncCompiles` option is now enabled by default in most cases. For more information, see [`-XX:[+|-]JITServerLocalSyncCompiles`](xxjitserverlocalsynccompiles.md).
 
+### Unmarked phantom references are cleared immediately
```
pshipton, line 98: There is more that changed than just this topic. Rather than try to list them all, we could just refer to the maintenance spec for the list of changes.
i.e. The java.specification.maintenance.version system property, Reference.clone(), Reference.enqueue(), 
Runtime.runFinalizersOnExit() etc. are all relevant even though they didn't required changes elsewhere in the docs. Most of these required OpenJ9 changes to support them.

pshipton:
@dmitripivkine pls take a look.

pshipton:
The changes are specific to Java 8. Java 11+ already had the same behaviour as Java 8 does now.

dmitripivkine:
Yes, I agree instead of "Unmarked phantom references are cleared immediately" makes sense to say "Phantom references are handled as Weak in Java 8 similar as current Java 11+ behaviour". If we have Java 8 specific doc for phantom it should be edited (the same way as for Java 11+)

Sreekala-Gopakumar:
Jenkins doc stage

Sreekala-Gopakumar:
@dmitripivkine @pshipton - I have made the changes. Is it ok now? Thanks!

pshipton:
I want to see a more generic topic in What's New for the maintenance spec update. The topic "Garbage collector handles phantom references like weak references in Java 8" refers to one thing that changed in the maintenance spec, but there are multiple things that changed, and all of them affect jdk8. If we only call out 4 in the What's New, user's may not realize there were a number of other things that changed.

Contents on maintenance spec.
1	Version Identification
2	java.specification.maintenance.version
3	Reference.clone()
4	PhantomReference
5	JNI Weak Global References and NewGlobalRef
6	Reference.enqueue()
7	Runtime.runFinalizersOnExit()

Sreekala-Gopakumar:
Jenkins doc stage

Sreekala-Gopakumar:
Jenkins doc stage

Sreekala-Gopakumar:
Added the "Java 8" graphic. Thanks!

pshipton (approved):

pshipton:
I've pushed the change to the openj9-docs-0.35.0 branch and re-published the docs.

Issue #1037, opened 2023-01-10
Deprecate support for running xlinux on CentOS 6


pshipton:
**Issue or pull request number:**
https://github.com/eclipse-openj9/openj9/pull/16532

**Overview:**
Support for running jdk8, jdk11 on CentOS 6 is deprecated. OpenJ9 will no longer test xlinux jdk11 on CentOS 6 after 0.36, and may stop testing jdk8 in the future. When CRIU is enabled, it's not supported on CentOS 6 and testing can no longer run on this OS (without doing extra work to support this). Given CentOS 6 is end of life since November 30th, 2020, it's easier to deprecate support than to reconfigure the tests.

@tajila @llxia fyi

**Release target:**
OpenJ9 0.36 for deprecation, following releases will not be tested on CentOS 6.

**Applies to the following JDK versions:**
Java 8, 11

**Applies to the following platforms:**
xlinux

doveye:
@pshipton if a customer has an issue on CentOS 6, presumably we will not be invetigating/fixing it (if we don't have the environments to test), so this is more of a removal (ie, saying we no longer support CentOS 6) rather than a deprecation (ie warning of future removal of support)?

pshipton:
@doveye We still have the CentOS 6 environments at this time, they may be removed in the future.

PR #1039, opened 2023-01-17
Deprecate support for running JDK 8 and 11 on CentOS 6

Sreekala-Gopakumar:
https://github.com/eclipse-openj9/openj9-docs/issues/1037

Updated the related topics with the information that support for running JDK 8 and JDK 11 on CentOS 6 is deprecated.


Sreekala-Gopakumar:
Jenkins doc stage

Sreekala-Gopakumar:
@pshipton - I have updated the documents. Please check and confirm. Thanks!

pshipton (approved):

pshipton:
The CentOS 6 parts lgtm.

pshipton:
Actually, I wonder if the update should be listed under "What's new in version 0.36.0", otherwise it would be easy to miss the change.

mpirvu (approved):
LGTM

mpirvu:
@pshipton Just for my own education, why isn't CentOS 6 deprecated for Java 17 as well? The linked issue mentions that CentOS is out of service since 2020.

pshipton:
Java 17 never support CentOS 6.
https://www.eclipse.org/openj9/docs/openj9_support/

pshipton:
We build Java 17 on CentOS 7, it may not run at all on CentOS 6.

Sreekala-Gopakumar:
> Actually, I wonder if the update should be listed under "What's new in version 0.36.0", otherwise it would be easy to miss the change.

It is available under the Binaries and supported environments section.

Sreekala-Gopakumar:
> Actually, I wonder if the update should be listed under "What's new in version 0.36.0", otherwise it would be easy to miss the change.

It is available under the Binaries and supported environments section.

pshipton:
Yes, but it doesn't have it's own line in the list of changes, it seems it would be easy to miss for anyone not reading carefully.

doveye:
I don't think we should start confusing that topic by putting info about supported environments in the feature list as well as in the section about supported environments, especially as this item isn't even really a change, just a deprecation. We should remove the bit about "OpenJ9 Windows® builds for OpenJDK 8 are now compiled with Microsoft® Visual Studio 2017." though - that's been in since 0.33 so isn't a new change any more. That would clear the section up a bit and make the new content more obvious.

At the end of each OpenJDK section in the supported environments topic, eg https://www.eclipse.org/openj9/docs/openj9_support/#openjdk-8, we have this statement:

> When public support for an operating system version ends, OpenJ9 can no longer be supported on that level.

If CentOS 6 has been out of service since 2020 but we're still declaring support for it (albeit deprecated), this sentence would appear to be contradicted. Unless "public support" is supposed to mean more than just official support from the maintainers of the operating system? (eg some companies are offering extended support for CentOS 6 into 2024.)

pshipton:
I suppose CentOS 6 is an exception, since public support ended but we are still testing and declaring support for it. For some other OSes we removed them from the supported list when they went out of support.

pshipton:
@Sreekala-Gopakumar suggested adding a "note icon" to the deprecation so it will stand out and make it more likely to be noticed. This solution sounds good to me.

doveye:
Thanks for the extra info. So maybe we approach the note in the "Supported environments" topic a bit differently, eg something like:

> Support for OpenJ9 on CentOS 6.10 might be available even though CentOS 6 is no longer supported by the CentOS project. However, this support is deprecated and will be removed in a future release.

Which fits better with the existing statement about support being removed when an o/s reaches its own end of support. 

Re the Note - the section is for changes to binaries and supported environments and this will be the only change in the section so I don't think a Note is required to highlight it. With deprecation, all we're saying is "beware this support will go away at some point" which was pretty much the situation before anyway. It's in the release notes too, so multiple chances to spot it :) . Actually, there's more info in the release notes than in the What's New topic. I think it would be useful to add to What's new that OpenJ9 won't be tested with OpenJDK 11 on CentOS 6.10 after 0.36.

Sreekala-Gopakumar:
Jenkins doc stage

pshipton:
@Sreekala-Gopakumar pls fix the merge conflict.

Sreekala-Gopakumar:
Jenkins doc stage

pshipton (approved):

PR #1089, opened 2023-04-14
OpenJDK 19 for 0.37 is out of support

Sreekala-Gopakumar:
https://github.com/eclipse-openj9/openj9-docs/issues/1087

What's new topic for 0.37.0 release updated.

Closes #1087
Signed-off-by: Sreekala Gopakumar <sreekala.gopakumar@ibm.com>

PR #1088, opened 2023-04-14
0.38.0 release doesn't support Java 19

Sreekala-Gopakumar:
https://github.com/eclipse-openj9/openj9-docs/issues/1086

Updated the Supported environments topic with no support for Java 19 in 0.38.0 release

Closes #1086
Signed-off-by: Sreekala Gopakumar <sreekala.gopakumar@ibm.com>

PR #1079, opened 2023-04-05
JDK8 Windows compiles with VS2019

Sreekala-Gopakumar:
https://github.com/eclipse-openj9/openj9-docs/issues/1073

Updated the topics related to JDK8 build change.

Closes: #1073
Signed-off-by: Sreekala Gopakumar <sreekala.gopakumar@ibm.com>

PR #1090, opened 2023-04-14
(0.37) Update release notes and support statement

pshipton:
Cherry pick of https://github.com/eclipse-openj9/openj9-docs/pull/1088 and https://github.com/eclipse-openj9/openj9-docs/pull/1089 for 0.37


keithc-ca:
This should also include #1079 as the 0.37.0 release branch includes eclipse-openj9/openj9#16706.

pshipton:
The 0.37.0 release only supports OpenJDK 19. Since it doesn't support 8, we haven't tested if 8 can even build, it makes more sense to include the compiler for 8 in the 0.38.0 release notes.

keithc-ca (approved):
It seems a little strange to include the tables of OpenJDK 8, 11 and 17 in a version that only relates to OpenJDK 19, but I suppose this isn't the first time we've done something like that and it probably not really worth the effort to address it.

PR #1093, opened 2023-05-03
Reference openjdk.org instead of openjdk.java.net

keithc-ca:
See eclipse-openj9/openj9#16057.


doveye (approved):
LGTM

doveye:
Thanks for the updates Keith 🙂

PR #1094, opened 2023-05-03
Document restrictions for the jvmtiVerboseGCSubscriber event

gacholio:
Document the fact that these events run under the same restrictions as the GarbageCollectionStart/End events.

### DIFF
--- a/docs/interface_jvmti.md
+++ b/docs/interface_jvmti.md
@@ -858,6 +858,8 @@ Macro declaration in the `ibmjvmti.h` file: `COM_IBM_REGISTER_VERBOSEGC_SUBSCRIB
 
     This function is called with each record of verbose logging data produced by the VM.
 
+    This function runs under the same restrictions as the `GarbageCollectionStart` and `GarbageCollectionFinish` events in the standard JVMTI specification. For more information about these events, see the [JVMTI Event Index](https://docs.oracle.com/en/java/javase/20/docs/specs/jvmti.html#EventIndex) for your OpenJDK version.
+
     The verbose logging record supplied to the subscriber function is valid only for the duration of the function. If the subscriber wants to save the data, the data must be copied elsewhere.
 
     If the subscriber function returns an error, the alarm function is called, and the subscription is deregistered.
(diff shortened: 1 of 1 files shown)


gacholio:
Note that these restrictions have always been in place but not properly documented.

amicic:
looks good to me
@pshipton

Sreekala-Gopakumar (changes requested):
1. Change "This function executes..." to "This function runs..."
2. We have mentioned restrictions same as `GarbageCollectionStart` and `GarbageCollectionFinish` but in our existing documents these events or the restrictions are not mentioned. We should add either the content related to that or add a link to any document where these events and the restrictions are mentioned.

gacholio:
The start and finish events are from the Oracle JVMTI specification. There is no link I know about that isn't specific to a particular JDK version. Do we want to link to Oracle documentation at all?

gacholio:
Comment 1 addressed.

gacholio:
Here is a version of the specification for the events: https://docs.oracle.com/en/java/javase/20/docs/specs/jvmti.html#GarbageCollectionStart

gacholio:
The relevant portion is:
```
This event is sent while the VM is still stopped, thus the event handler must not use JNI functions and must not use JVM TI functions except those which specifically allow such use (see the raw monitor, memory management, and environment local storage functions). 
```
It would be difficult to paraphrase this without excluding critical information.

doveye:
Linking to Oracle docs is fine, and such links already exist in this topic (in fact I think the doc you found is better than the existing link!) This text is in a section that is about the OpenJ9 extensions so we do need to clarify that it's talking about something that isn't OpenJ9-specific. I suggest something like:

> This function runs under the same restrictions as the `GarbageCollectionStart` and `GarbageCollectionFinish` events in the standard JVMTI specification. For more information about these events, see the [JVMTI Event Index](https://docs.oracle.com/en/java/javase/20/docs/specs/jvmti.html#EventIndex) for your OpenJDK version.

gacholio:
Thank you, I have made the suggested change.

doveye:
@gacholio Sorry, but can you please use the Markdown syntax for the link (technically, I think HTML will work but it's unnecessary) and put ` around the event names to highlight them in the normal text. ie 

```
This function runs under the same restrictions as the `GarbageCollectionStart` and `GarbageCollectionFinish` events in the standard JVMTI specification. For more information about these events, see the [JVMTI Event Index](https://docs.oracle.com/en/java/javase/20/docs/specs/jvmti.html#EventIndex) for your OpenJDK version.
```

Thanks!

gacholio:
Done.

doveye:
Thanks 🙂 . All merged now.

PR #1091, opened 2023-04-19
CRIU Support Documentation

Sreekala-Gopakumar:
https://github.com/eclipse-openj9/openj9-docs/issues/1053

Added content related to CRIU support

Closes #1053
Signed-off-by: Sreekala Gopakumar <sreekala.gopakumar@ibm.com>

PR #1114, opened 2023-06-21
Document new option -XX:codecachetotalMaxRAMPercentage

### LINKED ISSUE #1108: Document new option -XX:codecachetotalMaxRAMPercentage=N
https://github.com/eclipse-openj9/openj9/pull/17425

**Overview:**
The option `-XX:codecachetotalMaxRAMPercentage=N` sets a maximum limit for the total memory allocated by the JVM for code caches, limit expressed as a percentage of the available physical memory.
The value `N` needs to be a number between 1 and 100 (fractional numbers are allowed).
For example, with the option  `-XX:codecachetotalMaxRAMPercentage=30` the JVM is not allowed to use more than 30% of the available physical memory for its internal code caches.
Default: The default value for this option is 25, which means 25% of the available physical memory.

Note that this option does not indicate a prescribed total code cache value, but rather sets a maximum limit. In environments with low memory availability this option can shrink the default total code cache value (the default is set to 256 MB for 64-bit systems and 64 MB for 31/32-bit  systems), while in environments with plenty of memory it will have no effect. If you want to increase the total code cache value use the `-XX:codecachetotal=size` or the `-Xcodecachetotal=size` options.

Note that you can specify a prescribed total code cache value with  `-XX:codecachetotal=size` or  `-Xcodecachetotal=size` options. When that happens, the limit set with  `-XX:codecachetotalMaxRAMPercentage=N` is ignored.

**Release target:**
Eclipse OpenJ9 0.40.0

8 and later

All platforms 

- XX:codecachetotalMaxRAMPercentage=N
- Values are set:
	- Units for setting: %
	- Range: 1-100, can be fractional values.
	- Default value: 25
- Purpose: Set the maximum limit for the JVM's total code cache memory, limit expressed as a percentage of the physical memory available to the JVM. The intent is to prevent the JVM from using too much memory for its code cache in environments with low physical memory availability..
- Restrictions: If -XX:codecachetotal=<size> or -Xcodecachetotal<size> is specified, then -XX:codecachetotalMaxRAMPercentage=N is ignored. 
            - The specified percentage of the physical available memory is used only if the calculated total cache value is less than the default total code cache value set for the platform.

Sreekala-Gopakumar:
https://github.com/eclipse-openj9/openj9-docs/issues/1108

Created new topic for the new option and updated other related topics including the What's new in version 0.40.0 topic.

Closes #1108

### DIFF
--- a/docs/version0.40.md
+++ b/docs/version0.40.md
@@ -1,0 +1,49 @@
+<!--
+* Copyright (c) 2017, 2023 IBM Corp. and others
+*
+* This program and the accompanying materials are made
+* available under the terms of the Eclipse Public License 2.0
+* which accompanies this distribution and is available at
+* https://www.eclipse.org/legal/epl-2.0/ or the Apache
+* License, Version 2.0 which accompanies this distribution and
+* is available at https://www.apache.org/licenses/LICENSE-2.0.
+*
+* This Source Code may also be made available under the
+* following Secondary Licenses when the conditions for such
+* availability set forth in the Eclipse Public License, v. 2.0
+* are satisfied: GNU General Public License, version 2 with
+* the GNU Classpath Exception [1] and GNU General Public
+* License, version 2 with the OpenJDK Assembly Exception [2].
+*
+* [1] https://www.gnu.org/software/classpath/license.html
+* [2] https://openjdk.org/legal/assembly-exception.html
+*
+* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0 OR GPL-2.0-only WITH OpenJDK-assembly-exception-1.0
+-->
+
+# What's new in version 0.40.0
+
+The following new features and notable changes since version 0.39.0 are included in this release:
+
+- [New binaries and changes to supported environments](#binaries-and-supported-environments)
+- [New `-XX:codecachetotalMaxRAMPercentage` option added](#new-xxcodecachetotalmaxrampercentage-option-added)
+
+## Features and changes
+
+### Binaries and supported environments
+
+Eclipse OpenJ9&trade; release 0.40.0 supports OpenJDK 8, 11, 17, and 20.
+
+To learn more about support for OpenJ9 releases, including OpenJDK levels and platform support, see [Supported environments](openj9_support.md).
+
+### New `-XX:codecachetotalMaxRAMPercentage` option added
+
+In environments with low physical memory availability, the VM might use too much memory for JIT code caches, leaving little memory for critical operations. With the `-XX:codecachetotalMaxRAMPercentage` option, you can set an upper limit for the total code cache size, where the upper limit is specified as a percentage of the physical memory the VM process is allowed to use.
+
+For more information, see [`-XX:codecachetotalMaxRAMPercentage`](xxcodecachetotalmaxrampercentage.md).
+
+## Known problems and full release information
+
+To see known problems and a complete list of changes between Eclipse OpenJ9 v0.39.0 and v0.40.0 releases, see the [Release notes](https://github.com/eclipse-openj9/openj9/blob/master/doc/release-notes/0.40/0.40.md).
+
+<!-- ==== END OF TOPIC ==== version0.40.md ==== -->
--- a/docs/xcodecachetotal.md
+++ b/docs/xcodecachetotal.md
@@ -36,7 +36,7 @@ See [Using -X command-line options](x_jvm_commands.md) for more information abou
 
 ## Explanation
 
-By default, the total size of the JIT code cache is 256 MB for a 64-bit VM and 64 MB for a 31/32-bit VM. Long-running, complex, server-type applications can fill the JIT code cache, which can cause performance problems because not all of the important methods can be JIT-compiled. Use the `-Xcodecachetotal` option to increase or decrease the maximum code cache size to a setting that suits your application. The minimum size of the code cache is restricted to 2 MB.
+By default, the total JIT code cache size is 256 MB for a 64-bit VM and 64 MB for a 31/32-bit VM or 25% of the physical memory available to the VM process, whichever is lesser. Long-running, complex, server-type applications can fill the JIT code cache, which can cause performance problems because not all of the important methods can be JIT-compiled. Use the `-Xcodecachetotal` option to increase or decrease the maximum code cache size to a setting that suits your application. The minimum size of the code cache is restricted to 2 MB.
 
 The value that you specify is rounded up to a multiple of the code cache block size, as specified by the [-Xcodecache](xcodecache.md) option. If you specify a value for the `-Xcodecachetotal` option that is smaller than the default setting, that value is ignored.
 
@@ -46,6 +46,7 @@ The maximum size limits, for both the JIT code and data caches, that are in use 
 
 ## See also
 
+- [`-XX:codecachetotalMaxRAMPercentage`](xxcodecachetotalmaxrampercentage.md)
 - [-Xjit](xjit.md)
 
 
--- a/docs/xxcodecachetotalmaxrampercentage.md
+++ b/docs/xxcodecachetotalmaxrampercentage.md
@@ -1,0 +1,61 @@
+<!--
+* Copyright (c) 2017, 2022 IBM Corp. and others
+*
+* This program and the accompanying materials are made
+* available under the terms of the Eclipse Public License 2.0
+* which accompanies this distribution and is available at
+* https://www.eclipse.org/legal/epl-2.0/ or the Apache
+* License, Version 2.0 which accompanies this distribution and
+* is available at https://www.apache.org/licenses/LICENSE-2.0.
+*
+* This Source Code may also be made available under the
+* following Secondary Licenses when the conditions for such
+* availability set forth in the Eclipse Public License, v. 2.0
+* are satisfied: GNU General Public License, version 2 with
+* the GNU Classpath Exception [1] and GNU General Public
+* License, version 2 with the OpenJDK Assembly Exception [2].
+*
+* [1] https://www.gnu.org/software/classpath/license.html
+* [2] http://openjdk.java.net/legal/assembly-exception.html
+*
+* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0 OR GPL-2.0-only WITH OpenJDK-assembly-exception-1.0
+-->
+
+# -XX:codecachetotalMaxRAMPercentage
+
+This option sets the maximum limit for the total JIT code cache size based on the amount of physical memory available to the VM.
+
+## Syntax
+
+        -XX:codecachetotalMaxRAMPercentage=<number>
+
+| Setting                 | Value      | Default                     |
+|-------------------------|------------|:---------------------------:|
+| `<number>`    | Between 1 and 100 (fractional numbers are allowed) |  25                       |
+
+Where `<number>` is the maximum limit that is expressed as a percentage of the available physical memory.
+
+For example, if you specify `-XX:codecachetotalMaxRAMPercentage=30`, the VM is not allowed to use more than 30% of the available physical memory for the internal JIT code caches.
+
+## Explanation
+
+The default total size of the JIT code cache is computed as the minimum of the following two limits:
+
+- A platform-based limit that is expressed as an absolute memory value. This limit is set at 256 MB for 64-bit systems and at 64 MB for 31/32-bit systems.
+- A percentage of the amount of available physical memory the VM process is allowed to use. By default, the VM is not allowed to use more than 25% of the available physical memory for its code caches.
+
+In memory constrained environments, the percentage limit is relevant because the code cache size is then based on the available physical memory. The percentage limit prevents the VM from using too much memory for its code caches and thus, leaving too little memory for other VM or JIT data structures.
+
+For example, on a 64-bit system, the platform-specific code cache limit is 256 MB. If the VM is constrained to less than 1024 MB, say 512 MB, then the code cache limit becomes 128 MB (25% of 512 MB) because the percentage limit is less than the platform limit.
+
+To fine-tune the code cache size limit for your specific application as a percentage of the available physical memory, you can use the `-XX:codecachetotalMaxRAMPercentage` option.
+
+As an alternative, you can use the [`-XX:codecachetotal`](xxcodecachetotal.md) or the [`-Xcodecachetotal`](xcodecachetotal.md) options to set the code cache size limit as an absolute value (platform-based limit). The absolute value that is specified with these options takes precedence over the percentage that is specified with the `-XX:codecachetotalMaxRAMPercentage` option.
+
+:fontawesome-solid-triangle-exclamation:{: .warn aria-hidden="true"} **Restrictions:**
+
+- If the `-XX:codecachetotal` or `-Xcodecachetotal` option is specified, then the `-XX:codecachetotalMaxRAMPercentage` option is ignored.
+- The percentage of the available physical memory that is specified in the `-XX:codecachetotalMaxRAMPercentage` option is used only if the total cache value thus calculated is less than the default total code cache value set for the platform (minimum of the two limits).
+
+
+<!-- ==== END OF TOPIC ==== xxcodecachetotalmaxrampercentage.md ==== -->
--- a/docs/xxcodecachetotal.md
+++ b/docs/xxcodecachetotal.md
@@ -31,6 +31,8 @@ This option is an alias for the [`-Xcodecachetotal`](xcodecachetotal.md) option.
 
 : See [Using -X command-line options](x_jvm_commands.md) for more information about specifying the `<size>` parameter.
 
+## See also
 
+- [`-XX:codecachetotalMaxRAMPercentage`](xxcodecachetotalmaxrampercentage.md)
 
 <!-- ==== END OF TOPIC ==== xxcodecachetotal.md ==== -->
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -101,6 +101,7 @@ nav:
 
     - "Release notes" :
         - "Overview"                                                             : openj9_releases.md
+        - "Version 0.40.0"                                                       : version0.40.md
         - "Version 0.39.0"                                                       : version0.39.md
         - "Version 0.38.0"                                                       : version0.38.md
         - "Version 0.37.0"                                                       : version0.37.md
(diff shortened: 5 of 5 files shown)


Sreekala-Gopakumar:
Jenkins doc stage

Sreekala-Gopakumar:
@SajinaKandy, @mpirvu - Could you please let me know if the updated content is ok. If the update is approved, this can be merged. Thanks!

SajinaKandy:
@Sreekala-Gopakumar Thanks for the doc update. It looks good to me.

Sreekala-Gopakumar:
@mpirvu - If there are no further review comments, could you please approve and merge the changes. Thanks!

docs/xxcodecachetotalmaxrampercentage.md
```diff
@@ -0,0 +1,52 @@
...
+|-------------------------|------------|:---------------------------:|
+| `<number>`    | Between 1 and 100 (fractional numbers are allowed) |  25                       |
+
+Where `<number>` is the maximum limit expressed as a percentage of the available physical memory.
+
+For example, if you specify `-XX:codecachetotalMaxRAMPercentage=30`, the VM is not allowed to use more than 30% of the available physical memory for the internal code caches.
+
+## Explanation
+
+In environments with low physical memory availability, the VM might use too much memory for code caches, leaving little memory for the VM or JIT data structures. With the `-XX:codecachetotalMaxRAMPercentage` option, you can reduce the default total code cache value (the default is set to 256 MB for 64-bit systems and 64 MB for 31 or 32-bit systems). In environments with plenty of memory, the `-XX:codecachetotalMaxRAMPercentage` setting is ignored.
```
mpirvu, line 42: How about this:

By default, the total size of the JIT code cache is 256 MB for a 64-bit VM and 64 MB for a 31/32-bit VM.
Despite this limit, in environments with low physical memory availability, the VM might use too much memory for code caches, leaving little memory for the VM or JIT data structures. With the `-XX:codecachetotalMaxRAMPercentage` option, you can reduce (but not increase) the default total size of the JIT code cache. 
If you want to increase the total code cache size, you can use the [`-XX:codecachetotal`](xxcodecachetotal.md) or the [`-Xcodecachetotal`](xcodecachetotal.md) options.

docs/version0.40.md
```diff
@@ -0,0 +1,49 @@
...
+
+### Binaries and supported environments
+
+Eclipse OpenJ9&trade; release 0.40.0 supports OpenJDK 8, 11, 17, and 20.
+
+To learn more about support for OpenJ9 releases, including OpenJDK levels and platform support, see [Supported environments](openj9_support.md).
+
+### New `-XX:codecachetotalMaxRAMPercentage` option added
+
+In environments with low physical memory availability, the VM might use too much memory for code caches, leaving little memory for critical operations. With the `-XX:codecachetotalMaxRAMPercentage` option, you can set a maximum limit for the total memory that the VM allocates for code caches.
```
mpirvu, line 41: Ho about this:

In environments with low physical memory availability, the VM might use too much memory for JIT code caches, leaving little memory for critical operations. With the `-XX:codecachetotalMaxRAMPercentage` option, you can set an upper bound for the total code cache size, upper bound specified as a percentage of the physical memory the VM process is allowed to use.

mpirvu:
I am having second thoughts: Overall, I am not sure that an user is able to tell the total size of the code cache from the description we have for `-XX:codecachetotalMaxRAMPercentage` and `-XX:codecachetotal`.
Note that there is a change in behavior when running in memory constrained environments: the total size of the JIT code cache is computed as the minimum of two limits:
1. The "default" limit on the platform, which is 256 MB on 64-bit systems and 64 MB on 31/32 bit systems
2. 25% of the amount of physical memory the JVM process is allowed to use

Sreekala-Gopakumar:
> I am having second thoughts: Overall, I am not sure that an user is able to tell the total size of the code cache from the description we have for `-XX:codecachetotalMaxRAMPercentage` and `-XX:codecachetotal`. Note that there is a change in behavior when running in memory constrained environments: the total size of the JIT code cache is computed as the minimum of two limits:
> 
> 1. The "default" limit on the platform, which is 256 MB on 64-bit systems and 64 MB on 31/32 bit systems
> 2. 25% of the amount of physical memory the JVM process is allowed to use

Don't we need a third point related to the percentage specified in the `-XX:codecachetotalMaxRAMPercentage` option? Minimum of the three limits?

Sreekala-Gopakumar:
> > Don't we need a third point related to the percentage specified in the -XX:codecachetotalMaxRAMPercentage option? Minimum of the three limits?
> 
> No, because the 25% from the second point above comes from the default value of `-XX:codecachetotalMaxRAMPercentage`. This new option was introduced to be able to change that 25%. Your question is a signal that the documentation, as it stands, is not clear to a person not familiar with the implementation.

@mpirvu - But what if we specify `-XX:codecachetotalMaxRAMPercentage=30`? Won't the total size of the JIT code cache be computed as 30% of the amount of physical memory i.e. it will ignore both the "default" limit on the platform as well as the default value of `-XX:codecachetotalMaxRAMPercentage`?

Sreekala-Gopakumar:
> > But what if we specify -XX:codecachetotalMaxRAMPercentage=30? Won't the total size of the JIT code cache be computed as 30% of the amount of physical memory i.e. it will ignore both the "default" limit on the platform as well as the default value of -XX:codecachetotalMaxRAMPercentage?
> 
> You are absolutely right. There are two code cache limits one expressed as an absolute memory value and the other one expressed as a percentage of physical memory. The JVM uses the minimum of these 2 limits. There are also options to control the two limits mentioned above: -XX:codecachetotal for the first limit and -XX:codecachetotalMaxRAMPercentage for the second limit

@mpirvu - Again a doubt. Following is mentioned in the issue

> Note that you can specify a prescribed total code cache value with  -XX:codecachetotal=size or  -Xcodecachetotal=size options. When that happens, the limit set with  -XX:codecachetotalMaxRAMPercentage=N is ignored.

So, if absolute memory value is changed by `-XX:codecachetotal` then won't the -XX:codecachetotalMaxRAMPercentage=N be ignored? The VM won't check the minimum of those 2 limits then; it will consider only the `-XX:codecachetotal` value, right?

mpirvu:
> So, if absolute memory value is changed by -XX:codecachetotal then won't the -XX:codecachetotalMaxRAMPercentage=N be ignored?

Yes, it will be ignored. We wanted to avoid confusion by specifying limits with 2 different options. `-XX:codecachetotal` is the older and more established option for controlling the size of the code caches and this one take precedence. It the user wants to use a particular value for code cache size (with -XX:codecachetotal ), it's annoying to force him/her to add the other option as well (-XX:codecachetotalMaxRAMPercentage=N). 

Here's some text that hopefully is clearer, though I am sure it can be improved upon:

The default total size of the JIT code cache is computed as the minimum of two limits:
1. A platform based limit expressed as an absolute memory value. This limit is set at 256 MB for 64-bit systems  and at 64-MB for 31/32-bit systems
2. A percentage of the amount of physical memory the JVM process is allowed to use. By default, the JVM is not allowed to use more than 25% of physical memory for its code caches

Both limits mentioned above can be changed through command line options:
- To change the limit expressed as an absolute value, use the  `-XX:codecachetotal` option (or its alias `-Xcodecachetotal`)
- To change the limit expressed as a percentage of the physical memory, use the `-XX:codecachetotalMaxRAMPercentage` option.

If both options are specified on the command line, only the  `-XX:codecachetotal` will take effect and `-XX:codecachetotalMaxRAMPercentage` will be ignored.

Sreekala-Gopakumar:
@mpirvu - A couple of questions more:

1. In the following sentence, what does the "the calculated total cache value" refer to? The default percentage of the available physical memory i.e. 25% of the available physical memory that the JVM process is allowed to use?
> The percentage of the physical available memory that is specified in the `-XX:codecachetotalMaxRAMPercentage` option is used only if the calculated total cache value is less than the default total code cache value set for the platform.

Also, is the "default total code cache value set for the platform" referring to the minimum of the two limits or the "default" limit on the platform, which is 256 MB on 64-bit systems and 64 MB on 31/32 bit systems?

2. In the scenario, where the VM is running in memory constrained environments, the chances are that the 25% of the amount of physical memory is less than the platform defaults and therefore will be used to compute the total size of the JIT code cache, right? And if this 25% is still too much, is that when we use the `-XX:codecachetotalMaxRAMPercentage` option to reduce the total size of the JIT code cache further?

So does that mean that we will always specify a percentage that is less than 25%? Is there any scenario where we might specify a percentage more than 25%, even 90%? If not then why is the range set as 1%-100%? I am trying to reconcile the concept of the minimum of the 2 limits and the use of the `-XX:codecachetotalMaxRAMPercentage` option.

> Note that there is a change in behavior when running in memory constrained environments: the total size of the JIT code cache is computed as the minimum of two limits:

- The "default" limit on the platform, which is 256 MB on 64-bit systems and 64 MB on 31/32 bit systems
- 25% of the amount of physical memory the JVM process is allowed to use

mpirvu:
> what does the "the calculated total cache value" refer to?

Based on the value of `-XX:codecachetotalMaxRAMPercentage` and on the amount of physical memory, the JVM calculates a value for the total size of the code cache. If this value is greater than the 256 MB (or 64 MB on 32-bit) platform limit, then the previously calculated value is ignored. This is because the JVM takes the minimum of the two limits.

> is "default total code cache value set for the platform" referring to the minimum of the two limits or the "default" limit on the platform

It refers to the "default" limit which is 256/64 MB.

> In the scenario, where the VM is running in memory constrained environments, the chances are that the 25% of the amount of physical memory is less than the platform defaults and therefore will be used to compute the total size of the JIT code cache, right? And if this 25% is still too much, is that when we use the -XX:codecachetotalMaxRAMPercentage option to reduce the total size of the JIT code cache further?

Considering a 64-bit platform as an example, the platform specific code cache limit is 256 MB. If the JVM is constrained to less than 1024 MB, then 25% of physical memory is less than 256 MB, so the percentual limit will apply and the 256 MB limit will be ignored (again we take the minimum of the two). If the user finds that 25% is too much or too little, he can tweak that percentage value to something appropriate for his system.

> Is there any scenario where we might specify a percentage more than 25%, even 90%?

The idea behind the 25% value was that the Java heap is allowed to use 50% of the memory and we need at least 25% of the memory for other VM/GC/JIT data structures. This is dependent on the application. If the application uses little Java heap but compiles many methods and needs a larger code cache, then the user might want to use a percentage value larger than 25%. Is 90% a likely value? Probably not, but if the user decides to increase that value, probably he knows what he is doing, so why restrict him needlessly.

mpirvu:
> I am trying to reconcile the concept of the minimum of the 2 limits and the use of the -XX:codecachetotalMaxRAMPercentage option.

Starting with release 0.40.0 the 25% percentage limit is going to be in effect. For small containers (smaller than 1024 MB) this is the limit that dictates the total size of the code cache. If the code cache size computed based on the 25% limit is not appropriate for the application that is running, then the user may need to change it. This is where `-XX:codecachetotalMaxRAMPercentage` comes into picture. Yes, it's possible for the user to use the old option as well (-XX:codecachetotal), but this is not as flexible as `-XX:codecachetotalMaxRAMPercentage`. If the user decides to slightly adjust the memory for a container, the code cache size can scale up/down proportionally

Sreekala-Gopakumar:
Thanks a lot for the clarification!

docs/xxcodecachetotalmaxrampercentage.md
Sreekala-Gopakumar: > How about this:
> 
> By default, the total size of the JIT code cache is 256 MB for a 64-bit VM and 64 MB for a 31/32-bit VM. Despite this limit, in environments with low physical memory availability, the VM might use too much memory for code caches, leaving little memory for the VM or JIT data structures. With the `-XX:codecachetotalMaxRAMPercentage` option, you can reduce (but not increase) the default total size of the JIT code cache. If you want to increase the total code cache size, you can use the [`-XX:codecachetotal`](xxcodecachetotal.md) or the [`-Xcodecachetotal`](xcodecachetotal.md) options.

@mpirvu - One last question - :) In this you have said that "With the `-XX:codecachetotalMaxRAMPercentage` option, you can reduce (**but not increase**) the default total size of the JIT code cache." But in this  https://github.com/eclipse-openj9/openj9-docs/pull/1114#issuecomment-1609990807, you have mentioned that "If the user finds that 25% is too much or too little, he can tweak that percentage value to something appropriate for his system." and that "If the application uses little Java heap but compiles many methods and needs a larger code cache, then the user might want to use a percentage value larger than 25%." So doesn't that mean that we can actually increase the default total size of the JIT code cache.?

Sreekala-Gopakumar:
Jenkins doc stage

Sreekala-Gopakumar:
@mpirvu - I have updated the topics. Please check and confirm. Thanks!

Sreekala-Gopakumar:
Jenkins doc stage

Sreekala-Gopakumar:
@mpirvu - I have updated the topics. Please check and confirm. Thanks!

docs/xxcodecachetotalmaxrampercentage.md
```diff
@@ -0,0 +1,59 @@
...
+- A percentage of the amount of available physical memory the VM process is allowed to use. By default, the VM is not allowed to use more than 25% of the available physical memory for its code caches.
+
+For example, in a 64-bit platform, the platform specific code cache limit is 256 MB. If the VM is constrained to less than 1024 MB (say 900 MB), then 25% of physical memory is less than 256 MB (the platform-based limit). Therefore, the default total size is calculated as a percentage of the amount of available physical memory (225 - minimum of the two limits).
+
+In environments with low physical memory availability, the VM might use too much of the available memory for code caches, leaving little memory for the VM or JIT data structures. With the `-XX:codecachetotalMaxRAMPercentage` option, you can reduce the default total size of the JIT code cache.
+
+If you want to increase the total code cache size as an absolute value (platform-based limit), you can use the [`-XX:codecachetotal`](xxcodecachetotal.md) or the [`-Xcodecachetotal`](xcodecachetotal.md) options.
```
mpirvu, line 51: Another suggestion for rephrasing:

The default total size of the JIT code cache is computed as the minimum of the following two limits:

- A platform-based limit that is expressed as an absolute memory value. This limit is set at 256 MB for 64-bit systems and at 64 MB for 31/32-bit systems.
- A percentage of the amount of available physical memory the JVM process is allowed to use. By default, the JVM is not allowed to use more than 25% of the available physical memory for its code caches.

This second limit is relevant in memory constrained environments, as it prevents the JVM from using too much memory for its code caches, leaving too little memory for other VM/JIT data structures.

For example, on a 64-bit system, the platform specific code cache limit is 256 MB. If the JVM is constrained to less than 1024 MB, say 512 MB, then the code cache limit in effect becomes 128 MB (25% of 512 MB), because 128 MB is less than 256 MB (the platform-based limit).

To fine-tune the code cache size limit for your specific application use the `-XX:codecachetotalMaxRAMPercentage` option.
As an alternative, use the [`-XX:codecachetotal`](xxcodecachetotal.md) or the [`-Xcodecachetotal`](xcodecachetotal.md) options to set the code cache size limit as an absolute value (platform-based limit). Note that the absolute value specified by these options takes precedence over the percentage specified by the `-XX:codecachetotalMaxRAMPercentage` option.

docs/xcodecachetotal.md
```diff
@@ -46,6 +46,7 @@ The maximum size limits, for both the JIT code and data caches, that are in use
 
 ## See also
 
+- [`-XX:codecachetotalMaxRAMPercentage`](xxcodecachetotalmaxrampercentage.md)
```
mpirvu, line 49: Above we have to mention about the percentage option as well:

The default size is 256 MB for a 64-bit VM and 64 MB for a 31/32-bit VM, but no more than 25% of the physical memory available to the VM process,

docs/xxcodecachetotalmaxrampercentage.md
```diff
@@ -0,0 +1,59 @@
...
+
+        -XX:codecachetotalMaxRAMPercentage=<number>
+
+| Setting                 | Value      | Default                     |
+|-------------------------|------------|:---------------------------:|
+| `<number>`    | Between 1 and 100 (fractional numbers are allowed) |  25                       |
+
+Where `<number>` is the maximum limit that is expressed as a percentage of the available physical memory.
+
+For example, if you specify `-XX:codecachetotalMaxRAMPercentage=30`, the VM is not allowed to use more than 30% of the available physical memory for the internal JIT code caches.
```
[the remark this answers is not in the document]
Sreekala-Gopakumar: @mpirvu - Please refer to https://github.com/eclipse-openj9/openj9-docs/pull/1091#discussion_r1178822242 regarding why we use VM instead of JVM.

Sreekala-Gopakumar:
Jenkins doc stage

Sreekala-Gopakumar:
@mpirvu - I have incorporated the review feedback. Please check and confirm. Thanks!

Sreekala-Gopakumar:
@mpirvu - I have incorporated the review feedback. Please check and confirm. Thanks!

Sreekala-Gopakumar:
> @mpirvu - I have incorporated the review feedback. Please check and confirm. Thanks!

If there are no other feedback, could you please approve and merge. Thanks!

docs/xcodecachetotal.md
```diff
@@ -36,7 +36,7 @@ See [Using -X command-line options](x_jvm_commands.md) for more information abou
 
 ## Explanation
 
-By default, the total size of the JIT code cache is 256 MB for a 64-bit VM and 64 MB for a 31/32-bit VM. Long-running, complex, server-type applications can fill the JIT code cache, which can cause performance problems because not all of the important methods can be JIT-compiled. Use the `-Xcodecachetotal` option to increase or decrease the maximum code cache size to a setting that suits your application. The minimum size of the code cache is restricted to 2 MB.
+By default, the total JIT code cache size is 256 MB for a 64-bit VM and 64 MB for a 31/32-bit VM or 25% of the physical meomory available to the VM process, whichever is lesser. Long-running, complex, server-type applications can fill the JIT code cache, which can cause performance problems because not all of the important methods can be JIT-compiled. Use the `-Xcodecachetotal` option to increase or decrease the maximum code cache size to a setting that suits your application. The minimum size of the code cache is restricted to 2 MB.
```
mpirvu, line 39: Typo: meomey --> memory

Sreekala-Gopakumar:
Jenkins doc stage

Sreekala-Gopakumar:
@mpirvu - Done

mpirvu:
@Sreekala-Gopakumar Do you want me to merge the PR?

PR #1121, opened 2023-07-03
Persistent Shared Cache support on z/OS

### LINKED ISSUE #1097: Persistent Shared Cache support on z/OS

- https://github.com/eclipse-openj9/openj9/pull/17073

**Overview:**
The above PR adds support for persistent shared classes cache on z/OS. The default cache type on z/OS is still non-persistent. User can use `-Xshareclasses:` sub-option "[persistent](https://www.eclipse.org/openj9/docs/xshareclasses/#persistent)" to switch to persistent shared cache.

The following line can be removed from: https://www.eclipse.org/openj9/docs/xshareclasses/#persistent
`Note: Persisent caches are not supported on z/OS systems. z/OS supports only non-persistent caches.`

- On Java 8, the persistent shared cache is mapped below the 2G address range. There is machine configuration needed for the below 2G map.

- On Java 11 and up, the persistent shared cache is mapped above the 2G address range. There is machine configuration needed for the above 2G map. If user is running on z/OS version 2.4, an OS update needs to be installed. 

**Release target:**

Java 8 and up

z/OS

Sreekala-Gopakumar:
https://github.com/eclipse-openj9/openj9-docs/issues/1097

Updated the related topic with the support for persistent shared cache on z/OS.

Closes #1097


Sreekala-Gopakumar:
Jenkins doc stage

Sreekala-Gopakumar:
@babsingh - I have updated the documents. Please check and confirm. Thanks!

Sreekala-Gopakumar:
Jenkins doc stage

Sreekala-Gopakumar:
@babsingh - I have updated the documents. Please check and confirm. Thanks!

Sreekala-Gopakumar:
Jenkins doc stage

Sreekala-Gopakumar:
@babsingh - I have updated the documents. Please check and confirm.

https://eclipse-openj9.github.io/openj9-docs-staging/1121/version0.40/#support-for-persistent-shared-classes-cache-added-on-zos
https://eclipse-openj9.github.io/openj9-docs-staging/1121/xshareclasses/#persistent

Thanks!

babsingh:
++ @hangshao0 for review since he is the SCC SME.

@pshipton for backup.

docs/version0.40.md
```diff
@@ -42,6 +43,10 @@ In environments with low physical memory availability, the VM might use too much
 
 For more information, see [`-XX:codecachetotalMaxRAMPercentage`](xxcodecachetotalmaxrampercentage.md).
 
+### Support for persistent shared classes cache added on z/OS
+
+Persistent caches are now supported on z/OS systems. The default cache type on z/OS is still non-persistent. You can use the [`-Xshareclasses: persistent`](xshareclasses.md#persistent) option to switch to persistent shared cache.
```
babsingh, line 48: Extra space should be removed.
```suggestion
Persistent caches are now supported on z/OS systems. The default cache type on z/OS is still non-persistent. You can use the [`-Xshareclasses:persistent`](xshareclasses.md#persistent) option to switch to persistent shared cache.
```

pshipton:
@r30shah for review as well

docs/xshareclasses.md
r30shah, line 1: There are other section in this file that needs to be reviewed for supporting z/OS persistent shared class cache as well. For example, at line 422, we say z/OS supports only non-persistent caches which will no longer be true.

r30shah (changes requested):

docs/version0.40.md
hangshao0: > Persistent caches are now supported on z/OS systems

I suggest to change "Persistent caches" to "Persistent shared caches".

docs/xshareclasses.md
```diff
@@ -437,7 +437,28 @@ behavior, which can improve the performance of class loading from the shared cla
 
 :   Uses a persistent cache. The cache is created on disk, which persists beyond operating system restarts. Non-persistent and persistent caches can have the same name. On AIX, you must always use the `persistent` suboption when you run utilities such as `destroy` on a persistent cache.
 
-:   :fontawesome-solid-pencil:{: .note aria-hidden="true"} **Note:** Persisent caches are not supported on z/OS systems. z/OS supports only non-persistent caches.
+:   :fontawesome-solid-pencil:{: .note aria-hidden="true"} **Notes:**
+
+:    - On 31-bit VM on all versions of z/OS, the shared cache is memory mapped within the 0-2 GB address range. The size of the persistent shared class cache is limited by the system limit `MAXMMAPAREA`. For more information about the suggested value for the `MAXMMAPAREA` limit, see [Setting resource limits (z/OS)](https://www.eclipse.org/openj9/docs/configuring/#setting-resource-limits-zos).
```
hangshao0, line 442: > "The size of the persistent shared class cache"

Better to add "maximum". i.e. "The maximum size of the ..."

docs/xshareclasses.md
```diff
@@ -437,7 +437,28 @@ behavior, which can improve the performance of class loading from the shared cla
...
+     - On 64-bit VM, the shared cache is mapped as follows:
+         - On z/OS version 2.3 and earlier, the persistent shared cache is memory mapped within the 0-2 GB address range. The size of the persistent shared class cache is limited by the system limit `MAXMMAPAREA`.
```
hangshao0, line 444: > The size of the persistent shared class cache

and in the next paragraph:
> The size of persistent shared class cache is limited by the `MAXSHARE` value

Better to say "The maximum size of persistent shared class cache".

hangshao0, line 444: > On z/OS version 2.3 and earlier

Do we support version lower than 2.3 ? @joransiu  

If not, "and earlier" should be removed.

pshipton: We officially support 2.2 and later.
https://www.ibm.com/docs/en/sdk-java-technology/8?topic=installing-supported-environments

docs/xshareclasses.md
```diff
@@ -437,7 +437,28 @@ behavior, which can improve the performance of class loading from the shared cla
...
+         - On z/OS version 2.4 and later, the persistent shared cache is mapped above the 2 GB address range. The size of persistent shared class cache is limited by the `MAXSHARE` value within the `SMFLIMxx` member of `SYS1.PARMLIB`. For the above 2 GB memory map, the following machine configuration is required:
+
+              1. Add the following line in the SMFLIMxx member of SYS1.PARMLIB.
+
+             ```
+              REGION JOBNAME(*) MAXSHARE(2147483647)
```
joransiu, line 450: I worry about the recommendation that this specific line is **required**.  We are essentially saying to allow maximum shared pages for all jobs on the system, which may / may not be desirable.  This should be customized by the user, based on their needs.  Should we consider a suggested minimum size like we had for MAXMMAPAREA instead?

r30shah: I believe this value corresponds to the maximum allowed shared class cache that can be allocated using map64. May be we should put a value that is required for default shared class cache size and mention that size of maximum shared class cache is directly controlled by the value put in this parmlib?


Surely there would have been a rationale behind setting MAXMMAPAREA value to 40960 with respect to SCC. Just for knowledge, @hangshao0 , Joran do you know?

joransiu: Well, from https://www.ibm.com/docs/en/zos/2.5.0?topic=values-statements-parameters-smflimxx, 2147483647 is the maximum value allowed.  I agree, a size that is closer to what the typical SCC size (default x 2) would be more amenable as a starting point.  MAXMMAPAREA is in 4k pages, so 40960 ==> 160MB?

> MAXSHARE(nn)
    Specifies the maximum of source and target shared pages that can be used at one time by problem state callers using the IARVSERV SHARE services. This corresponds to IEFUSI parameter word 7, subword 4 output field. The maximum value for source and target shared pages is 2147483647.

r30shah: To keep it inline with what we recommended with `MAXMMAPAREA` (160 MB), may be `167772160` would be a good starting point for recommended value of MAXSHARE ?

Sreekala-Gopakumar: So, should I replace `2147483647` with `167772160` in `REGION JOBNAME(*) MAXSHARE(2147483647)`?

Sreekala-Gopakumar: @hangshao0 @r30shah - Any update on this? What change is required in this content? Thanks!

hangshao0: > Any update on this? What change is required in this content? Thanks!

Rahil is reaching out to the z/OS team on this. I believe he will update once we get answers from the z/OS expert.

hangshao0: I am fine with either 1GB or 2GB limit. Does the value of `MAXSHARE(1073741823)` mean bytes or pages ? If it is pages, it is much more than 1GB.

r30shah: It is actually a number of pages. Worked offline with Hang and also checked with z/OS expert. Based on that, I think we should use `MAXSHARE(524288)` that would allow us to use 2GB SCC which is the maximum size we can have with SCC. 

@hangshao0 , @joransiu  would appreciate your comment.

hangshao0: I am fine with `MAXSHARE(524288)`

Sreekala-Gopakumar:
Jenkins doc stage

Sreekala-Gopakumar:
@joransiu @babsingh @hangshao0 @r30shah @pshipton - I have incorporated the suggested changes. Please check and confirm. Thanks!

Sreekala-Gopakumar:
> @Sreekala-Gopakumar I had another offline discussion with Joran, summarizing here what we discussed.
> 
> Default SCC on JDK11 and onwards is 300 MB, the way this value works is it is the maximum number of pages which would be taken away from memory for shared memory. Specifying `524288` would block 2GB for shared memory, which on smaller system, would be too much.
> 
> Also we should also not recommend to set the value that allows only default shared class cache as if the consumers are running with multiple JVM, they would not be able to use persistent SCC, so specifying a value that would allow default SCC * 3 times (So 1GB) as originally suggested by Hang, and documenting if some one wants to scale-up (Having `n` number of JVMs using seperate SCC with default SCC size where `n*300` exceeds `1024MB` then how to update the `MAXSHARE`.

Could please clarify what are the changes that I should make? What value should I specify for `MAXSHARE`?  
So, in addition to the value change, you are suggesting that I should add a sentence specifying that you can scale up?

Also, if yes that the user can scale up, then could you please specify how they can update the `MAXSHARE`. 
Thanks!

r30shah:
Hi @Sreekala-Gopakumar , sorry for bit late response. Did more investigations on how the `MAXSHARE` value would impact. After discussions and consultation to zOS team, we should keep the recommended MAXSHARE value to `262144` which would allow users to use upto 1GB Shared class cache. That is the only change we are recommending, thanks in advance and again, apologies for couple of delayed response.

Sreekala-Gopakumar:
> Hi @Sreekala-Gopakumar , sorry for bit late response. Did more investigations on how the `MAXSHARE` value would impact. After discussions and consultation to zOS team, we should keep the recommended MAXSHARE value to `262144` which would allow users to use upto 1GB Shared class cache. That is the only change we are recommending, thanks in advance and again, apologies for couple of delayed response.

One more clarification - In the content, this configuration follows the lead in line " For the above 2 GB memory map, following is the machine configuration:" 

So, here we are saying "For the above **2 GB** memory map, " "the recommended MAXSHARE value is 262144 which would allow users to use upto **1 GB** Shared class cache."? This "above 2 GB" in the starting sentence and upto 1 GB in the recommended value is kind of confusing.

Thanks!

r30shah:
@Sreekala-Gopakumar I see the confusion. To be clear, it is mapping SCC in above 2GB address range which is different. Here is my suggestion,

```diff
- For the above 2 GB memory map, the following machine configuration is required:
+ Following machine configuration would allow user to create up to 1GB shared class cache mapped above the bar.
```

Sreekala-Gopakumar:
Jenkins doc stage

Sreekala-Gopakumar:
@r30shah - I have updated the document. Please check and confirm. If there are no other comments, this can merged. Thanks!

hangshao0 (approved):
Looks good to me.

r30shah (approved):

joransiu (approved):

pshipton (approved):

PR #1166, opened 2023-09-13
Add .project to .gitignore

gacholio:
Fixes: [#18133](https://github.com/eclipse-openj9/openj9/issues/18133)


pshipton:
Actually, I should have checked earlier. OpenJ9 doesn't check in a .project file, but it does have the following in the .gitignore file. I think this is what you want instead.
```
/.project
```

gacholio:
Pressed the wrong button, but at least there were no changes.

PR #1200, opened 2023-10-27
Improved algorithm with OpenSSL 3.0.12 and 3.1.4

Sreekala-Gopakumar:
https://github.com/eclipse-openj9/openj9-website-publish/pull/6

Added the information related to improved algorithm with the use of OpenSSL version 3 and later

### DIFF
--- a/docs/version0.41.md
+++ b/docs/version0.41.md
@@ -33,6 +33,7 @@ The following new features and notable changes since version 0.40.0 are included
 - [New `-XX:[+|-]EnableDynamicAgentLoading` option added](#new-xx-enabledynamicagentloading-option-added)
 - [New `-XX:[+|-]UseZlibNX` option added](#new-xx-usezlibnx-option-added)
 - [Support for OpenSSL 3.x](#support-for-openssl-3x)
+- [Performance improvement](#performance-improvement)
 
 ## Features and changes
 
@@ -81,6 +82,10 @@ AIX&reg; system adds the `zlibnx` library directory path in the `LIBPATH` enviro
 ### Support for OpenSSL 3.x
 
 OpenSSL 3.x is now supported on all operating systems. For more information about OpenSSL support, see [`Cryptographic operations`](introduction.md#cryptographic-operations).
+
+### Performance improvement
+
+Performance of algorithms with the use of OpenSSL version 3 and later is now enhanced. Improved algorithms include SHA256, AES, HmacSHA256, and ChaCha20. To obtain these improved algorithms and further enhance the performance of these algorithms on Linux&reg; and AIX, use OpenSSL 3.0.12 or a later 3.0.x version, or 3.1.4 or a later 3.1.x version.
 
 ## Known problems and full release information
 


Sreekala-Gopakumar:
Jenkins doc stage

pshipton:
Same comment as https://github.com/eclipse-openj9/openj9-website-publish/pull/6#issuecomment-1783221915, pls add the "or later".

Sreekala-Gopakumar:
Jenkins doc stage

Sreekala-Gopakumar:
Jenkins doc stage

Sreekala-Gopakumar:
@pshipton - I have incorporated the suggested change. Please check and confirm. Thanks!

pshipton:
Sorry, I should have looked closer. This update should go into the What's New rather than the OpenSSL section. It's not the switch to OpenSSL v3 which provided perf benefits, this switch originally regressed perf. However there were a number of changes made to improve the performance with OpenSSL v3. Previous user's of OpenSSL v3 on Linux will see these improvements when 0.41 is used. For the best performance the fix added to OpenSSL 3.0.12 (3.1.4) should also be used.

Sreekala-Gopakumar:
> Sorry, I should have looked closer. This update should go into the What's New rather than the OpenSSL section. It's not the switch to OpenSSL v3 which provided perf benefits, this switch originally regressed perf. However there were a number of changes made in 0.41 to improve the performance with OpenSSL v3. Previous user's of OpenSSL v3 on Linux will see these improvements when 0.41 is used. For the best performance the fix added to OpenSSL 3.0.12 (3.1.4) should also be used.

So I should delete the newly added content from the openssl topic? 
> Performance of algorithms is enhanced with the use of OpenSSL version 3 and later. Improved algorithms include SHA256, AES, HmacSHA256, and ChaCha20. To obtain these improved algorithms on Linux and AIX, use OpenSSL 3.0.12 or a later 3.0.x version, or 3.1.4 or a later 3.1.x version because OpenSSL is not bundled with the Semeru JDK on those platforms.

This should be now added to the What's new in version 0.41.0?

pshipton:
> So I should delete the newly added content from the openssl topic?

If you mean the content added by this PR, then yes.

> This should be now added to the What's new in version 0.41.0? 

Correct.

> From your comments, it seems that the content itself needs a change. Should it be as follows?

I don't think we should state "You must use these versions", in particular the word "must". They are not required, they just provide the best v3 performance.

Sreekala-Gopakumar:
Jenkins doc stage

Sreekala-Gopakumar:
> > So I should delete the newly added content from the openssl topic?
> 
> If you mean the content added by this PR, then yes.
> 
> > This should be now added to the What's new in version 0.41.0?
> 
> Correct.
> 
> > From your comments, it seems that the content itself needs a change. Should it be as follows?
> 
> I don't think we should state "You must use these versions", in particular the word "must". They are not required, they just provide the best v3 performance.

I have deleted the last line "You must use these versions on Linux and AIX because OpenSSL is not bundled with the Semeru JDK on those platforms." because I think it is not necessary here. I hope that is ok. Thanks!

pshipton (approved):

PR #1263, opened 2024-01-30
Glossary-0.43.0

Sreekala-Gopakumar:
Glossary topic added with first set of glossary terms.

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -1,0 +1,203 @@
+{:shortdesc: .shortdesc}
+{:new_window: target="_blank"}
+
+# Glossary
+*Last update: 2024/01/29*
+
+
+This glossary provides terms and definitions for Runtimes OpenJ9 &reg;.
+{:shortdesc}
+
+The following cross-references are used in this glossary:
+
+- *See* refers you from a nonpreferred term to the preferred term or from an abbreviation to the spelled-out form.
+- *See also* refers you to a related or contrasting term.
+
+
+<!--If you do not want letter links at the top of your glossary, delete the text between these comment tags.-->
+
+[A](#glossa)
+[B](#glossb)
+[C](#glossc)
+[D](#glossd)
+[E](#glosse)
+[F](#glossf)
+[G](#glossg)
+[H](#glossh)
+[I](#glossi)
+[J](#glossj)
+[K](#glossk)
+[L](#glossl)
+[M](#glossm)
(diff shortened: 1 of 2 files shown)


Sreekala-Gopakumar:
Jenkins doc stage

docs/glossary.md
```diff
@@ -0,0 +1,203 @@
...
+
+
+## O
+{: #glosso}
+
+
+### OpenSSL
+{: #x10447142}
+
+An open source version of the SSL and TLS security protocols, which provide encryption and server authentication over the internet.
```
pshipton, line 137: It's more than this. Copying from openssl.org it's also a "toolkit for general-purpose cryptography", which is the part of OpenSSL used by OpenJ9 with OpenJDK.

docs/glossary.md
```diff
@@ -0,0 +1,203 @@
...
+{: #x2038004}
+
+A security protocol that provides communication privacy. With SSL, client/server applications can communicate in a way that is designed to prevent eavesdropping, tampering, and message forgery.
+
+### SSL
+{: #x2483907}
+See [Secure Sockets Layer](#x2038004).
+
+
+### system dump (SDUMP)
```
pshipton, line 176: Where does SDUMP come from, does OpenJ9 use this term?

Sreekala-Gopakumar: We are expected to use the Terminology-in-ibm for creating glossaries. The idea is to use standardised terms and definition across IBM. We can suggest changes and make it specific to each product. SDUMP might be an existing term that is being reused. I am coordinating with Carla Quinn. I will pass on these comments to her and work on correcting it. Please do let me know any other concerns you might have. Thanks!

docs/glossary.md
```diff
@@ -0,0 +1,203 @@
...
+An open source version of the SSL and TLS security protocols, which provide encryption and server authentication over the internet.
+
+## P
+{: #glossp}
+
+
+### page
+{: #x2003565}
+
+A block of memory that equals 4096 bytes.
```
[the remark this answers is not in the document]
Sreekala-Gopakumar: How can we generalise it? Maybe as follows:

A contiguous block of virtual memory, usually of 4096 bytes size. A page size can be 2^n.

Sreekala-Gopakumar:
Creating a new pull request to use the latest content for glossary creation in 2025.

PR #1289, opened 2024-02-29
Update requirements.txt

Sreekala-Gopakumar:
Jenkins build was failing because of the following error: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts. poetry 1.8.1 requires packaging>=23.1, but you have packaging 22.0 which is incompatible. 
Updated the Requirements.txt with packaging==23.2

### DIFF
--- a/buildenv/requirements.txt
+++ b/buildenv/requirements.txt
@@ -63,7 +63,7 @@ mkdocs-material==9.5.11
     # via -r requirements.in
 mkdocs-material-extensions==1.3
     # via mkdocs-material
-packaging==22.0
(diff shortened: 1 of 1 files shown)


Sreekala-Gopakumar:
@pshipton - I have tested with this packaging and the documents are fine. Following is the console output for the build that was failing.
https://runtimes-jenkins.swg-devops.com/job/Build_mkdocs_for_kc_no_Docker/728/console

Could you please merge this file. Thanks!

pshipton (approved):

PR #1399, opened 2024-09-04
Release 0.46.1 What's new topic

Sreekala-Gopakumar:
Created a draft What's new topic for 0.46.1

### DIFF
--- a/docs/version0.46.1.md
+++ b/docs/version0.46.1.md
@@ -1,0 +1,43 @@
+<!--
+* Copyright (c) 2017, 2024 IBM Corp. and others
+*
+* This program and the accompanying materials are made
+* available under the terms of the Eclipse Public License 2.0
+* which accompanies this distribution and is available at
+* https://www.eclipse.org/legal/epl-2.0/ or the Apache
+* License, Version 2.0 which accompanies this distribution and
+* is available at https://www.apache.org/licenses/LICENSE-2.0.
+*
+* This Source Code may also be made available under the
+* following Secondary Licenses when the conditions for such
+* availability set forth in the Eclipse Public License, v. 2.0
+* are satisfied: GNU General Public License, version 2 with
+* the GNU Classpath Exception [1] and GNU General Public
+* License, version 2 with the OpenJDK Assembly Exception [2].
+*
+* [1] https://www.gnu.org/software/classpath/license.html
+* [2] https://openjdk.org/legal/assembly-exception.html
+*
+* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0 OR GPL-2.0-only WITH OpenJDK-assembly-exception-1.0
+-->
+
+# What's new in version 0.46.1
+
+The following new features and notable changes since version 0.46.0 are included in this release:
+
+- [New binaries and changes to supported environments](#binaries-and-supported-environments)
+
+
+## Features and changes
+
+### Binaries and supported environments
+
+Eclipse OpenJ9&trade; release 0.46.1 supports OpenJDK 8, 11, 17, 21, and 22.
+
+To learn more about support for OpenJ9 releases, including OpenJDK levels and platform support, see [Supported environments](openj9_support.md).
+
+## Known problems and full release information
+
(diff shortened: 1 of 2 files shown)


docs/version0.46.1.md
```diff
@@ -0,0 +1,47 @@
...
+- [New binaries and changes to supported environments](#binaries-and-supported-environments)
+
+
+## Features and changes
+
+### Binaries and supported environments
+
+Eclipse OpenJ9&trade; release 0.46.1 supports OpenJDK 8, 11, 17, 21, and 22.
+
+Support for running OpenJ9 with all versions of OpenJDK on CentOS 6, CentOS 7, Red Hat Enterprise Linux (RHEL) 6, and RHEL 7 is removed.
```
pshipton, line 37: I know this is a copy of the statement in 0.46, but I wonder if it can be worded in a different way. We didn't change the JVM or how it's built, but we are no longer claiming these OSes are supported platforms.

docs/version0.46.1.md
```diff
@@ -0,0 +1,47 @@
...
+
+RHEL 8.6 and 9.0 are out of support. RHEL 8.8 is the new minimum operating system level.
```
pshipton, line 39: It should be RHEL 8.8 and 9.2.

docs/version0.46.1.md
Sreekala-Gopakumar: Can we delete this line then because the support was already removed in 0.46.0 release? Or is there any other information that is required here?

docs/version0.46.1.md
Sreekala-Gopakumar: If RHEL 8.8 and 9.2 are now out of support then which is the new minimum operating system level? 

Also, do when then have to change the Supported environments topic? https://eclipse.dev/openj9/docs/openj9_support/

docs/version0.46.1.md
pshipton: Yes, it can be deleted since it isn't a change from 0.46

docs/version0.46.1.md
pshipton: The correction is:
RHEL 8.6 and 9.0 are out of support. RHEL 8.8 **and 9.2** are the new minimum operating system levels.

Like the previous comment, this isn't a change from 0.46.0. You could make this fix in that topic and remove the line here.

docs/version0.46.1.md
pshipton: Although it would be nice to update the 0.46 topic with new wording, to be clear about what "support is removed" means.

Sreekala-Gopakumar: Is the following content applicable for OpenJ9?

No release binaries or packages will be produced for OpenJ9 on these platforms and no support will be provided (including interim fixes in support cases).

pshipton: Technically OpenJ9 is a source code only project. These things refer to Semeru, which is a down stream product that uses OpenJ9.

Although we aren't testing OpenJ9 on these out of public support platforms, it can still be built and run there, there were no code changes to prevent it. Any request for OpenJ9 open source support for an OS specific problem on these OSes will be refused, but for a bug in OpenJ9 unrelated to the OS, it doesn't really matter on what OS it was found.

pshipton: What I don't like about the existing wording is that it may imply we changed something in OpenJ9. What I'm looking for is something like:

CentOS 6, CentOS 7, Red Hat Enterprise Linux (RHEL) 6, and RHEL 7 are removed from the list of supported platforms.

Sreekala-Gopakumar:
Jenkins doc stage

docs/version0.46.1.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Binaries and supported environments
+
+Eclipse OpenJ9&trade; release 0.46.1 supports OpenJDK 8, 11, 17, 21, and 22.
+
+To learn more about support for OpenJ9 releases, including OpenJDK levels and platform support, see [Supported environments](openj9_support.md).
+
+## Known problems and full release information
+
+To see known problems and a complete list of changes between Eclipse OpenJ9 v0.46.0 and v0.46.1 releases, see the [Release notes](https://github.com/eclipse-openj9/openj9/blob/master/doc/release-notes/0.46.1/0.46.1.md).
```
pshipton, line 41: I assume you will create the release note document so the link works?

pshipton (approved):

docs/version0.46.1.md
Sreekala-Gopakumar: https://github.com/eclipse-openj9/openj9/pull/20123

PR #1444, opened 2024-10-29
Pull the Ubuntu 22.04 image from Amazon instead of Dockerhub

pshipton:
Related to the Nov 15, 2024 change in Docker's subscription model, and ease of building the image at companies building the docs image for testing.

I tried building locally using the instructions at https://github.com/eclipse-openj9/openj9-docs/wiki/Testing-OpenJ9-MkDocs-dependencies and it worked.

### DIFF
--- a/buildenv/Dockerfile
+++ b/buildenv/Dockerfile
@@ -20,7 +20,7 @@
 # SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0 OR GPL-2.0-only WITH OpenJDK-assembly-exception-1.0
 #
 
-FROM ubuntu:22.04@sha256:6042500cf4b44023ea1894effe7890666b0c5c7871ed83a97c36c76ae560bb9b
+FROM public.ecr.aws/lts/ubuntu:22.04_stable@sha256:1582c29f34a48752e406f1a261fe9545fad895da3f6bb4be55bc82485557564e
 MAINTAINER openj9.bot <openj9.bot@gmail.com>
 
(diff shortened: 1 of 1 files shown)


buildenv/Dockerfile
```diff
@@ -20,7 +20,7 @@
 # SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0 OR GPL-2.0-only WITH OpenJDK-assembly-exception-1.0
 #
 
-FROM ubuntu:22.04@sha256:6042500cf4b44023ea1894effe7890666b0c5c7871ed83a97c36c76ae560bb9b
+FROM public.ecr.aws/lts/ubuntu:22.04_stable@sha256:1582c29f34a48752e406f1a261fe9545fad895da3f6bb4be55bc82485557564e
```
keithc-ca, line 23: How does this improve security? Either way by specifying a SHA256 hash, we'll get the image we expect.

Also, why is the hash changing?

pshipton: It's not an attempt to improve security, it's just a change to use amazon to get the image instead of dockerhub. I hard coded the hash I got from pulling the amazon image. The latest hash at dockerhub is different from what we've been using, we should be regularly updating the image to the latest which we haven't been doing. The latest hash at dockerhub isn't the same as the one at amazon, possibly they are updated at different times and don't necessary have the same content.

AdamBrousseau:
Should we kick off the image rebuild?
https://openj9-jenkins.osuosl.org/view/Website-Doc/job/Build-Doc-Docker_Container/

pshipton:
Yes, I should have done that. I've started it now.
https://openj9-jenkins.osuosl.org/view/Website-Doc/job/Build-Doc-Docker_Container/42

PR #1526, opened 2025-05-23
Remove unwanted colon in reference to `-XXallowvmshutdown` option

keithc-ca:


keithc-ca:
Credit to @babsingh for noticing the error.

pshipton (approved):

PR #1555, opened 2025-07-21
Document Offheap

### LINKED ISSUE #1540: Document Offheap
Overview:
Balanced GC has a new throughtput improvement feature. For the most part it's transparent to the user (it's enabled by default - no user action needed), but some docs should be update that already refer to under-the-hood implementation - specifically arraylets, that are not used any more.

https://eclipse.dev/openj9/docs/gc/

```
Most objects are easily contained within the minimum region size of 512 KB. However, to support large arrays, which cannot be contained in a region, the balanced GC policy uses an arraylet representation in the heap. For more information about structure and layout, see [Arraylets](https://eclipse.dev/openj9/docs/allocation/#arraylets).

Note: With arraylets, JNI access to array data might involve reconstituting arraylets as contiguous arrays, which can significantly slow down processing.
```

This should state that now all arrays are contagious, but if larger than a region size they are stored into separate area. That area is a few times larger than the main heap (aka Offheap), but it's only committed for the live arrays currently stored there. While increase of virtual address space will be observed, the total amount of committed memory (resident) will typically stay the same or marginally increase.

Verbose GC changes are updated:
- initalize stanza will show that Offhaep is used an how large it is
- each GC cycle will show how many offheap live objects are at GC start and GC end.

Release notes should mention that there is a tangible throughput improvement for Balanced GC policy, with no or small negative impact on committed memory. 

Also could  be mentioned that this functionality is a superset of Double Mapping and user that enabled it could remove that command line option (although, we never documented that option,  and provided it only directly to a small number of users - should we refer to it then!?). More importantly, all platforms may experience significant improvement with JNI Critical API, when dealing with larger arrays (somehting that Double Mapping provided on request only to Linux based platforms).

Release target:
Eclipse OpenJ9 0.53

Applies to the following JDK versions:
All JDKs

Applies to the following platforms:
All but zTPF

For new command line options:
While we have an command line option to disable Offheap (go back to Arraylet based Balanced GC), we should not expose it, and only suggest to the user directly in a case of unexpected problems.

Sreekala-Gopakumar:
https://github.com/eclipse-openj9/openj9-docs/issues/1540

Offheap support added to the balanced gc policy

Closes #1540

### DIFF
--- a/docs/gc.md
+++ b/docs/gc.md
@@ -116,7 +116,6 @@ However, even though pause times are typically evened out across GC operations, 
 
 ### GC processing
 
-
 During VM startup, the GC divides the heap memory into regions of equal size. These regions remain static for the lifetime of the VM and are the basic unit of garbage collection and allocation operations. For example, when the heap is expanded or contracted, the memory committed or released corresponds to a certain number of regions. Although the Java heap is a contiguous range of memory addresses, any region within that range can be committed or released from a pool as required. This enables the Balanced GC to contract the heap more dynamically and aggressively than other garbage collectors, which typically require the committed portion of the heap to be contiguous.
 
 Regions impose a maximum object size. Objects are always allocated within the bounds of a single region and are never permitted to span regions. The region size is always a power of two; for example, 512 KB, 1 MB, and so on (where KB is 2<sup>10</sup> bytes and MB is 2<sup>20</sup> bytes). The region size is selected at startup based on the maximum heap size. The collector chooses the smallest power of two which will result in less than 2048 regions, with a minimum region size of 512 KB. Except for small heaps (less than about 512 MB) the VM aims to have between 1024 and 2047 regions.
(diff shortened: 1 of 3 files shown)


Sreekala-Gopakumar:
Jenkins doc stage

Sreekala-Gopakumar:
Jenkins doc stage

Sreekala-Gopakumar:
@amicic @pshipton - I have updated the topics. Please let me know if any changes are required. Thanks!

docs/gc.md
```diff
@@ -133,9 +132,9 @@ A global sweep operation also runs to reclaim memory so that it can create empty
...
+Before the 0.53.0 release, to support large arrays, which cannot be contained in a region, the `balanced` GC policy used an *arraylet* representation in the heap. With arraylets, JNI access to array data might involve reconstituting arraylets as contiguous arrays, which can significantly slow down processing. For more information about structure and layout, see [Arraylets](allocation.md#arraylets).
 
-:fontawesome-solid-pencil:{: .note aria-hidden="true"} **Note:** With arraylets, JNI access to array data might involve reconstituting arraylets as contiguous arrays, which can significantly slow down processing.
+All arrays are contiguous, but now if larger than a region size, the arrays are stored into a separate area. That area, the *Offheap*, is a few times larger than the main heap to prevent fragmentation, but it is only committed for the live arrays currently stored there. While increase of virtual address space will be observed, the total amount of committed memory (resident) will typically stay the same or increase marginally.
```
amicic, line 137: I don't want to go too deep into details, but slightly more correct is to say that all array's data is contiguous. And than we can say that array's data  (rather than whole array) is stored into a separate area.

To clarify: array's header is always in the main heap, whether the data is adjacent to the header (small arrays, data in main heap) or non-adjacent to the header (larger array, data in Offheap).  We will have a blog providing more details - no need for that level of details here, but perhaps we could use array's data - not sure myself?

Sreekala-Gopakumar:
Jenkins doc stage

amicic (approved):

Sreekala-Gopakumar:
@amicic - Can this be merged now?

amicic:
@pshipton please, have a quick look and merge

Sreekala-Gopakumar:
Jenkins doc stage

pshipton (approved):

PR #1627, opened 2025-11-12
Update default heap value when container limit >= physical memory

### LINKED ISSUE #1624: Update default heap value when container limit >= physical memory
**Subject:** If container limit >= physical memory then default maximum heap is 25% as opposed to 75% as mentioned in this doc.
*https://eclipse.dev/openj9/docs/xxusecontainersupport/*

VM treats container limit is not set when memlimit exceeds the node’s capacity(https://github.com/eclipse-omr/omr/blob/28882169571db41cd824be0b280327b7fa0b9323/port/unix/omrsysinfo.c#L6970-L6974)

**Suggestion:**
Add 4th point under "The default heap size for containers takes affect only when the following conditions are met:" something as

```
4. The memory limit set for the container is less than physical memory.
```

Sreekala-Gopakumar:
https://github.com/eclipse-openj9/openj9-docs/issues/1624

Added information for clarity.

Closes #1624


sharanpatil123:
@Sreekala-Gopakumar - LGTM, Thanks.

pshipton:
If the container limit is >= physical memory then it falls back to the -Xmx defaults as outlined in https://eclipse.dev/openj9/docs/openj9_defaults/

PR #1635, opened 2025-11-27
Adjustment of default maximum heap size based on initial size

### LINKED ISSUE #1634: Document silent adjustment of default maximum heap size based on initial size
I guess required comments to be added to `-Xmx` and `-Xms` articles, as well as `-XX:MaxRAMPercentage` and `-XX:InitialRAMPercentage`.

Description of the case:
If maximum object heap size `-Xmx` is not specified explicitly, so default setting is used and if `-Xms` initial object heap size is specified, if initial size is larger than default maximum increase default maximum size to match initial.

For example for machine with 16 GB RAM default maximum size is 25% of RAM, so 4 GB.
If initial size is specified as `-Xms6g` for such configuration older JVMs failed to start due exceeding 4 GB (not obvious behaviour because customer might not realized amount of RAM on given machine).
With current adjustment default maximum heap size is going to be increased silently to 6 GB and allows JVM to start.

Sreekala-Gopakumar:
https://github.com/eclipse-openj9/openj9-docs/issues/1634

Updated the -Xms/-Xmx topic and added the information in the 0.57.0 What's new topic.

Closes #1634


dmitripivkine:
Looks good for me.

PR #1685, opened 2026-03-30
Document new -XX:[+/-]UseMediumPageSize

### LINKED ISSUE #1678: Document new -XX:[+/-]UseMediumPageSize
https://github.com/eclipse-openj9/openj9/issues/19052

**Overview:**
On AIX, launchers add a new environment variable definition by default:  `LDR_CNTRL` is set to `TEXTPSIZE=64K@DATAPSIZE=64K@STACKPSIZE=64K@SHMPSIZE=64K`. This can be disabled via the new command-line option `-XX:-UseMediumPageSize`, or explicitly (re-)enabled with `-XX:+UseMediumPageSize`.

**Release target:**
Eclipse OpenJ9 0.59.0

All versions

AIX

- Option name and syntax: *-XX:[+|-]UseMediumPageSize*
- Purpose: Controls whether `LDR_CNTRL` is added to the environment or not.

Sreekala-Gopakumar:
https://github.com/eclipse-openj9/openj9-docs/issues/1678

Documented the new option -XX:[+|-]UseMediumPageSize. Created the What's new topic and updated mkdocs.yml file. Updated the related topics.

Closes #1678

### DIFF
--- a/docs/xxusemediumpagesize.md
+++ b/docs/xxusemediumpagesize.md
@@ -1,0 +1,53 @@
+﻿<!--
+* Copyright (c) 2017, 2026 IBM Corp. and others
+*
+* This program and the accompanying materials are made
+* available under the terms of the Eclipse Public License 2.0
+* which accompanies this distribution and is available at
+* https://www.eclipse.org/legal/epl-2.0/ or the Apache
+* License, Version 2.0 which accompanies this distribution and
+* is available at https://www.apache.org/licenses/LICENSE-2.0.
+*
+* This Source Code may also be made available under the
+* following Secondary Licenses when the conditions for such
+* availability set forth in the Eclipse Public License, v. 2.0
+* are satisfied: GNU General Public License, version 2 with
+* the GNU Classpath Exception [1] and GNU General Public
+* License, version 2 with the OpenJDK Assembly Exception [2].
+*
+* [1] https://www.gnu.org/software/classpath/license.html
+* [2] https://openjdk.org/legal/assembly-exception.html
+*
+* SPDX-License-Identifier: EPL-2.0 OR Apache-2.0 OR GPL-2.0-only WITH Classpath-exception-2.0 OR GPL-2.0-only WITH OpenJDK-assembly-exception-1.0
+-->
+
+# -XX:[+|-]UseMediumPageSize
+
+**AIX only**
+
+This option is used to enable or disable the default definition setting of the `LDR_CNTRL` environment variable.
+
+## Syntax
+
+        -XX:[+|-]UseMediumPageSize
+
+| Setting                      | Effect  | Default                                                                        |
+|------------------------------|---------|:------------------------------------------------------------------------------:|
+| `-XX:+UseMediumPageSize` | Enable  |      :fontawesome-solid-check:{: .yes aria-hidden="true"}<span class="sr-only">yes</span>    |
+| `-XX:-UseMediumPageSize` | Disable |                                                                           |
+
+## Explanation
+
+On AIX systems, by default the Java launcher sets the `LDR_CNTRL` environment variable as `TEXTPSIZE=64K@DATAPSIZE=64K@STACKPSIZE=64K@SHMPSIZE=64K` to specify medium page sizes (64 KB) for the text, data, stack, and shared memory segments. This default setting is used to improve performance.
+
+You can use `-XX:-UseMediumPageSize` to disable the default setting of the `LDR_CNTRL` environment variable. When this option is used, the Java launcher uses the default AIX page sizes (4 KB for each segment).
+
+You can re-enable the default setting with the `-XX:+UseMediumPageSize` option.
+
+## See also
+
+- [What's new in version 0.59.0](version0.59.md#new-xx-usemediumpagesize-option-is-added)
+- [Configuring your systems](configuring.md#aix-systems)
+- [-Xlp:codecache](xlpcodecache.md#aix)
+
+<!-- ==== END OF TOPIC ==== xxusemediumpagesize.md ==== -->
--- a/docs/configuring.md
+++ b/docs/configuring.md
@@ -183,6 +183,9 @@ The following example sets 4 KB for text and 64 KB for stack, native data, and h
 ```
 LDR_CNTRL=TEXTPSIZE=4K@STACKPSIZE=64K@DATAPSIZE=64K
 ```
+On AIX systems, the `LDR_CNTRL` environment variable is set to `TEXTPSIZE=64K@DATAPSIZE=64K@STACKPSIZE=64K@SHMPSIZE=64K` by default to specify medium page sizes (64K) for the text, data, stack, and shared memory segments to improve performance.
+
+From the 0.59.0 release onwards, you can use the `-XX:-UseMediumPageSize` option to disable this default setting of the `LDR_CNTRL` environment variable. For more information, see [`-XX:[+|-]UseMediumPageSize`](xxusemediumpagesize.md).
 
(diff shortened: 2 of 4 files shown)


docs/xxusemediumpagesize.md
```diff
@@ -0,0 +1,53 @@
...
+| Setting                      | Effect  | Default                                                                        |
+|------------------------------|---------|:------------------------------------------------------------------------------:|
+| `-XX:+UseMediumPageSize` | Enable  |      :fontawesome-solid-check:{: .yes aria-hidden="true"}<span class="sr-only">yes</span>    |
+| `-XX:-UseMediumPageSize` | Disable |                                                                           |
+
+## Explanation
+
+On AIX systems, by default the Java launcher sets the `LDR_CNTRL` environment variable as `TEXTPSIZE=64K@DATAPSIZE=64K@STACKPSIZE=64K@SHMPSIZE=64K` to specify medium page sizes (64 KB) for the text, data, stack, and shared memory segments. This default setting is used to improve performance.
+
+You can use `-XX:-UseMediumPageSize` to disable the default setting of the `LDR_CNTRL` environment variable. When this option is used, the Java launcher uses the default AIX page sizes (Standard page size - 4 KB).
```
keithc-ca: Suggest the parenthetical comment should be "(standard page size is 4kB)"
* standard should not be capitalized
* use "is" instead of the hyphen
* the multiplier "k" should be lowercase

keithc-ca (changes requested):

docs/xxusemediumpagesize.md
Sreekala-Gopakumar: So, you mean in this instance of page sizes "kB" (which is equal to 1000 bytes) is applicable instead of KB (1024 bytes)?

If that is so, then should I change "KB" to "kB" in the following topics as well?

https://eclipse-openj9.github.io/openj9-docs-staging/1685/configuring/#aix-systems
https://eclipse-openj9.github.io/openj9-docs-staging/1685/xlpcodecache/

As per the IBM style guide, there should be a space between the number and the unit of measurement, so I would recommend writing 4kB as 4 kB

keithc-ca: In the SI system "k" means 1000, but in the computing realm, "k" means 1024. In neither domain is "K" meaningful (although humans can usually infer the intent).

The space in "4 kB" is fine.

I would support improving other topics which use K instead of k as a unit.

Sreekala-Gopakumar: > In the SI system "k" means 1000, but in the computing realm, "k" means 1024. In neither domain is "K" meaningful (although humans can usually infer the intent).
> 
> The space in "4 kB" is fine.
> 
> I would support improving other topics which use K instead of k as a unit.

is this change (kB not KB) specific to only page sizes or do you mean this is applicable across all instances in the entire docs? I am asking because I don't see any instance of "kB" in any of the docs that I am dealing with. 

Based on the IBM style guide, kB seems to be the SI system, like you said. But, maybe KB is the United States customary system (or the American system), the system that is the commonly used system of measurement in the US.

And since we use the American english in the IBM documents, then are we supposed to use "KB" and not "kB"? I guess I will have to look into this deeper before I make this change.

keithc-ca: Case is a significant difference between `M` and `m` (a factor of a billion or more); usually it's clear which is meant from the context, but we should not expect readers to guess - I'm just suggesting using the proper case for `k` as well. I don't think this has anything to do with the American system.

I've also seen documents that use `B` and `b` inappropriately (the former is a byte or 8 bits, the latter is a single bit: In this case, it's often less clear which is intended.

Sreekala-Gopakumar: I was referring to this section in the IBM style guide https://ibmdocs-test.dcs.ibm.com/docs/en/ibm-style?topic=measurement-units and https://ibmdocs-test.dcs.ibm.com/docs/en/ibm-style?topic=measurement-units#multiplier-prefixes-for-bits-and-bytes__title__1, where the SI system and the American system is discussed.

Like I said, if you are recommending that I change "KB" to "kB" in all the topics (whether it is related to page sizes or not), then I would like to have a day or two to discuss this with other content experts because it is a huge change and will have an impact on not just OpenJ9 docs but also SDK 8 and Semeru docs as well.

keithc-ca: Sure, that's reasonable, but I still maintain we don't want another inappropriate occurrence of KB here.

Sreekala-Gopakumar: @keithc-ca - After much discussion, it is recommended that instead of "**kB**", "**KiB**" should be used. So, can I go ahead with changing all instances of "KB" to "KiB" across OpenJ9 and SDK 8 and Semeru documents? Thanks!

keithc-ca: If we're going to start using `KiB` rather than `kB` (which, although I'm not thrilled about, makes sense), then we should be using `MiB`, `GiB`, etc. instead of `MB`, `GB`, etc.

pshipton: Not sure why you think `kB` is appropriate for memory (x 1024), it's `KB`. 
I agree that if we are changing to `KiB` then all of them should be changed.

keithc-ca: See https://en.wikipedia.org/wiki/Metric_prefix; all prefixes smaller than `M` are lowercase. The computing world followed that lead.

pshipton: I think you want https://en.wikipedia.org/wiki/Binary_prefix for powers of 2.

keithc-ca: I'm a little surprised at that, but you're right: the convention for the 1024 multiplier seems to have (almost) always been the uppercase K. I stand corrected: This should use `KB`.

keithc-ca: Now if we can only get people to use M and m properly: A computer with 64mb (less than 0.008 bytes) of memory would not much use. ;-)

Sreekala-Gopakumar: Thanks Keith and Peter for helping to arrive at a conclusion. I will retain "KB" in the docs.

Sreekala-Gopakumar:
Jenkins doc stage

Sreekala-Gopakumar:
@keithc-ca @pshipton - Please let me know if any other changes are required. Thanks!

docs/xxusemediumpagesize.md
```diff
@@ -0,0 +1,53 @@
...
+| Setting                      | Effect  | Default                                                                        |
+|------------------------------|---------|:------------------------------------------------------------------------------:|
+| `-XX:+UseMediumPageSize` | Enable  |      :fontawesome-solid-check:{: .yes aria-hidden="true"}<span class="sr-only">yes</span>    |
+| `-XX:-UseMediumPageSize` | Disable |                                                                           |
+
+## Explanation
+
+On AIX systems, by default the Java launcher sets the `LDR_CNTRL` environment variable as `TEXTPSIZE=64K@DATAPSIZE=64K@STACKPSIZE=64K@SHMPSIZE=64K` to specify medium page sizes (64 KB) for the text, data, stack, and shared memory segments. This default setting is used to improve performance.
+
+You can use `-XX:-UseMediumPageSize` to disable the default setting of the `LDR_CNTRL` environment variable. When this option is used, the Java launcher uses the default AIX page sizes (4 KB).
```
keithc-ca: The last part here feels awkward: It says "sizes", but only mentions one size. Suggest one of the two following:

> default AIX page size (4 KB)

> default AIX page sizes (4 KB for each segment)

Sreekala-Gopakumar:
Jenkins doc stage

Sreekala-Gopakumar:
@keithc-ca - Are there any other changes? Can this be merged? Thanks!

keithc-ca (approved):

PR #1723, opened 2026-06-25
When building doc, run jenkins git checkout outside the container

pshipton:
Sometimes doc staging builds fail because of "nothing added to commit" even though there were changes that should have been added.

The jenkins git launcher can't run inside the container and runs outside. Although the steps are sequential, the checkout outside the container completes when the git plugin returns, but the filesystem writes may not yet be visible in the container. Doing the git checkouts before running in the container ensures proper sequence.

There is a warning about the jenkins git launcher in the job output:

Warning: JENKINS-30600: special launcher
org.jenkinsci.plugins.docker.workflow.WithContainerStep$Decorator$1@4adfedbe; decorates RemoteLauncher[hudson.remoting.Channel@5098dcf9:cent9-x86-2] will be ignored (a typical symptom is the Git executable not being run inside a designated container)

Example failing build https://openj9-jenkins.osuosl.org/job/PullRequest-Doc-test_on_staging_site/893/

### DIFF
--- a/JenkinsFile_build_doc.groovy
+++ b/JenkinsFile_build_doc.groovy
@@ -79,19 +79,36 @@ timeout(time: 6, unit: 'HOURS') {
                 def TMP_DESC = (currentBuild.description) ? currentBuild.description + "<br>" : ""
                 currentBuild.description = TMP_DESC + "<a href=${JENKINS_URL}computer/${NODE_NAME}>${NODE_NAME}</a>"
 
+				// The jenkins git launcher can't run in the container, run it outside to avoid file sync problems.
+                stage('Checkout repos') {
+                    dir(BUILD_DIR) {
+                        checkout changelog: false, poll: false,
+                            scm: [$class: 'GitSCM',
+                                branches: [[name: CLONE_BRANCH]],
+                                extensions: [
+                                    [$class: 'CheckoutOption', timeout: 30],
+                                    [$class: 'CloneOption', timeout: 30]
+                                ],
+                                userRemoteConfigs: [[refspec: REFSPEC, url: "${HTTP}${OPENJ9_REPO}"]]]
+                    }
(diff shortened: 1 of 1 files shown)


pshipton:
jenkins doc stage

pshipton:
https://openj9-jenkins.osuosl.org/job/PullRequest-Doc-test_on_staging_site/897/ shows the change is working.

keithc-ca:
Even with this change, git repositories are still modified within the container, and consumed outside. If we can't rely on file-system consistency between those two contexts, then problems remain.

pshipton:
The push happens from inside the container, since it's using `sh "git push ..."`, and not the jenkins git launcher. You'll notice there is no more `Warning: JENKINS-30600` or other codes in the test build.

keithc-ca:
Not all pushes happen inside the container. The calls to `push_doc_with_cred()` on lines 156, 160, and 189 are from within the container while the call on line 204 is outside.

pshipton:
I expect once you exit the container all writes are synced. The fix assumes similarly, that writes that occur before the container is started will be available inside the container. My understanding is the problem is with the jenkins git launcher, which doesn't wait for file syncing.

keithc-ca:
If the issues stem from use of the jenkins git launcher within the container, then I think we still will have problems with the `BUILD_TYPE = "RELEASE"` case which uses `stash` inisde, paired with `unstash` outside.

Otherwise, this appears to be an improvement, and we can delay fixing the remaining issue if/when we experience it. Thoughts?

pshipton:
We've only noticed a problem running the `doc stage`, I've never noticed a problem doing a release.

keithc-ca (approved):

keithc-ca:
Releases are much less frequent than PR staging, which may explain our experience. As I said, I think this is an improvement worth merging.

PR #1724, opened 2026-06-30
Document concurrentKickoffTenuringHeadroom

### LINKED ISSUE #1714: Document concurrentKickoffTenuringHeadroom
We need to document the following tuning command line option:

-Xgc:concurrentKickoffTenuringHeadroom=xx

xx is a percentage, and valid values are 0 to 99

(an example option that uses percantage -Xgc:dnssexpectedtimeratiomaximum - it's unrelated, just mentioning for uniformity of how we  deal with options that expect percentage).

xx is a percentage of free memory that was observed at the end of previous global GC
That free memory is added on top of automatic concurrent kickoff calculation.

For example, the heurstic (based on steady state metrics) may say that kickoff should be at 10% of free memory and if the option says 5% (to compensate for spikes in tenuring), than the total is 15%.

That headroom is already non-0: 2% for Stop-The-World and 10% for Concurrent Scavenge.
If something is to be specified, it should be above these defaults to compensate for spikes in Tenuring (a typical scenario where this would be used).

Applies to gencon and optavgpause.

Sreekala-Gopakumar:
https://github.com/eclipse-openj9/openj9-docs/issues/1714

Added the new option to the xgc topic. Updated the What's new topic.

Closes #1714

### DIFF
--- a/docs/xconcurrentslack.md
+++ b/docs/xconcurrentslack.md
@@ -23,22 +23,61 @@
 
 # -Xconcurrentslack
 
-Attempts to keep the specified amount of the heap space free in concurrent collectors by starting the concurrent operations earlier.
-
-Using this option can sometimes alleviate pause time problems in concurrent collectors at the cost of longer concurrent cycles, affecting total throughput.
+This option specifies an absolute amount of free space or headroom (in bytes) in the tenure space on top of the automatic concurrent kickoff point. By using this option, you can trigger the concurrent global garbage collection (GC) cycle earlier than the GC heuristic computed kickoff point.
 
 ## Syntax
 
         -Xconcurrentslack<size>
 
-: See [Using -X command-line options](x_jvm_commands.md) for more information about specifying the `<size>` parameter.
+: See [Using -X command-line options](x_jvm_commands.md) for more information about the `<size>` parameter.
 
-## Default behavior
+## Explanation
 
-The default value is 0, which is optimal for most applications.
+Ideally, concurrent GC needs to complete its operation before application consumes all free memory. Otherwise, GC will transition into the stop-the-world (STW) mode, which means longer pauses. Due to fluctuations in application characteristics such as allocation rate, object survival rate, CPU utilization, GC might mis-predict the concurrent GC kickoff point, and start it too late or progress too slowly.
 
-This option is not supported with the balanced GC policy (`-Xgcpolicy:balanced`), the *optimize for throughput* policy (`-Xgcpolicy:optthruput`), or metronome GC policy (`-Xgcpolicy:metronome`).
+To compensate for such fluctuations, you can specify this option so that the VM starts concurrent GC operations earlier to try to keep free heap space available.
 
+Starting concurrent GC operations earlier can reduce pause times, but it might increase frequency because of suboptimal use of free memory.
 
+The default value is `0`, which is optimal for most applications.
+
+:fontawesome-solid-pencil:{: .note aria-hidden="true"} **Note:** This option is not supported by the following GC policies:
+
+- Balanced ([`-Xgcpolicy:balanced`](gc.md#balanced-policy))
+- Optimize for throughput ([`-Xgcpolicy:optthruput`](gc.md#optthruput-policy))
+- Metronome ([`-Xgcpolicy:metronome`](gc.md#metronome-policy))
+
+### How free space is interpreted
+
+The meaning of free space depends on the GC policy in use.
+
+**Optimize for average pause time (`optavgpause`) policy**
+
+In this policy, application threads and the concurrent GC kickoff heuristic measure free space in the same way. Therefore, the `-Xconcurrentslack` value can be used directly.
+
+**Generational concurrent (`gencon`) policy**
+
+The `<size>` value that you specify for `-Xconcurrentslack` corresponds to additional free space for allocation. But, in `gencon` policy, application threads and the concurrent GC kickoff heuristic do not measure free space in the same way. Application threads allocate mostly in nursery space, but the concurrent GC kickoff heuristic monitors free memory in tenure space.
+
+Because the concurrent GC kickoff heuristic monitors tenure space rather than nursery space, the `<size>` value does not map directly to the free space seen by allocating threads.
+
+Therefore, in `gencon` policy, the amount of headroom is calculated as follows:
+
+> Headroom = (`<size>` ÷ Average promotion rate) × Average application allocation rate
+
+where:
+
+- **`<size>`** is the free memory in tenure space that is specified in the `-Xconcurrentslack` option.
+- **Average promotion rate** is the average rate at which objects are promoted from nursery space into tenure space.
+- **Average application allocation rate** is the average allocation rate from nursery.
+
+:fontawesome-solid-pencil:{: .note aria-hidden="true"} **Note:** The average promotion rate and average application allocation rate can be derived from verbose GC logs. But, in a typical generational workload, the promotion rate is approximately ten times  lesser than the allocation rate. Hence, the formula can be approximated with `<size>` * 10.
+
+In the case of `gencon` policy, use the [`-Xgc:concurrentKickoffTenuringHeadroom`](xgc.md#concurrentkickofftenuringheadroom) option instead of the `-Xconcurrentslack` option for ease.
+
+## See also
+
+- [`-Xgc:concurrentKickoffTenuringHeadroom`](xgc.md#concurrentkickofftenuringheadroom)
+- [Garbage collection policies](gc_overview.md)
 
 <!-- ==== END OF TOPIC ==== xconcurrentslack.md ==== -->
--- a/docs/xgc.md
+++ b/docs/xgc.md
@@ -36,6 +36,7 @@ Options that change the behavior of the garbage collector.
 | [`breadthFirstScanOrdering`         ](#breadthfirstscanordering         ) | Sets the scan mode to breadth first.                                             |
 | [`classUnloadingKickoffThreshold`   ](#classunloadingkickoffthreshold   ) | Sets a threshold to start an early concurrent global garbage collection (GC) cycle due to recent, heavy class loading activity  |
 | [`classUnloadingThreshold`          ](#classunloadingthreshold          ) | Sets a threshold to trigger a class unloading operation in a global GC cycle                                     |
+| [`concurrentKickoffTenuringHeadroom`](#concurrentkickofftenuringheadroom) | Sets an additional percentage of free tenure memory as headroom for concurrent global GC kickoff timing.
 | [`concurrentScavenge`               ](#concurrentscavenge               ) | Enables a GC mode with less pause times.                                             |
 | [`disableEstimateFragmentation`     ](#disableestimatefragmentation     ) | Disables the calculating and reporting of the estimates of the macro fragmentation                                             |
 | [`dnssExpectedTimeRatioMaximum`     ](#dnssexpectedtimeratiomaximum     ) | Sets the maximum percentage of time to spend on local GC pauses                                             |
@@ -58,14 +59,14 @@ Options that change the behavior of the garbage collector.
 | [`suballocatorInitialSize`          ](#suballocatorinitialsize          ) | Sets the initial size in bytes for the `subAllocator` area that is used for compressed references.                               |
 | [`suballocatorQuickAllocDisable`    ](#suballocatorquickallocdisable    ) | Disables mmap-based allocation of the compressed references `subAllocator` area. (Linux only)                            |
 | [`suballocatorQuickAllocEnable`     ](#suballocatorquickallocenable     ) | Enables mmap-based allocation of the compressed references `subAllocator` area. (Linux only)                            |
-| [`synchronousGCOnOOM`               ](#synchronousgconoom               )     | Stops an application to allow GC activity.                                                                             |
-| [`targetPausetime`                  ](#targetpausetime                  )   | Sets the target GC pause time for the `metronome` and `balanced` GC policies.                            |
-| [`targetUtilization`                ](#targetutilization                )   | Sets application utilization for the `metronome` GC policy.                                                   |
+| [`synchronousGCOnOOM`               ](#synchronousgconoom               ) | Stops an application to allow GC activity.                                                                             |
+| [`targetPausetime`                  ](#targetpausetime                  ) | Sets the target GC pause time for the `metronome` and `balanced` GC policies.                            |
+| [`targetUtilization`                ](#targetutilization                ) | Sets application utilization for the `metronome` GC policy.                                                   |
 | [`tlhIncrementSize`                 ](#tlhincrementsize                 ) | Sets the size of the thread local heap (TLH) increment.                                                   |
 | [`tlhInitialSize`                   ](#tlhinitialsize                   ) | Sets the initial size of the thread local heap.                                                           |
 | [`tlhMaximumSize`                   ](#tlhmaximumsize                   ) | Sets the maximum size of the thread local heap.                                                           |
 | [`verboseFormat`                    ](#verboseformat                    ) | Sets the verbose GC format.                                                                               |
-| [`verbosegcCycleTime`               ](#verbosegccycletime                    ) | Sets the criteria for verbose GC logging.                                                                              |
+| [`verbosegcCycleTime`               ](#verbosegccycletime               ) | Sets the criteria for verbose GC logging.                                                                              |
 
 ### `breadthFirstScanOrdering`
 
@@ -92,6 +93,34 @@ Options that change the behavior of the garbage collector.
 : This option sets a threshold that is used to trigger an optional GC class unloading operation in a global GC cycle, irrespective of how the global GC cycle is triggered. The default value is 6.
 
 : This option is applicable to the following GC policies: `gencon`, `optavgpause`, and `optthruput`.
+
+### `concurrentKickoffTenuringHeadroom`
+
+        -Xgc:concurrentKickoffTenuringHeadroom=<value>
+
+: | Setting       | Value          |  Default headroom                                      |
+  |---------------|----------------|------------------------------------------------|
+  | `<value>`     | [percentage]   |    2% for stop-the-world (STW) scavenge <br>    10% for concurrent scavenge       |
+
+: where `<value>` is a percentage (0–99) of the current free memory size, every time kickoff point is calculated.
+
+: :fontawesome-solid-pencil:{: .note aria-hidden="true"} **Note:** Concurrent scavenge has high default headroom because it is more susceptible to promotion fluctuations due to shared survivor and allocate space during an active concurrent scavenge cycle.
+
+: This option is used to add the specified headroom on top of the automatic concurrent kickoff calculation to trigger the concurrent global GC cycle earlier than the GC heuristic computed kickoff point.
+
+: The GC heuristic computes a kickoff point based on steady-state metrics, such as allocation rate and marking speed, to determine when the concurrent global GC cycle should start. The default headroom is added to this kickoff point: 10% when concurrent scavenge is enabled ([`-Xgc:concurrentScavenge`](#concurrentscavenge)), or 2% for stop-the-world scavenge. With the `concurrentKickoffTenuringHeadroom` option, you can override the default headroom.
+
+: For example, if concurrent scavenge is enabled and the heuristic determines that kickoff should occur when 8% of tenure memory is free, the concurrent global GC cycle starts when 18% of tenure memory is free. This 18% = 8% (heuristically based value) + 10% (default headroom for concurrent scavenge).
+
+: If the `concurrentKickoffTenuringHeadroom` option is set to `15`, the effective kickoff point becomes 23% of free tenure memory, where 23% = 8% (heuristically based value) + 15% (`<value>` specified for `-Xgc:concurrentKickoffTenuringHeadroom`).
+
+: The typical scenario where this option is useful is to compensate for spikes in tenuring that would otherwise fill the tenure space faster than the concurrent GC can finish. The headroom triggers the concurrent cycle early enough to absorb those spikes.
+
+: This option was added in the [0.15.0 release](version0.15.md#new-xgc-parameter-related-to-concurrent-kickoff-calculation-is-added) and applies only to the `gencon` GC policy.
+
+: Even though the `-Xconcurrentslack` option is also used to add headroom but in the case of `gencon` GC policy, use the `concurrentKickoffTenuringHeadroom` option.
+
+: For more information about headroom, see [`-Xconcurrentslack`](xconcurrentslack.md).
 
 ### `concurrentScavenge`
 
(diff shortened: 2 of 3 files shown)


Sreekala-Gopakumar:
Jenkins doc stage

Sreekala-Gopakumar:
@amicic - I have added the new content. Please check and let me know if the content is ok. Thanks!

https://eclipse-openj9.github.io/openj9-docs-staging/1724/version0.60/

Sreekala-Gopakumar:
> @amicic - I have added the new content. Please check and let me know if the content is ok. Thanks!
> 
> https://eclipse-openj9.github.io/openj9-docs-staging/1724/version0.60/

@amicic - Could you please check the content and let me know if changes are required. Thanks!

Sreekala-Gopakumar:
@pshipton @amicic - If any changes are required in the updated content, please let me know. If there are no changes, then can this be merged? Thanks!

amicic:
The description itself is good. However, we need to cross reference [-XconcurrentSlack](https://eclipse.dev/openj9/docs/xconcurrentslack/), which is a very similar optionm, but itself not sufficiently documented. I have to think of some new wording for that one, too....

The new one is simpler to understand, should be preferred.

Sreekala-Gopakumar:
Jenkins doc stage

Sreekala-Gopakumar:
@amicic - I have updated the topics. Could you please check and let me know if any changes are required. Thanks!

https://eclipse-openj9.github.io/openj9-docs-staging/1724/version0.60/
https://eclipse-openj9.github.io/openj9-docs-staging/1724/xgc/#concurrentkickofftenuringheadroom
https://eclipse-openj9.github.io/openj9-docs-staging/1724/xconcurrentslack/

docs/xconcurrentslack.md
```diff
@@ -23,22 +23,59 @@
...
 ## Syntax
 
         -Xconcurrentslack<size>
 
-: See [Using -X command-line options](x_jvm_commands.md) for more information about specifying the `<size>` parameter.
+: See [Using -X command-line options](x_jvm_commands.md) for more information about the `<size>` parameter.
+
+## Explanation
+
+Concurrent GC needs time to complete. If the VM waits too long to start concurrent GC, application allocation can keep consuming memory while GC is still marking or reclaiming objects. In that case, the amount of free space can fall too far before the concurrent cycle finishes. Then, the VM might have to use a more urgent collection path, such as stop-the-world GC, which can mean longer pause and reduced application response.
```
amicic: Some things are not quite correct, plus I'd just simplify this, anyway. Something like (but :
```
Ideally, Concurrent GC needs to complete its operation before application consumes all free memory. Otherwise, GC will transition in the stop-the-world mode, which means longer pauses.
```

But I'd something like this:
```
Due to fluctuations in application characteristics such as  allocation rate, object survival rate, CPU utilization, GC may mis-predict the Concurrent GC kickoff point, and start it too late or progress too slowly.
```

Then, try to append this to the next sentence 'to compensate for those fluctuations'

docs/xconcurrentslack.md
```diff
@@ -23,22 +23,59 @@
...
+
+By specifying this option, the VM starts concurrent GC operations earlier to try to keep free heap space available.
+
+Starting concurrent GC operations earlier can reduce pause times, but it can also lengthen concurrent cycles and reduce total throughput.
```
amicic: It will not lengthen it, but it may increase frequency (due to suboptimal use of free memory)

docs/xconcurrentslack.md
```diff
@@ -23,22 +23,59 @@
...
+
+**Generational concurrent (`gencon`) policy**
+
+The `<size>` value that you specify for `-Xconcurrentslack` corresponds to additional free space for allocation. But, in `gencon` policy, application threads and the concurrent GC kickoff heuristic do not measure free space in the same way. Application threads allocate mostly in nursery space, but the concurrent GC kickoff heuristic monitors free memory in tenure space.
+
+Because the concurrent GC kickoff heuristic monitors tenure space rather than nursery space, the `<size>` value does not map directly to the free space seen by allocating threads.
+
+Therefore, in `gencon` policy, the amount of headroom is calculated as follows:
+
+> Headroom = (`<size>` ÷ Average promotion rate) × Average nursery allocation size
```
amicic: correction (I provided it wrong from the beginning): 
Average nursery allocation size -> average application allocation rate

~~perhaps it's good to be more specific:
Average promotion rate - > average tenure promotion rate~~
I see you have extra explanation later

docs/xgc.md
```diff
@@ -93,6 +94,28 @@ Options that change the behavior of the garbage collector.
 
 : This option is applicable to the following GC policies: `gencon`, `optavgpause`, and `optthruput`.
 
+### `concurrentKickoffTenuringHeadroom`
+
+        -Xgc:concurrentKickoffTenuringHeadroom=<value>
+
+: | Setting       | Value          |  Default    |
+  |---------------|----------------|-------------|
+  | `<value>`     | [percentage]   |    0        |
```
amicic: defaults are 10 for CS and 2 for STW Scavenge

docs/xgc.md
```diff
@@ -93,6 +94,28 @@ Options that change the behavior of the garbage collector.
...
+
+: Where `<value>` is a percentage (0–99) of free tenure memory that was observed at the end of the previous global GC cycle.
+
+: This option adds extra headroom on top of the automatic concurrent kickoff calculation to trigger the concurrent global GC cycle earlier than the GC heuristic computed kickoff point.
+
+: The GC heuristic computes a kickoff point based on steady-state metrics, such as allocation rate and marking speed, to determine when the concurrent global GC cycle should start. This kickoff point includes a built-in internal headroom: 10% when concurrent scavenge is enabled ([`-Xgc:concurrentScavenge`](#concurrentscavenge)), or 2% for stop-the-world (STW) scavenge. With the `concurrentKickoffTenuringHeadroom` option, you can add additional headroom beyond that built-in headroom.
```
amicic: it will replace the built-in (default) headroom

docs/xgc.md
```diff
@@ -93,6 +94,28 @@ Options that change the behavior of the garbage collector.
...
+
+: For example, if concurrent scavenge is not enabled and the heuristic determines that kickoff should occur when 10% of tenure memory is free, this 10% already includes 2% built-in headroom for STW. If the `concurrentKickoffTenuringHeadroom` option is set to `5`, the effective kickoff point becomes 15% of free tenure memory.
```
amicic: actually not

the original calculation would be say 18 = 8 (heuristically based) + 10 (default headroom for CS)

if user specifies 15, it will replace the default 10, so it will be 23 = 8 + 15

amicic: you may add somewhere that CS already has high default headroom, because is more susceptible to promotion fluctuations due to shared survivor-allocate space during an active Concurrent Scavenege cycle.

Sreekala-Gopakumar:
Jenkins doc stage

Sreekala-Gopakumar:
Jenkins doc stage

Sreekala-Gopakumar:
Jenkins doc stage

docs/xgc.md
```diff
@@ -93,6 +94,32 @@ Options that change the behavior of the garbage collector.
...
 
+### `concurrentKickoffTenuringHeadroom`
+
+        -Xgc:concurrentKickoffTenuringHeadroom=<value>
+
+: | Setting       | Value          |  Default headroom                                      |
+  |---------------|----------------|------------------------------------------------|
+  | `<value>`     | [percentage]   |    2% for stop-the-world (STW) scavenge <br>    10% for concurrent scavenge       |
+
+: Where `<value>` is a percentage (0–99) of free tenure memory that was observed at the end of the previous global GC cycle.
```
amicic: 'That' is making this confusing potentially incorrect. 'as it' or 'which' might be better

There is lots of free memory created in tenure after a global GC. It's gradually consumed by Scavenges. The kickoff point is constantly evaluated as the tenure free memory is consumed. The option says what part of free memory (observed at the end of global GC) should be dedicated as the headroom. (Rather than applying the percentage at the current free memory size, every time kickoff point is calculated.)

docs/xconcurrentslack.md
```diff
@@ -23,22 +23,59 @@
...
+
+> Headroom = (`<size>` ÷ Average promotion rate) × Average application allocation rate
 
-## Default behavior
+where:
 
-The default value is 0, which is optimal for most applications.
+- **`<size>`** is the free memory in tenure space that is specified in the `-Xconcurrentslack` option.
+- **Average promotion rate** is the average rate at which objects are promoted from nursery space into tenure space.
+- **Average application allocation rate** is the average amount of memory that is allocated in nursery space by application threads.
```
amicic: it's not correct - it's a rate (allocated memory over time) 
not sure if it really needs extra explanation

I actually overcomplicated - both parameters could've been downgraded from rate to memory (rather than upgrading the Nursery memory to allcation rate). But let's leave it with rate to minimize further changes.

docs/xconcurrentslack-backup.md
```diff
@@ -0,0 +1,79 @@
...
+
+: See [Using -X command-line options](x_jvm_commands.md) for more information about the `<size>` parameter.
+
+## Explanation
+
+Instead of waiting until the heuristic's computed kickoff point, you can specify the `-Xconcurrentslack` so that the VM kicks off concurrent garbage collection (GC) earlier — specifically when the amount of free tenure memory falls to the heuristic's kickoff point plus the headroom.
+
+The default value is `0`, which is optimal for most applications. That means that the VM starts GC at the GC heuristic computed kickoff point.
+
+Starting concurrent operations earlier with the help of `-Xconcurrentslack` option reduces pause times but increases the length of concurrent cycles, which can reduce total throughput.
```
amicic: increases the frequency (not length)

docs/xconcurrentslack-backup.md
```diff
@@ -0,0 +1,79 @@
...
+
+Free space is measured from the perspective of allocating threads, which is the same memory that the concurrent kickoff heuristic sees. The `-Xconcurrentslack` value is applied directly as the amount of free tenure memory to keep in reserve.
+
+**Generational concurrent (`gencon`)**
+
+In `gencon` policy, object allocation happens mostly in nursery space, but the concurrent kickoff heuristic monitors free memory in tenure space. Tenure space is consumed only incrementally through scavenge cycles, not directly by allocating threads.
+
+As a result, expressing the slack as a simple amount of free tenure memory does not translate meaningfully to allocating-thread behavior. The effective slack in gencon terms is:
+
+> (Slack to apply to tenure space) ÷ (average promotion rate) × (average nursery allocation size)
```
amicic: needs to be aligned with the other spot

docs/xconcurrentslack-backup.md
Sreekala-Gopakumar: This file was a backup and should have been deleted. I have deleted it now.

docs/xconcurrentslack-backup.md
Sreekala-Gopakumar: This file was a backup and should have been deleted. I have deleted it now.

Sreekala-Gopakumar:
Jenkins doc stage

docs/xconcurrentslack.md
Sreekala-Gopakumar: If you want me to change this into something that you think would be more accurate, you can let me know.

docs/xconcurrentslack.md
```diff
@@ -23,22 +23,59 @@
...
+
+> Headroom = (`<size>` ÷ Average promotion rate) × Average application allocation rate
 
-## Default behavior
+where:
 
-The default value is 0, which is optimal for most applications.
+- **`<size>`** is the free memory in tenure space that is specified in the `-Xconcurrentslack` option.
+- **Average promotion rate** is the average rate at which objects are promoted from nursery space into tenure space.
+- **Average application allocation rate** is the average rate of allocated memory over time.
```
amicic: either 'rate' or 'over time' is redundant

could be
is the average amount of memory allocated from nursery over time
or as simple as 
is the average allocation rate from nursery

docs/xconcurrentslack.md
```diff
@@ -23,22 +23,59 @@
 
 # -Xconcurrentslack
 
-Attempts to keep the specified amount of the heap space free in concurrent collectors by starting the concurrent operations earlier.
-
-Using this option can sometimes alleviate pause time problems in concurrent collectors at the cost of longer concurrent cycles, affecting total throughput.
+This option specifies an absolute amount of free space or headroom (in MB) in the tenure space on top of the automatic concurrent kickoff point. By using this option, you can trigger the concurrent global garbage collection (GC) cycle earlier than the GC heuristic computed kickoff point.
```
amicic: in bytes, not MB

Sreekala-Gopakumar:
Jenkins doc stage

amicic:
Looks pretty good. I'll have another pass with a fresh brain tomorrow. Meanwhile, @pshipton please have a quick look (terminology, language, style etc).

pshipton:
lgtm

Sreekala-Gopakumar:
@amicic @pshipton - Could you please let me know the corresponding development issue or PR number for this?

I saw a PR that was created way back in 2019 where this option is being discussed - https://github.com/eclipse-openj9/openj9/pull/6196

Doesn't that mean this option is not a new option that was introduced in 0.60.0 release? This is just an option that was not documented when it was introduced  (in 0.15.0 release?). 

So that also means that I should not be mentioning this in the What's new in version 0.60.0 topic, isn't it?

amicic:
Correct, the option is not new, it was just not documented. 
Not sure how we handle that case for release notes (@pshipton ?).

docs/xconcurrentslack.md
```diff
@@ -23,22 +23,59 @@
...
-## Default behavior
+where:
 
-The default value is 0, which is optimal for most applications.
+- **`<size>`** is the free memory in tenure space that is specified in the `-Xconcurrentslack` option.
+- **Average promotion rate** is the average rate at which objects are promoted from nursery space into tenure space.
+- **Average application allocation rate** is the average allocation rate from nursery.
 
-This option is not supported with the balanced GC policy (`-Xgcpolicy:balanced`), the *optimize for throughput* policy (`-Xgcpolicy:optthruput`), or metronome GC policy (`-Xgcpolicy:metronome`).
+:fontawesome-solid-pencil:{: .note aria-hidden="true"} **Note:** Because the average promotion rate and average nursery allocation size must usually be derived from verbose GC logs, `-Xconcurrentslack` can be difficult to calibrate for `gencon`. Therefore, in the case `gencon` policy, use the [`-Xgc:concurrentKickoffTenuringHeadroom`](xgc.md#concurrentkickofftenuringheadroom) option for ease.
```
amicic: We can add that while is non-trivial to derive those number precisely the promotion rate is an order of magnitude lower than allocation rate, in a typical generational workload, so that the formula can be approximated to <size> x 10. But, again, it's just simpler to use the other option.

pshipton:
> So that also means that I should not be mentioning this in the What's new in version 0.60.0 topic, isn't it?

It's new that we are documenting it. We can still add it to What's New, but we could perhaps mention an older release version that it works from.

amicic:
One fine detail is that both options can be used, and the effect is additive (both will add their own headroom/slack). But  the user is unlikely to try so (and is discouraged to do), so not sure if we want to go into those details.

Sreekala-Gopakumar:
> > So that also means that I should not be mentioning this in the What's new in version 0.60.0 topic, isn't it?
> 
> It's new that we are documenting it. We can still add it to What's New, but we could perhaps mention an older release version that it works from.

I would recommend not doing that because the What's new topic lists new features and notable changes in OpenJ9 since the previous version and not what was newly added in the document. I can maybe update What's new in version 0.15.0 release and give a link from the content where this option is documented.

pshipton:
> I can maybe update What's new in version 0.15.0 release and give a link from the content where this option is documented.

Works for me.

docs/xconcurrentslack.md
Sreekala-Gopakumar: @amicic - Could you please simplify this? Thanks!

Sreekala-Gopakumar:
> One fine detail is that both options can be used, and the effect is additive (both will add their own headroom/slack). But the user is unlikely to try so (and is discouraged to do), so not sure if we want to go into those details.

Can we maybe provide a note to say that they should use only one of these options and not both?

docs/xconcurrentslack.md
amicic: The average promotion rate and average nursery allocation size can be derived from verbose GC logs. In a typical generational workload the promotion rate is an order of magnitude lower than allocation rate. Hence, the formula can be approximated with size x 10, but (for gencon) it's recommended  just use the -Xgc:concurrentKickoffTenuringHeadroom, instead.

amicic:
> Can we maybe provide a note to say that they should use only one of these options and not both?

Let's skip this detail, after all. We clearly already recommended to use the new option (for gencon).

Sreekala-Gopakumar:
Jenkins doc stage

Sreekala-Gopakumar:
@amicic @pshipton - Please check and let me know if any further changes are required. Thanks!

docs/xconcurrentslack.md
```diff
@@ -23,22 +23,61 @@
...
+
+where:
 
-## Default behavior
+- **`<size>`** is the free memory in tenure space that is specified in the `-Xconcurrentslack` option.
+- **Average promotion rate** is the average rate at which objects are promoted from nursery space into tenure space.
+- **Average application allocation rate** is the average allocation rate from nursery.
 
-The default value is 0, which is optimal for most applications.
+:fontawesome-solid-pencil:{: .note aria-hidden="true"} **Note:** The average promotion rate and average application allocation rate can be derived from verbose GC logs. But, in a typical generational workload, the promotion rate is lesser than the allocation rate. Hence, the formula can be approximated with `<size>` * 10.
```
amicic: If you want to use `lesser`, it should be emphasized like significantly. It's certainly always lesser.

I intentionally used `an order of magnitude`, which indeed means 10x (where the approximated formula comes from).

Sreekala-Gopakumar:
Jenkins doc stage

Sreekala-Gopakumar:
@amicic @pshipton - Is it ok now? Thanks!

pshipton (approved):

amicic (approved):

gjdeval:
Clearly I am late to the party :)  Aleks invited me to comment a couple of days ago, but I only got time to do so this morning.

My instinct is that the reader would benefit from a brief summary of why one might want to use this option, and the general likely effects. I've drafted a couple of paragraphs based on my understanding (quite likely not entirely technically accurate). I would put (something like) this right under the setting/value/default table at the top.

...
This option is used to start (kickoff) concurrent work sooner than the default GC behavior. Starting sooner makes it more likely that concurrent work will complete before free memory is exhausted, reducing the concurrent collection stop-the-world (STW) duration. This option may help if the garbage collection log shows long global GC pauses, where concurrent work has ended prematurely because of tenure allocation failure.

Concurrent work is triggered when remaining free memory at the end of a scavenge falls below a threshold percentage. The default threshold is 2% for STW scavenge and 10% for concurrent scavenge. The <value> used with this option is added to the default threshold, so that kickoff will happen when there is more free memory. This will generally make global GC pauses shorter, with the trade-off that global GCs will be more frequent because the tenure memory area is not completely dirtied before GC launch.
...

gjdeval:
This phrase is very unclear to me:

    where <value> is a percentage (0–99) of the current free memory size, every time kickoff point is calculated.

a) Does "free memory size" refer to tenure area only?
b) Is the kickoff point calculated at the end of a global GC?

If so, wording like this would be much clearer to me:

    <value> is a percentage (0-99) of the free tenure memory at the end of the prior global collection

amicic:
> This phrase is very unclear to me:
> 
> ```
> where <value> is a percentage (0–99) of the current free memory size, every time kickoff point is calculated.
> ```
> 
> a) Does "free memory size" refer to tenure area only? b) Is the kickoff point calculated at the end of a global GC?
> 
> If so, wording like this would be much clearer to me:
> 
> ```
> <value> is a percentage (0-99) of the free tenure memory at the end of the prior global collection
> ```

This should be corrected.  I commented on it (originally, there was word 'that' that I did not like), but seemes that my comment made it worse, but I somehow missed to re-review it.

Probably we can go with your wording (very similar to the original one without the 'that' word), but have a look at my comment which hopefully answers your questions: 
https://github.com/eclipse-openj9/openj9-docs/pull/1724#discussion_r3572845780

amicic:
> ... This option is used to start (kickoff) concurrent work sooner than the default GC behavior. Starting sooner makes it more likely that concurrent work will complete before free memory is exhausted, reducing the concurrent collection stop-the-world (STW) duration. This option may help if the garbage collection log shows long global GC pauses, where concurrent work has ended prematurely because of tenure allocation failure.
> 


We've added similar wording in the Explanation for XconcurrentSlack option, 

```
Ideally, concurrent GC needs to complete its operation before application consumes all free memory. Otherwise, GC will transition into the stop-the-world (STW) mode, which means longer pauses. Due to fluctuations in application characteristics such as allocation rate, object survival rate, CPU utilization, GC might mis-predict the concurrent GC kickoff point, and start it too late or progress too slowly.

To compensate for such fluctuations, you can specify this option so that the VM starts concurrent GC operations earlier to try to keep free heap space available.

Starting concurrent GC operations earlier can reduce pause times, but it might increase frequency because of suboptimal use of free memory.
```

But we can add more for concurrentKickoffTenuringHeadroom, for what you probably refer to. This is the option we already have this early

```This option is used to add the specified headroom on top of the automatic concurrent kickoff calculation to trigger the concurrent global GC cycle earlier than the GC heuristic computed kickoff point.
```

and later this that

```
: The typical scenario where this option is useful is to compensate for spikes in tenuring that would otherwise fill the tenure space faster than the concurrent GC can finish. The headroom triggers the concurrent cycle early enough to absorb those spikes.
```

Perhaps we can append this part with the reason why (to reduce pauses), or put it earlier. 


For this part 

> Concurrent work is triggered when remaining free memory at the end of a scavenge falls below a threshold percentage. The default threshold is 2% for STW scavenge and 10% for concurrent scavenge. The used with this option is added to the default threshold, so that kickoff will happen when there is more free memory. This will generally make global GC pauses shorter, with the trade-off that global GCs will be more frequent because the tenure memory area is not completely dirtied before GC launch.

we already have this

```
: The GC heuristic computes a kickoff point based on steady-state metrics, such as allocation rate and marking speed, to determine when the concurrent global GC cycle should start. The default headroom is added to this kickoff point: 10% when concurrent scavenge is enabled ([`-Xgc:concurrentScavenge`](#concurrentscavenge)), or 2% for stop-the-world scavenge. With the `concurrentKickoffTenuringHeadroom` option, you can override the default headroom.
```

Note that the option will replace/override the default headroom (not add to the defaut headroom ). There is an example just after, explaining that.

```
For example, if concurrent scavenge is enabled and the heuristic determines that kickoff should occur when 8% of tenure memory is free, the concurrent global GC cycle starts when 18% of tenure memory is free. This 18% = 8% (heuristically based value) + 10% (default headroom for concurrent scavenge).
```

If you still want some changes at either of these sposts, please comment directly after the lines that should be changed.

PR #1742, opened 2026-07-28
-Xtune:throughput and -Xtune:footprint

### LINKED ISSUE #1739: -Xtune:throughput and -Xtune:footprint
**Overview:**
Updating external documentation to include Xtune:throughput and Xtune:footprint, similar to similar to https://eclipse.dev/openj9/docs/xtunevirtualized/ 

**Release target:**
Eclipse OpenJ9 0.60.0
however, one feature is already available in the current release, we are just updating documentation to match.

All

- Option name and syntax: `-Xtune:footprint` and `-Xtune:throughput`
- No values set => no units
- purpose: to configure a number of settings on OpenJ9 to tune performance towards certain ends. Similar to `-Xtune:virtualized` designed for CPU contstrained environments. `-Xtune:footprint` minimizes memory usage and `-Xtune:throughput` maximizes throughput at the expense of CPU and footprint.
- restrictions: If these options are used together, the last one takes precedence

Sreekala-Gopakumar:
https://github.com/eclipse-openj9/openj9-docs/issues/1739

Combined the content for `throughput` and `footprint` with the `-Xtune:virtualized` topic. Renamed the topic to `Xtune`.

Closes #1739


Sreekala-Gopakumar:
Jenkins doc stage

Sreekala-Gopakumar:
@KavinSatheeskumar @pshipton - I have created the draft topic so that you can have a better view of the topic after I have combined all these options. You can let me know if combining makes sense and whether it is technically correct.

Also, we might need more content in the footprint and throughput sections to match the content in the virtualized section.

I will update the What's new in version 0.60 topic (and maybe the 0.30 or 0.31 topic as well) after this topic is finalized.

Thanks!

docs/xtune.md
```diff
@@ -0,0 +1,69 @@
...
+| [`virtualized`](#virtualized)   | Reduces OpenJ9 VM CPU consumption when idle            |
+
+:fontawesome-solid-pencil:{: .note aria-hidden="true"} **Note:** These parameters are mutually exclusive and if these parameters are used together, the last one that is specified in the command-line takes precedence.
+
+### `footprint`
+
+       -Xtune:footprint
+
+: Optimizes Eclipse OpenJ9 VM function for virtualized environments, such as a cloud, by minimizing OpenJ9 VM memory usage. When `-Xtune:footprint` is specified, the VM adjusts its internal heuristics to reduce memory consumption, possibly at the expense of other performance metrics.
+
```
KavinSatheeskumar, line 56: You should probably add a note saying something about how AOT is disabled when this flag is enabled (similar to `-Xnoaot`)

KavinSatheeskumar: Suggested


Optimizes Eclipse OpenJ9 VM function for virtualized environments, such as a cloud, by minimizing OpenJ9 VM memory usage. When `-Xtune:footprint` is specified, the VM adjusts its internal heuristics to reduce memory consumption, possibly at the expense of other performance metrics.

Internally, the options makes the JVM's heuristics around garbage collection, and SCC disclaiming more aggressive. In addition, it also disables the use of AOT (similar to the `-Xnoaot` flag)

mpirvu: I would not go into details as these may change over time.
Also, remember that AOT is not disabled due to an implementation "bug". That's fine because I didn't see any improvement from not using AOT.

mpirvu: "Optimizes Eclipse OpenJ9 VM function for virtualized environments"  I would eliminate the reference to virtualized environments. The footprint reduction applies equally to non-virtualized environments. Also, if this option is used specifically for virtualized environments, then one may ask what is -Xtune:virtualized used for, or is everything under -Xtune:footprint under -Xtune:virtualized as well?

docs/xtune.md
```diff
@@ -0,0 +1,69 @@
...
+       -Xtune:throughput
+
+: Increases resource utilization to maximize throughput. This setting has higher precedence over other performance metrics. In particular, if not explicitly provided, the initial heap size (`-Xms`) is set to the maximum heap size value (-`Xmx`). Moreover, the JIT internally sets some options to optimize the code more aggressively.
+
+### `virtualized`
+
+       -Xtune:virtualized
+
+: Optimizes Eclipse OpenJ9 VM function for virtualized environments, such as a cloud, by reducing OpenJ9 VM CPU consumption when idle.
```
mpirvu: Here, I would eliminate "when idle". Some of the heuristics apply to no-idle as well.

docs/xtune.md
```diff
@@ -0,0 +1,69 @@
...
+
+: :fontawesome-solid-pencil:{: .note aria-hidden="true"} **Note:** Performance is optimized if there is a large shared classes cache (SCC) and AOT space in the SCC is not capped.
+
+: This option is recommended for CPU-constrained environments, such as those found in cloud deployments that use containers. Internally, the option makes the JIT compiler more conservative with inlining and recompilation decisions, which saves CPU resources. The Garbage Collector also reduces the rate of heap expansion, which reduces the memory footprint. These changes to reduce the amount of CPU that is consumed are at the expense of a small loss in throughput.
+
+: When `-Xtune:virtualized` is used along with the [`-Xshareclasses`](xshareclasses.md) option, the JIT compiler is more aggressive with its use of [AOT-compiled code](aot.md) compared to setting only `-Xshareclasses`. This action provides additional CPU savings during application start-up and ramp-up, but comes at the expense of an additional small loss in throughput.
```
mpirvu: "but comes at the expense" --> "but may come at the expense"
We have seen that AOT code is sometimes even better than jitted code because of the extra optimizations we do in the cold runs.

mpirvu (commented):
I would also add a note to prevent users adding several of these options on the command line.:
"The -Xtune options are designed to optimize the JVM for a specific performance objective by adjusting a set of internal heuristics. Using more than one -Xtune option at a time can lead to conflicting tuning decisions, as the selected objectives may be mutually competitive. Consequently, the resulting behavior might not reflect the intended benefits of any individual tuning mode."

mpirvu:
Since they are all on the same page, maybe we should also mention -Xquickstart which is an alias for -Xquickstart

Sreekala-Gopakumar:
> I would also add a note to prevent users adding several of these options on the command line.: "The -Xtune options are designed to optimize the JVM for a specific performance objective by adjusting a set of internal heuristics. Using more than one -Xtune option at a time can lead to conflicting tuning decisions, as the selected objectives may be mutually competitive. Consequently, the resulting behavior might not reflect the intended benefits of any individual tuning mode."

@mpirvu - But, we are also saying that "if these parameters are used together, the last one that is specified in the command-line takes precedence.". 

So, if only the last one takes precedence then would "Using more than one -Xtune option at a time lead to conflicting tuning decisions, as the selected objectives may be mutually competitive. Consequently, the resulting behavior might not reflect the intended benefits of any individual tuning mode."?

Sreekala-Gopakumar:
Jenkins doc stage

Sreekala-Gopakumar:
@mpirvu @KavinSatheeskumar @pshipton  - Please check if the changes are as discussed here. Thanks!

docs/xtune.md
```diff
@@ -0,0 +1,83 @@
...
+## Parameters
+
+| Parameter                       | Effect                                                           |
+|---------------------------------|------------------------------------------------------------------|
+| [`footprint`](#footprint)       | Minimizes OpenJ9 VM memory usage                                 |
+| [`quickstart`](#quickstart)     | Causes the JIT compiler to run with a subset of optimizations    |
+| [`throughput`](#throughput)     | Increases resource utilization to maximize throughput            |
+| [`virtualized`](#virtualized)   | Reduces OpenJ9 VM CPU consumption when idle                      |
+
+:fontawesome-solid-pencil:{: .note aria-hidden="true"} **Note:** These parameters are mutually exclusive and if these parameters are used together, the last one that is specified in the command-line takes precedence.
```
mpirvu, line 41: "These parameters are mutually exclusive and if these parameters are used together, the last one that is specified in the command-line takes precedence."
I looked at the code and that is correct; I forgot this is enforced by the VM code.

docs/xtune.md
```diff
@@ -0,0 +1,83 @@
...
+
+: Causes the JIT compiler to run with a subset of optimizations, which can improve the performance of short-running applications.
+
+: `-Xtune:quickstart` is an alias of [`-Xquickstart`](xquickstart.md).
+
+### `throughput`
+
+       -Xtune:throughput
+
+: Increases resource utilization to maximize throughput. This setting has higher precedence over other performance metrics. In particular, if not explicitly provided, the initial heap size (`-Xms`) is set to the maximum heap size value (-`Xmx`). Moreover, the JIT internally sets some options to optimize the code more aggressively.
```
mpirvu: "This setting has higher precedence over other performance metrics."
It's not clear to me what this means.

docs/xtune.md
```diff
@@ -0,0 +1,83 @@
...
+
+### `virtualized`
+
+       -Xtune:virtualized
+
+: Optimizes Eclipse OpenJ9 VM function for virtualized environments, such as a cloud, by reducing OpenJ9 VM CPU consumption.
+
+: :fontawesome-solid-pencil:{: .note aria-hidden="true"} **Note:** Performance is optimized if there is a large shared classes cache (SCC) and AOT space in the SCC is not capped.
```
mpirvu: "Performance is optimized if there is a large shared classes cache (SCC) and AOT space in the SCC is not capped."
Performance can mean many things. I would rephrase to read something along the lines: "The effect of -Xtune:virtualized is maximized when a large shared classes cache (SCC) is provided and the AOT space in the SCC is not capped"

mpirvu (approved):
I only have some small comments.

mpirvu:
> ** But, we are also saying that "if these parameters are used together, the last one that is specified in the command-line takes precedence.".**

I looked at the code and indeed, the VM ensures that only that last one takes effect.

docs/xtune.md
Sreekala-Gopakumar: @mpirvu  - I referred to the content given in https://github.com/eclipse-openj9/openj9/pull/14277#issue-1102566807. How would you like me to rephrase it so that it makes sense? Did that content mean

> By using this option, the VM gives more importance to throughput over other performance metrics.

Or something similar?

 Thanks!

mpirvu: Yes, that one sounds good, or something along the lines:
"When this option is used, the JVM will change its internal heuristics to favor throughput over other performance metrics. In particular, ...."

Sreekala-Gopakumar:
Jenkins doc stage

Sreekala-Gopakumar:
@mpirvu @KavinSatheeskumar @pshipton - Any other changes? Can this be merged?

mpirvu:
LGTM

pshipton (approved):

PR #1744, opened 2026-07-29
Update the python version to match the updated image

pshipton:
Changed from python3.10 to python3.12 due to https://github.com/eclipse-openj9/openj9-docs/pull/1743


keithc-ca (approved):
Confirmed that path is present in the image I build previously.

PR #1745, opened 2026-07-30
In the Dockerfile delete the unneeded ubuntu user

pshipton:
This allows the jenkins user to use 1000:1000. It defaults to these values but set it explicitly since it matters. It needs to match the id of the host system where it runs.


keithc-ca (approved):
We ought to remove the need to use a specific uid/gid, but that can be addressed later.

pshipton:
https://openj9-jenkins.osuosl.org/view/Website-Doc/job/Build-Doc-Docker_Container/102/

